PR #168, opened 2018-06-20
code coverage settings

kevinpapst:
Unfortunately that one decreases the code coverage, but it becomes more realistic this way...


kevinpapst:
Decided to revert all changes as its unclear yet, whether the integration tests should count onto the code coverage or not

PR #180, opened 2018-06-24
language specific money and date display

### LINKED ISSUE #110: Set PHP locale for date format and month names
Add subscriber that sets the locale by the current users language:
`setlocale(LC_ALL, $locale1, $locale2);`

Testing:
- Invoices: the "timesheet" renderer displays a month name, which must be translated properly
- Timesheets: the rate must be formatted correctly

kevinpapst:
- Display month name according to the users locale, instead of being translated with the PHP system locale.
- Display rates according to the users locale
- Display date saccording to the users locale 
- Sneaked in a UI improved in the template list view (action buttons).
- Added tests and checked code styles.
- Language specific configurations for date formats 

Fixes #110

### DIFF
--- a/tests/Twig/DateExtensionsTest.php
+++ b/tests/Twig/DateExtensionsTest.php
@@ -60,7 +60,6 @@ class DateExtensionsTest extends TestCase
             'de' => ['date_short' => 'd.m.Y'],
             'en' => ['date_short' => 'Y-m-d'],
             'ru' => ['date_short' => 'd.m.Y'],
-
         ]);
         $this->assertEquals($result, $sut->dateShort($date));
     }


simonschaufi:
Can you write a short step by step instruction how to test this?

kevinpapst:
Quick and simple:
- Checkout branch
- You need an installation with a lot of timesheet entries with different customers and currencies (us the test fixtures)
-  Switch the languages

Expected results:
- the date format in the data tables should change (en = Y-m-D, de = d.m.Y)
- the currency format should change for the rates (de: 1.234,54 €, en: 1,234.54 €) 
- the month name can be tested by changing the invoice template to "Timesheet" - there is a month name in the upper left invoice like "January 2018" - that must change accordingly to the user language as well

simonschaufi:
The currency should be customer specific, not depending on the language. I once had a customer in swiss and I created the invoice in swiss francs.

Issue #207, opened 2018-07-12
Feature request - Monthly overview / report


eporsche:
## Description 
Instead of a list of all time records, it would be nice to see the full month. And for which days a time recording has been created. That way an employee could easily see for which days a time record might be missing. Also national holidays and weekends could be added to this view. 
With this information added to the monthly overview an employee can see how much hours he has worked this month and how much hours he has still to work. Holidays and weekends would then be taken into consideration.

## Expected Behavior
* Monthly overview (calendar view)
* Ability to enter planned working time per employee (not per project)
   * time calculation model: monthly, yearly, accumulated
   * support for full time (planned working time 100% per week) / half time employees (planned working time 50% per week)  or pay per hour workers (0% planned per week)
* Ability to enter holiday budget per employee
* Ability to calculate overtime
* Ability to enter times absent due to overtime
* Ability to enter absence due to illness
* Ability to enter absence due to holiday
* Holidays and weekends highlighted 
* PDF Download for monthly report
* Disable past month time recording for accounting

## Actual Behavior
* list of all time records per employee

kevinpapst:
these are a lot if ideas on one topic, but i like them! its already planned to work in the reporting features, so i move this ticket to the next milestone. i'll come back for discussion in a few weeks.

hmr-it-jr:
+1 for these:
- Ability to enter holiday budget per employee
- Ability to calculate overtime
- Ability to enter times absent due to overtime
- Ability to enter absence due to illness
- Ability to enter absence due to holiday

We "solved" this in Kaimai 1 with creating a customer "loss of working hours" and different projects like "illnes", "holidays", ... and an activity "not in office".

kevinpapst:
Guys, I started to work on the calendar view, please see #236 (there is a screenshot attached).
Please note that this Feature request is basically 10 in one, so this is impossible to add in "one session".

I will treat this PR as the basic calendar functionality including display of public holidays. If that is done, I will close this PR. All your other ideas (which are great 👍 and shouldn't get lost!) need to be rewritten in single feature requests (so people can vote). 

Please take some time and write them one by one and then we can prioritize these features and plan them for the next milestones.

mr-jingles:
The calendar view looks great. I installed you pull request locally.

On my wish-list would be the following  
- a click in a free area on the calendar opens the "add new time page" in  a modal window - I don't think that more fields are required here
- click on a existing entry opens the modal window to edit the time
[...]

PR #269, opened 2018-08-12
dashboard widgets are configurable via config

kevinpapst:
Added configurable dashboard widgets.

Widgets are in defined in kimai.yaml at `kimai.widgets` and the dashboard itself is configured at `kimai.dashboard`. The Katter one references widgets to display in each row, e.g.:

```yaml
kimai:
    dashboard:
        user:
            title: dashboard.you
            permission: ROLE_USER
            widgets: [userDurationToday, userDurationWeek, userDurationMonth, userDurationYear]

    widgets:
        userDurationToday: { title: stats.durationToday, query: duration, user: true, begin: '00:00:00', end: '23:59:59', icon: duration, color: green }
        userDurationWeek: { title: stats.durationWeek, query: duration, user: true, begin: 'monday this week 00:00:00', end: 'sunday this week 23:59:59', icon: duration, color: blue }
        userDurationMonth: { title: stats.durationMonth, query: duration, user: true, begin: 'first day of this month 00:00:00', end: 'last day of this month 23:59:59', icon: duration, color: purple }
        userDurationYear: { title: stats.durationYear, query: duration, user: true, begin: '01 january this year 00:00:00', end: '31 december this year 23:59:59', icon: duration, color: yellow }
        userDurationTotal: { title: stats.durationTotal, query: duration, user: true, icon: duration, color: red }
```

Now its possible to define the data inside the widgets by the person hosting Kimai 2.

You can simply overwrite the settings in `local.yaml`

```yaml
    dashboard:
        user: { widgets: [userDurationWeek, userDurationMonth, userDurationYear, userDurationTotal] }
```

There are some widgets pre-defined, which are not displayed.

See #207

**TODO**
- [x] add an embed and example for a chart
- [x] add documentation

### DIFF
--- a/var/docs/dashboard.md
+++ b/var/docs/dashboard.md
@@ -122,5 +122,5 @@ all of the following widgets in the configured section (the above example overwr
 
 In order to work, the chart widget `userRecapTwoYears` needs to be the first in the section.
 
-This widgets setup amd configuration will likely change in the future, so keep an eye on this config if this widget doesn't work 
-after one of the next updates! 
+This widgets setup and configuration will likely change in the future, so keep an eye on this config if this widget  
+doesn't work after one of the next updates! 


sideeffect42:
Thanks for having a go at this. Simply awesome, and exactly what I needed!

Is there a way to reorder the dashboard rows? I added a custom row in my `local.yaml`, but it's always displayed at the bottom.

kevinpapst:
@sideeffect42 reordering is not possible, but you can simply overwrote one or all of the pre-configred sections by defining them in your `local.yaml` - I added a [documentation chapter about it](https://github.com/kevinpapst/kimai2/blob/b95dff2872ad7dcefd60e7dfbf95e06145222ce0/var/docs/dashboard.md) (right at the bottom).
Please check if that works for you.

sideeffect42:
@kevinpapst Overwriting all the default sections does work, it is a bit of an unfortunate solution, tough.

I thought about how this could be solved, but couldn't think of a really good solution.
The best one I could come up with would be to prepend all names with an integer and then sorting the hash map before rendering it.

kevinpapst:
@sideeffect42 please check out the latest commit with section ordering

sideeffect42:
@kevinpapst Awesome. I really like the solution with the `order` property.

Another idea that came to my mind is that it would be really cool if the tiles could be filtered by customer.
I imagine that this wouldn't be too easy to implement, though.

kevinpapst:
Customers would be great indeed, but please create a new feature request for this - so I can add it later.

sideeffect42:
@kevinpapst Sure, just wanted to ask you first if you could think of such a feature. See #272.

kevinpapst:
@sideeffect42 if you want to help, you could check the latest commit.
I added a chart widget which is not activated by default. Docu [is here](https://github.com/kevinpapst/kimai2/pull/269/commits/e6260e8dabed5465876e1667006f2a4448551749#diff-35e03ae5399d30e5d8bb72c7eee30ea4). Simply put that into your `local.yaml`:
```
kimai:
    widgets:
        userRecapTwoYears: { title: stats.durationToday, query: monthly, user: true, begin: '01 january last year 00:00:00', end: '31 december this year 23:59:59', color: '#3b8bba|rgba(0,115,183,0.6);#c1c7d1|rgb(210,214,222,0.9)' }

    dashboard:
        user_duration:
            type: chart
            widgets: [userRecapTwoYears, userDurationToday, userDurationWeek, userDurationMonth, userDurationYear]
```

PR #308, opened 2018-09-11
automatically stop other time sheets when starting a new one.

danielvandenberg95:
## Description
Automatically stop other time sheets when starting a new one to allow for faster switching between tasks.

## Types of changes

## Checklist
- [x] I have updated the documentation ~~accordingly~~ _where neccesary_
- [x] I have added tests to cover my changes _where neccesary_

### DIFF
--- a/src/Controller/TimesheetController.php
+++ b/src/Controller/TimesheetController.php
@@ -110,11 +110,11 @@ class TimesheetController extends AbstractController
     {
         $user = $this->getUser();
 
-	//Stop currently running actions.
-	$activeEntries = $this->getRepository()->getActiveEntries($user);
-	foreach ($activeEntries as $activeEntry) {
-		$this->stop($activeEntry, 'timesheet');
-	}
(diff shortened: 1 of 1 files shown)


kevinpapst:
Hi @danielvandenberg95, I really appreciate your PR, thanks for that 👍 

I would like to see some additions, as having multiple running timers is a MUST-HAVE feature for some user. 
There is a visual sign if you have more than X running entries, see this config:
https://github.com/kevinpapst/kimai2/blob/master/config/packages/kimai.yaml#L97

Probably we could use that config not only as a soft-warning, but as a hard limit:
- If it is set to 1, then the currently running entry is stopped and the new one is started.
- If the limit is configured to 2 or higher, just show a flash message if you try to start a new entry but already have x running entries

What do you think?

And btw: please execute `bin/console kimai:phpcs --fix` to apply Kimai code-styles.

kevinpapst:
closing in favor of #386

PR #259, opened 2018-08-08
Support global activities

kevinpapst:
**Attention: this branch needs extensive testing, as I had to touch all main components of Kimai.**

Fixes: #200 
Fixes: #250 
Fixes: #325 

**Tasks**
- [x] Activities do not need to have a project assigned
- [x] Timesheets have an activity and project assigned
- [x] add migration for new project_id field in timesheet table
- [x] add javascript for loading data in Customer/Project/Activity step-based forms
- [x] support selection of Activity & Project in TimesheetEditForm
- [x] adapt "recent activities" drop-down to new global activities
- [x] adapt Activity API (filter for globals, project, visible)
- [x] fix Unit Tests
- [x] add tests for new Code (AI) and changed code coverage (see Scrutinizer)
- [x] search for new FIXME/TODO in PR
- [x] make columns in timesheet not nullable

**Nice to have**
- filter for "global" in Activity Admin (needs new feature request)
- use a query plugin for the select boxes in all edit/create forms that supports searching (see #392)

PR #332, opened 2018-09-26
API endpoint for Timesheet entries

### LINKED ISSUE #315: API: Timesheets
I want to be able to query for open timesheets in order to update existing time sheets (set end date) or to create new timesheets within a shell script (using curl).

Add three REST endpoints:
  1) GET /api/timesheets ->
Similar to the existing EPRs such as /api/activities and return all timesheets. As far as I can see a new model needs to be defined here (or at least added to the API documentaiton). Some kind of filter function would be great here (something like adding a get parameter "enddate"="" or "startdate"=2018/09/21" or startdate="21.09.2018"), so that only non-finished timesheets are returned. Maybe it makes sense to return only parts of the model (maybe only id/startdate/enddate while skipping comments)
  2) GET /api/timesheets/{id}
Delivers all the details for a single timesheet (including comments). Can also be used to check if id already exists (HTTP 404 if id is not used... HTTP 200 including details in any other case)
  3) PUT /api/timesheets/{id}
Either creates a new timesheet (if id was not used before) or updates an existing one

Not much... However, I can think of various implementations of the filter method suggested for the GET /api/timesheets. One could either use get parameters or maybe various endpoints (something like GET /api/timesheets/open, /api/timesheets/closed, ...). I personally would prefer get parameters...

**Additional context**
The new EPRs should also support the authentication mechanisms of the existing REST endpoints.

### LINKED ISSUE #333: API: Add filtering/sorting/limits
If I want to find the latest id of an entity used by kimai2, I need to retrieve all the entities, sort them by a date/id and then filter. If this would be done within the API, responses could be much smaller.

GET Parameters in the API calls. Something like: "api/activities?sortBy=date&sortOrder=DESC&limit=2"
In this case only the two newest activities would be returned.

Another way of filtering would less SQL inspired, but more related to query by example. Something like "api/activities?date=>2018-08-08&limit=2" could also achieve something similar (however, i prefer the sql inspired solution :-)).

**Additional context**
Personally this is not mandatory to work with the API, but for larger systems I think that responses will be quite big without this feature (and so the API might become unusable).

kevinpapst:
Closes #315 
Closes #333  

Allows further work on https://github.com/bastilimbach/timeu/pull/8

**TODO**
- [x] fix TODO in TimesheetController
- [x] check if virtual property project_id should be replaced by project entity (use different serialization groups for full entity and sub-resource entity)


bastilimbach:
I just tested the branch locally. Works great so far!
I think the API could really benefit from some date filters. I'm imagining, that this endpoint could get really slow if the user has many timesheet entries and we always fetch everything at once. The API could also use pagination instead of date filters. Then the client could specify the number of entries he wants to retrieve.

Also, the API needs to change quite a bit after global activities are implemented, right? Should PR #259 be merged before implementing this endpoint? Global activities and the API are currently the two major blockers of the Kimai 1 to 2 migration for my employer.

kevinpapst:
Just wanted to let you know that I am still here ;-) I just have some "super-duper-24-hours-busy" private stuff going on. I hope I will find time again to work on Kimai 2 in the next weeks, when things calm down.

kevinpapst:
Added an endpoint to create timesheet entries:
```
curl -X POST \
  http://127.0.0.1:8000/api/timesheets \
  -H 'Content-Type: application/json' \
  -H 'X-AUTH-USER: susan_super' \
  -H 'X-AUTH-TOKEN: api_kitten' \
  -d '{
	"activity": 1,
	"begin": "2018-10-29 12:29",
	"end": "2018-10-29 20:37",
	"description": "sdfsdf",
	"fixedRate": null,
	"hourlyRate": 85
}'
```

kevinpapst:
As I merged global activities, you need to submit the `project` field now.

kevinpapst:
@bastilimbach this PR includes some BC breaks, as the API fields were renamed and changed.
Collections include less information than entities and I updated all existing endpoints.

Can you test it and let me know if there is something fundamentally wrong or missing?

kevinpapst:
@bastilimbach I want to release 0.6 soon. Can you estimate if you find the time next week? No problem if not, then I will merge it and we can fix/adapt stuff later.

bastilimbach:
I just tested the new API. Works great so far. I, however, noticed one or two small issues:

`POST` **If the body contains everything except the project, the error message isn't correct:**
###### Request body
```json
{
  "activity": 1,
  "begin": "2018-10-29 12:29",
  "end": "2018-10-29 20:37",
  "description": "",
  "fixedRate": 2,
  "hourlyRate": 85
}
```
###### Response body
```json
{
  "code": 400,
  "message": "Validation Failed",
  "errors": {
    "children": {
      "begin": {},
      "end": {},
      "customer": {},
      "project": {
        "errors": [
          "A timesheet must have a project.",
          "This value should not be null."
        ]
      },
      "activity": {
        "errors": [
          "This value is not valid."
        ]
      },
      "description": {},
      "fixedRate": {},
      "hourlyRate": {}
    }
  }
}
```
###### Expected result
It should just show "A timesheet must have a project. [...]". The activity is correct in this example. It works correctly if you send a project but forget the activity. This issue also occurs if the entered project ID is wrong.

`POST` **If the begin timestamp isn't provided, the server returns a 500 error:**
###### Request body
```json
{
  "activity": 1,
  "project": 1,
  "end": "2018-10-29 20:37",
  "description": "sdfsdf",
  "fixedRate": null,
  "hourlyRate": 85
}
```
###### Response body
```json
{
  "code": 500,
  "message": "Call to a member function getTimestamp() on null"
}
```
###### Expected result
The server should return an error stating, that the begin time is missing.

`POST` **Timestamp formatting:**
Currently, the API only accepts timestamps which are formatted as ISO 8601 and without the "T" separator. Example: `2018-11-16 20:25:29+01:00`
Every other format returns the following error:
```json
{
  "code": 400,
  "message": "Validation Failed",
  "errors": {
    "children": {
      "begin": {
        "errors": [
          "This value is not valid."
        ]
      },
      "end": {},
      "customer": {},
      "project": {},
      "activity": {},
      "description": {},
      "fixedRate": {},
      "hourlyRate": {}
    }
  }
}
```
The following formats are affected:
- ISO 8601 with "T" separator: `2018-11-16T20:29:24+01:00`
- UNIX timestamp (epoch time): `1542396637`
- RFC 1123/2822: `Fri, 16 Nov 2018 20:31:12 +0100`

If you do not want to support these formats, we should mention the right format in the API documentation.

`GET` **Page query parameter irrelevant without size parameter:**
If the `size` query parameter is not provided, the `page` parameter shouldn't impact the request, correct? Maybe the API could return an error if the request is sent with a `page` parameter but without a `size` parameter.

**Additional points:**
1. A timesheet entry is created, even if the `end` parameter is not provided. That's the expected behavior, isn't it?
2. If no `hourlyRate` is provided, the timesheet entry gets created with a default value of `83.0`. Is that correct?

kevinpapst:
I'll merge it, as I want to release 0.6. As everything until 1.0 is "in progress" we can continuously improve it.

PR #358, opened 2018-10-17
Default country (#2)

tobybatch:
## Description
I've added some config so that you can change the default country, currency and timezone. I've added a key to the kimai.yaml file and used that to pre-fill the customer object in the customer edit controller.

## Types of changes

## Checklist

### DIFF
--- a/config/packages/kimai.yaml
+++ b/config/packages/kimai.yaml
@@ -156,9 +156,8 @@ kimai:
         activeRecordings: { title: stats.activeRecordings, query: active, icon: duration, color: red }
 
     # Default settings used to populate forms
-    defaults:
-        # New customer form
-        customer:
-            timezone: Europe/Berlin
-            country: DE
-            currency: EUR
+    #defaults:
+    #    customer:
+    #        timezone: Europe/Berlin
+    #        country: DE
+    #        currency: EUR
--- a/src/Controller/Admin/CustomerController.php
+++ b/src/Controller/Admin/CustomerController.php
@@ -32,7 +32,7 @@ class CustomerController extends AbstractController
      * @var array
      */
     private $defaults;
-    
+
     /**
      * @param array $defaults
      */
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -311,9 +311,10 @@ class Configuration implements ConfigurationInterface
         $node = $builder->root('defaults');
 
         $node
-            ->requiresAtLeastOneElement()
-                ->useAttributeAsKey('key')
-                ->arrayPrototype()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('customer')
+                    ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('timezone')->defaultValue('Europe/Berlin')->end()
                         ->scalarNode('country')->defaultValue('DE')->end()
--- a/var/docs/configurations.md
+++ b/var/docs/configurations.md
@@ -233,12 +233,13 @@ kimai:
                 factor: 1.5
 ```
 
-## Customer (kimai.yaml)
+## Forms (kimai.yaml)
 
-### Defaults
+You can set some defaults for various forms within Kimai, to easify the creation process.
 
-You can set some defaults for customer creation.  These override the default timezone, currency and country set in kimai.yml
+### Customer creation
 
+Define the default values for a customer like this: 
 ```yaml
 kimai:
     defaults:


kevinpapst:
Hi @tobybatch ! Thanks for the contribution 👍 
Can you ping me, once you think the PR is not WIP any more, so I do a review?!
The SystemPreferences table is a really great idea, I wanted to work on this one for a longer time now. 
If you want to add it in this PR (which is not necessary by any means) it needs some more improvements. Also the migrations need to be done for SQLite and MySQL (which I could help with).

I don't want to talk you out of it ;-) just want to raise, that is is not mandatory to introduce it for this feature - it will just complicate things much more.

tobybatch:
Sorry @kevinpapst that was sloppy of me.  I started thinking that I'd add a system config page to set these things and realised that I'd need to add a config section to the layout.  That's not really where my strength lie.  So Occam's razor was applied and I went for a yaml change instead.

I'll back out the DB changes (I'll leave them in a branch on my fork), and get it passing the tests.  It should be only trivial changes then.

src/Controller/Admin/CustomerController.php
```diff
@@ -67,7 +67,14 @@ public function indexAction($page, Request $request)
      */
     public function createAction(Request $request)
     {
-        return $this->renderCustomerForm(new Customer(), $request);
+        $defaults = $this->getParameter('kimai.defaults')['customer'];
```
kevinpapst: Please use dependency injection for these parameters, instead of accessing the super-global container here:
```
        $defaults = $this->defaults;
```
and something like 
```
    public function __construct(array $defaults) {
        $this->defaults = $this->defaults;
    }
```
Please have a look at the `config/services.yaml` how to configure the values that will be passed to the constructor. As this is an array it can't be auto wired: https://github.com/kevinpapst/kimai2/blob/master/config/services.yaml#L49

src/DependencyInjection/Configuration.php
```diff
@@ -303,4 +304,32 @@ protected function getDashboardNode()
...
+
+    protected function getDefaultsNode()
+    {
+        $builder = new TreeBuilder();
+        $node = $builder->root('defaults');
+
+        $node
+            ->requiresAtLeastOneElement()
+                ->useAttributeAsKey('key')
+                ->arrayPrototype()
```
kevinpapst: This whole section looks like Copy&Paste and partially wrong. Please use `getLanguagesNode()` or `getThemeNode()` as template.

config/packages/kimai.yaml
```diff
@@ -154,3 +154,9 @@ kimai:
         activeUsersYear: { title: stats.userActiveYear, query: users, begin: '01 january this year 00:00:00', end: '31 december this year 23:59:59', icon: user, color: yellow }
         activeUsersTotal: { title: stats.userActiveTotal, query: users, icon: user, color: red }
         activeRecordings: { title: stats.activeRecordings, query: active, icon: duration, color: red }
+
+    defaults:
```
kevinpapst: Please comment that complete block and also add a one-line documentation what it is - so something like `# default values for various forms`.
If you get the `Configuration` object correct, it will work without these config settings in `kimai.yaml`. Those values should be set by the administrator in `local.yaml`.

After changing the Configuration object and commenting this block, use `bin/console cache:clear` to test the changes.

kevinpapst:
Occam's razor - never heard of that, good one :-D

config/packages/kimai.yaml
tobybatch: The local overrides work.  Mine show up as GBP, London and UK.

src/DependencyInjection/Configuration.php
tobybatch: Sorry, I'd not used this config specification mechanism before.  I home the new code is right.

tobybatch:
Clever fellow Occam.  I often think he should be compulsory teaching for engineers. ;)

src/Controller/Admin/CustomerController.php
```diff
@@ -28,6 +28,23 @@
  */
 class CustomerController extends AbstractController
 {
+    /**
+     * Default form settings from kinai.yml/local.yml
```
kevinpapst: Can you please remove complete line or at least the part `from kinai.yml/local.yml`?

src/Controller/Admin/CustomerController.php
```diff
@@ -28,6 +28,23 @@
...
+     *
+     * @var array
+     */
+    private $defaults;
+    
+    /**
+     * COnstruct a new customer controller with a parameters array.
```
kevinpapst: Please delete that comment

config/packages/kimai.yaml
kevinpapst: Fine for me then

src/DependencyInjection/Configuration.php
kevinpapst: Well, actually I think it should be:
```php
        $builder = new TreeBuilder();
        $node = $builder->root('defaults');

        $node
            ->addDefaultsIfNotSet()
            ->children()
                 ->scalarNode('timezone')->defaultValue('Europe/Berlin')->end()
                 ->scalarNode('country')->defaultValue('DE')->end()
                 ->scalarNode('currency')->defaultValue('EUR')->end()
            ->end()
        ;

        return $node;
```

kevinpapst:
Hey @tobybatch, I pushed a small change regarding the config and documentation.
Now its ready to be merged, thanks again for that feature 👍

PR #377, opened 2018-10-27
Dev/demo dockerfile

tobybatch:
## Description

Single docker file for a dev/demo instance.  I'll get to work a more complicated set of files in a new repo.

## Types of changes

## Checklist
- N/A My code follows the code style
- N/A All methods have a doc header with type declarations
- N/A I have added tests to cover my changes

### DIFF
--- a/var/docs/docker.md
+++ b/var/docs/docker.md
@@ -2,23 +2,30 @@
 
 We bundle a simple docker that can be used for test purposes.  Either pull it from docker hub or build it.
 
-## Build the docker:
+## Build the docker
 
     docker build -t kimai/kimai2:dev .
 
 ## Run the docker
 
-    docker run -ti -p 8001:8001 --name kimai2 --rm kimai/kimai:dev2
+    docker run -ti -p 8001:8001 --name kimai2 --rm kimai/kimai2:dev
 
-The you can then access the site on http://127.0.0.1:8001 or on Mac you may need to use the docker IP:
+You can then access the site on http://127.0.0.1:8001. If that doesn't work check the IP of your docker:
 
     docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' kimai2
 
-You can hit kimai at that ip on port 8001.
+You can find Kimai at that IP on port 8001.
+
+### Mac using docker-machine
+
+When using dock-machine on your Mac, you need to use the IP of your machine. 
+Considering you started the machine named `default`, you find the IP with:
+
+    docker-machine ip default
 
 ## Running commands in the docker
 
-You can run any command in the container in this fashion once it is started.  Add -ti to attach a terminal.
+You can run any command in the container in this fashion once it is started.  Add `-ti` to attach a terminal.
 
     docker exec -ti kimai2 bash
 


Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+        zip \
+        && \
+    docker-php-ext-install \
+        gd \
+        intl \
+        ldap \
+        pdo_mysql \
+        zip \
+        && \
+    docker-php-ext-install \
```
kevinpapst: I haven't setup a docker machine in a while and haven't seen these commands as well, so this question might be stupid... : exactly the same command is executed before (line 22), why are these commands duplicated here?

kevinpapst:
Trying to test this ... and I am using my normal dev environment.
First I have built the container, which worked fine. but I had some problems...

1.
Then I tried `docker run -ti -p 8001:8001 --name kimai2 --rm $USER/kimai2` but couldn't access Kimai afterwards in the browser. Neither with 0.0.0.0:8001 nor 127.0.0.1:8001.
I rebuilt the image with the `CMD /opt/kimai/bin/console server:run 0.0.0.0:8080` to use the `ENV PORT=8080` and I tried to add `EXPOSE 8001` but nothing worked => still not able to access it in my browser. 

2.
Then I tried `docker run -ti -p 8001:8001 --name kimai2 -v $(pwd):/opt/kimai --rm $USER/kimai2` but that didn't start due to `Unable to write in the logs directory (/opt/kimai/var/log)`. 

3.
Then I tried `docker run $USER/kimai2` which started the webserver, but I couldn't stop it and nothing worked, no kill or background signal worked. I had to exit the bash session. Probably a Mac / docker-machine problem... any idea about that? 

4.
I would suggest we don't user $USER in the commands, as that makes support here more complicated if somebody asks here for help. Any objection to renaming the image to "kimai/kimai2"? Does that interfere with the published kimai2 image at docker hub?

Can you please help me, so I get it up and running and we can improve the docu for everyone else?

kevinpapst:
For 1. I found a solution:
Seems docker-machine on Mac is different from docker on Linux (what I used before). So with `docker-machine ip default` I got the IP which I then could use for the browser request. In my case I could access docker with `http://192.168.99.100:8001`.

var/docs/docker.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Build the docker:
+
+    docker build -t $USER/kimai2 .
+
+## Run the docker
+
+    docker run -ti -p 8001:8001 --name kimai2 --rm $USER/kimai2
+
+There is an admin user with the password admin but no fixtures installed. To install the fixtures:
```
kevinpapst: I think this is not true, as you don't execute the create-user command

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    git clone https://github.com/kevinpapst/kimai2.git /opt/kimai && \
+    sed "s/prod/dev/g" .env.dist > .env && \
+    composer install --dev --optimize-autoloader && \
+    bin/console doctrine:database:create && \
+    bin/console doctrine:schema:create && \
+    bin/console doctrine:migrations:version --add --all && \
+    bin/console cache:warmup && \
+    chown -R www-data:www-data var
+
+ENV PORT=8080
```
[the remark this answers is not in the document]
kevinpapst: Shall we add an `EXPOSE 8001` as well?

var/docs/docker.md
```diff
@@ -0,0 +1,52 @@
...
+
+    docker exec kimai2 bin/console kimai:reset-dev
```
kevinpapst: As these commands only work after an image was started, shall we explain the demonize `-d` flag?

Dockerfile
tobybatch: That's a legacy from an (old) alpine based image I used for a drupal 8 project.  The was a bug where the php extensions needed to build twice.  Not only has it been fixed in the fpm image we have swapped to debian/apache.  I've cleaned it up.

Dockerfile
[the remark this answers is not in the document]
tobybatch: You're right about the expose, we don't need the ENV at all as I'll elinated the need for a start up script.

We could use a build arg to set the port so one could change it when image was built.  Maybe a future improvement..

We'll add the expose as it's a cheap operation but I'm not sure when someone would use it.

tobybatch:
I've changed it to kimai/kimai2.  The only issue is a potential name space clash if they then try to pull our image over the top.

I guess if people are building their own image then they should realise this.

var/docs/docker.md
tobybatch: Yes.  But we are straying onto writing a docker howto.  I've added the words "once started" to the Running commands section but I think we need to assume a certain level of ability.

Dockerfile
kevinpapst: 👍

var/docs/docker.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Running commands in the docker
+
+You can run any command in the container in this fashion once it is started.  Add -ti to attach a terminal.
+
+    docker exec -ti kimai2 bash
+
+## Create a user and dummy data
+
+See the docs [here](installation.md) for full instructions, but this crates a user admin/admin with all privileges.
```
kevinpapst: Typo: creates

Dockerfile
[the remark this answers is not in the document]
tobybatch: The -p makes it available of the hosts network/ip.  I just wonder who will want another container to forward into this one directly.  Normally this is used to guard against network/port clashes.

So at the moment on my work station have 3 or 4 lamp stacks running in container clusters.  All the mysql's are listening on 3306 but in different networks, each isolated from each other by the docker daemon.

It costs us (relatively) nothing to expose the port.

var/docs/docker.md
```diff
@@ -0,0 +1,66 @@
...
+
+    docker build -t kimai/kimai2 .
+
+## Run the docker
+
+    docker run -ti -p 8001:8001 --name kimai2 --rm kimai/kimai2
+
+The you can then access the site on http://127.0.0.1:8001 or on Mac you may need to use the docker IP:
+
+    docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' kimai2
```
kevinpapst: This does not work on my Mac, only `docker-machine ip default` gives me the right solution. If you don't mind, I will fix the typos and add a section about Mac directly in your repo

tobybatch: @kevinpapst Please go ahead, I don't have a mac to test on :)

var/docs/docker.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Developing against the docker
+
+The following installs assume you have cloned the repo and opened a terminal in the root of the project.
+
+### Grab an initialized sqlite database
+
+If you already have a sqlite DB set up you can skip this step.
+
+    docker run -v /tmp:/tmp kimai/kimai2 cp /opt/kimai/var/data/kimai.sqlite /tmp
```
kevinpapst: I tried that, but wasn't able to use an already existing sqlite. Yes, this again is duplicating docker docu, but I'd like to explain clearly how it works. So assuming I have my existing SQLITE db file at `/home/kimai/kimai2/data.sqlite` what do I need to execute to use it in my docker image?
I am pretty sure: if we don't explain that now, there will be support requests... ;-)

Dockerfile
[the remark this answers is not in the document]
kevinpapst: Ok. I hope you don't mind all the questions. I just like to understand the stuff, so I am able to answer questions if somebody needs help.
Sounds as if we don't need EXPOSE at all, forwarding the Kimai UI doesn't really add any benefit. But as it doesn't hurt either.

var/docs/docker.md
tobybatch: I've changed the documentation to try and better explain what is happening.

var/docs/docker.md
kevinpapst: True. I just experienced time after time in all the years of managing open source projects, that people will execute stuff from the docu step-by-step, then something doesn't work and they come to ask here with the question instead of reading the docker docs. But you are right, lets wait and see what happens

Dockerfile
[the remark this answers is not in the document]
tobybatch: That's no problem.  This is your baby.

The only reason I could think of is that you may want to front the service with an nginx or apache to protect the php server that is running the service.  But as a dev image that's very unlikely.

The production version will address this.

var/docs/docker.md
kevinpapst: Yes: noob question, sorry again ^^ 
But what am I doing wrong here? Inside the container:
```
docker exec -ti kimai2 bash
www-data@48ba19368a12:/opt/kimai$ ls -la /tmp/
total 8
drwxrwxrwt  2 root root 4096 Oct 28 21:17 .
drwxr-xr-x 76 root root 4096 Oct 29 13:41 ..
www-data@48ba19368a12:/opt/kimai$ ls -la var/data/
total 860
drwxr-xr-x 2 www-data www-data   4096 Oct 28 21:17 .
drwxr-xr-x 8 www-data www-data   4096 Oct 29 13:41 ..
-rw-r--r-- 1 www-data www-data      0 Oct 28 21:17 .gitkeep
-rw-r--r-- 1 www-data www-data 102400 Oct 28 21:17 kimai.sqlite
-rw-r--r-- 1 www-data www-data 770048 Oct 28 21:17 kimai_test.sqlite
```
Outside the container:
```
ls -la var/data
total 8416
drwxr-xr-x   7 kevin  staff      224 29 Okt 14:40 .
drwxr-xr-x  10 kevin  staff      320 18 Sep 00:20 ..
-rw-r--r--   1 kevin  staff        0 12 Jan  2018 .gitkeep
-rw-r--r--   1 kevin  staff  1323008  2 Okt 21:43 kimai.sqlite
-rw-r--r--   1 kevin  staff   999424 25 Okt 23:48 kimai2.sqlite
-rw-r--r--   1 kevin  staff   266240  8 Sep 10:51 kimai_import.sqlite
-rw-r--r--   1 kevin  staff   770048 21 Sep 21:48 kimai_test.sqlite
```
Totally different files, so simply copying the SQLite file to the location outside the container won't affect Kimai inside the container (do I need a re-built in between?).

And I still don't get what that command is supposed to do: `docker run -v /tmp:/tmp kimai/kimai2 cp /opt/kimai/var/data/kimai.sqlite /tmp`
I tried it, but I couldn't see any changes inside and outside the box.

kevinpapst: I posted just before you committed ;-) Okay, I guess I do understand now what should happen, but it doesn't:
```
ls -la var/data
total 8416
drwxr-xr-x   7 kevin  staff      224 29 Okt 14:40 .
drwxr-xr-x  10 kevin  staff      320 18 Sep 00:20 ..
-rw-r--r--   1 kevin  staff        0 12 Jan  2018 .gitkeep
-rw-r--r--   1 kevin  staff  1323008  2 Okt 21:43 kimai.sqlite
-rw-r--r--   1 kevin  staff   999424 25 Okt 23:48 kimai2.sqlite
-rw-r--r--   1 kevin  staff   266240  8 Sep 10:51 kimai_import.sqlite
-rw-r--r--   1 kevin  staff   770048 21 Sep 21:48 kimai_test.sqlite

docker run -v /tmp:/tmp kevin/kimai2 cp var/data/kimai.sqlite /tmp

docker exec -ti kimai2 bash
www-data@48ba19368a12:/opt/kimai$ ls -la /tmp/
total 8
drwxrwxrwt  2 root root 4096 Oct 28 21:17 .
drwxr-xr-x 76 root root 4096 Oct 29 13:41 ..
www-data@48ba19368a12:/opt/kimai$ exit
exit

ls -la /tmp/
total 0
drwxrwxrwt    6 root   wheel   192 29 Okt 14:41 .
drwxr-xr-x    6 root   wheel   192  3 Okt  2017 ..
drwx------    4 kevin  wheel   128 28 Okt 22:30 .vbox-kevin-ipc
drwx------    3 kevin  wheel    96 17 Okt 07:55 com.apple.launchd.dNTaaolgeB
drwx------    3 kevin  wheel    96 17 Okt 07:55 com.apple.launchd.pyPDOHwTwk
drwxr-xr-x  287 kevin  wheel  9184 24 Okt 10:25 github-changelog-http-cache
```
Neither in the container nor on my host is the kimai.sqlite file in /tmp/ ?

tobybatch: I see now the gaps in my docs.

When you build or pull the image it creates an initialised sqlite file.  This has the schema and migrations run in it.  The run command starts a new instance image (a container) and copies the initialised DB to /tmp.  Because we had mount our tmp into the container when the container exists the db is still available on our host system. We then copy that into the project on the host system.

When we run the image as a developer we mount our local project folder over the one shipped in the image.  The sqlite db that we copied out is now available to the project using the php stack in the container.

Equally you should be able to pick any sqlite db from another kimai2 and copy it to var/data/kimai.sql and when you start the container it will be used.  This is all true as long as the local project id mounted over the one shipped in the container.

It all relies on the ```-v $(pwd):/opt/kimai``` to mount the local folder into the container at run time.

tobybatch: That's wierd:

```bash
tobias@tobias:kimai2 (docker=) $ ls /tmp/*.sqlite
ls: cannot access '/tmp/*.sqlite': No such file or directory
tobias@tobias:kimai2 (docker=) $ docker run -v /tmp:/tmp kimai/kimai2 cp /opt/kimai/var/data/kimai.sqlite /tmp
tobias@tobias:kimai2 (docker=) $ ls /tmp/*.sqlite
/tmp/kimai.sqlite
tobias@tobias:kimai2 (docker=) $ 
```

tobybatch: All those command are on my host machine, I'm never executing into an existing container.

I think you may have a running container. Check with a ```docker ps -a```

You may want ```docker system prune`` once you've stopped any old containers.  That cleans up stopped containers.

kevinpapst: Alright, I'll wait for your next ping

tobybatch:
I have radically simplified the docs, removing a lot of the advanced and hence confusing.  When I have written better docs in my repo I'll add a ling to them here.

For now the docs cover starting in a demo mode and starting with an existing DB and src tree.

You can see the tests here: 

https://github.com/tobybatch/kimai-docker

kevinpapst:
That looks much clearer to me. If someone know wants to use it for development, its easy to adjust to own needs (e.g. mounting the complete folder instead of src/ and the sqlite file only).
I like it. Will rephrase some smaller stuff and then merge it if you are ok with it.

tobybatch:
@kevinpapst That's good by me.  I'll keep working on a production docker and better docs in the other repo.  I'll ping you when there's something to see.

Thanks for your patients.

kevinpapst:
Thanks for your support, I really appreciate that! 👍 
It just gets better through discussions and with more contributor.

PR #382, opened 2018-10-31
Doc update of install/update instructions re: username

srdco:
## Description
Simple addition of notes regarding username to installation & updating instructions to cover use case of non-server installation. 

No code was changed - documentation update only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ Our roadmap is open for changes and input from the community, please [sent us](i
 
 ## Installation
 
+> **NOTE**
+>
+> sudo -u www-data (a command which grants the "www-data" user temporary administrator/super-user privileges) is intended primarily for server applications. If you are installing Kimai 2 on a desktop computer - even for use in a local network, but where the computer primarily serves as a single user computer - you will avoid permission errors by subsituting "www-data" in the relevant commands with your username (i.e. sudo -u myusername ...).
+
 There are [further infos about installation](var/docs/installation.md) if you have to use FTP or want to develop with Kimai. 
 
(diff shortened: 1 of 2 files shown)


kevinpapst:
Thanks for the suggestion. But I want to avoid blowing up the main README with more information. 
Could you please put a link instead of the new info text at these places and link to var/docs/installation.md
Then move your info text to that file in a new section about permissions.

README.md
```diff
@@ -118,6 +122,10 @@ Installation complete: enjoy time-tracking :-)
 
 ## Updating Kimai
 
+> **NOTE**
+> 
+> sudo -u www-data (a command which grants the "www-data" user temporary administrator/super-user privileges) is intended primarily for server applications. If you are installing Kimai 2 on a desktop computer - even for use in a local network, but where the computer primarily serves as a single user computer - you will avoid permission errors by subsituting "www-data" in the relevant commands with your username (i.e. sudo -u myusername ...).
```
kevinpapst: `sudo -u myusername` shouldn't be necessary, you could simply skip the sudo completely. Considering that you cloned via git with your own user, all files should already belong to it.
Maybe rephrase it to `subsituting "www-data" in the relevant commands with the username that runs the server`?

kevinpapst: Actually for updating I might be wrong, but for the first installation I am pretty sure.

And thats a big problem, normally these permission setups are so varying between installations that you won't be able to write something that works for everyone.

srdco:
Sure. Will do.

srdco:
Moved non-server install to separate section in install doc.

kevinpapst:
Thanks @srdco 👍

PR #387, opened 2018-11-03
updated the documentation, reflecting the changes in #306

lduer:
## Description
Missing documentation updates in the developers-section about the `kimai.invoice.documens` and removed parts for `renderer`, `calculator` and `number_generator`

## Types of changes

## Checklist
- [ ] My code follows the code style
- [ ] All files have a license header 
- [ ] All methods have a doc header with type declarations

### DIFF
--- a/var/docs/developers.md
+++ b/var/docs/developers.md
@@ -186,6 +186,30 @@ otherwise they might get lost during an update.
 
 In the config `kimai.invoice.documents`, you can add a list of directories with the locations of your invoice document templates ([read more](invoices.md)).
 
+## Adding invoice calculator
+
+An invoice calculator is a class extending `App\Invoice\CalculatorInterface` and is responsible for calculating 
+invoice rates, taxes and taking care of all timesheet entries that should be displayed.   
+
+Every invoice calculator class will be picked up when refreshing the application cache by the [InvoiceServiceCompilerPass](https://github.com/kevinpapst/kimai2/blob/master/src/DependencyInjection/Compiler/InvoiceServiceCompilerPass.php):
+
+The ID of the calculator must be unique, please prefix it with your vendor or bundle name and make sure it only contains
+character as it will be stored in a database column.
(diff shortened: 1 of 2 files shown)


kevinpapst:
Ups, seems as if I missed something ;-) Thanks for the notification!
Whileyour changes are true for the invoice renderer, the documentation for the calculator and number-generator is only partially wrong. They should not be deleted but updated. Could you add something like this:
```
## Adding invoice calculator

An invoice calculator is a class extending `App\Invoice\CalculatorInterface` and is responsible for calculating 
invoice rates, taxes and taking care of all timesheet entries that should be displayed.   

Every invoice calculator class will be picked up when refreshing the application cache by the [InvoiceServiceCompilerPass](https://github.com/kevinpapst/kimai2/blob/master/src/DependencyInjection/Compiler/InvoiceServiceCompilerPass.php):

The ID of the calculator must be unique, please prefix it with your vendor or bundle name and make sure it only contains
character as it will be stored in a database column.

## Adding invoice-number generator

An invoice-number generator is a class extending `App\Invoice\NumberGeneratorInterface` and its only task is to generate 
a number for the invoice. In most cases you do not want to mix multiple invoice-number generators throughout your invoices.   

Every invoice number-generator class will be picked up when refreshing the application cache by the [InvoiceServiceCompilerPass](https://github.com/kevinpapst/kimai2/blob/master/src/DependencyInjection/Compiler/InvoiceServiceCompilerPass.php):

The ID of the number generator must be unique, please prefix it with your vendor or bundle name and make sure it only contains
character as it will be stored in a database column.
```
Then I will merge the changes.

lduer:
I added your text to the developer docs and added some words to the translation docs about the new translation-files for the invoice-caclulator, the invoice number-generator and the invoice-documents.

kevinpapst:
Merci Lukas! Really appreciate your contribution 👍

PR #396, opened 2018-11-08
Added links to docker hub and my docker repo

tobybatch:
## Description

I've added direction for people with container questions to go straight to me.  I don't want you get bothered witch questions about the containers.

I'm going to invite the other docker authors to contribute to the container repo too.

## Types of changes

## Checklist

It's just documentation

### DIFF
--- a/var/docs/docker.md
+++ b/var/docs/docker.md
@@ -1,5 +1,7 @@
 # Docker
 
+There is an official docker repo for the kimai app, https://github.com/tobybatch/kimai2 and a dockerhub repo https://hub.docker.com/r/kimai/kimai2.  Please raise any container issues [there](https://github.com/tobybatch/kimai2/issues).
+
 We bundle a simple docker that can be used for test purposes.  Either pull it from docker hub or build it.
 
 ## Build the docker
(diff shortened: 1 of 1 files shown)


var/docs/docker.md
```diff
@@ -1,5 +1,7 @@
 # Docker
 
+There is an official docker repo for the kimai app, https://github.com/tobybatch/kimai2 and a dockerhub repo https://hub.docker.com/r/kimai/kimai2.  Please raise any container issues [there](https://github.com/tobybatch/kimai2/issues).
```
kevinpapst: Probably rephrase to something like `There is an official docker repo for Kimai 2 production usage`, just to make sure that the difference between these two docker images is clear for everyone who wasn't involved.

kevinpapst: Wait, its not only for production usage, but for dev also? Then it would probably be even more confusing what I wrote...

But could you please make proper links out of the URLs, so they work also when rendered directly in Kimai on the "Manual" page.

tobybatch: I've re-worded it and I think fixed the links as you want.

kevinpapst: Cool, thanks Toby!

Issue #397, opened 2018-11-08
Standard Docker Image booting but database not writeable


Haidy777:
**Describe the bug**
As described in #376 I tried to run the docker image. Sadly without luck. Today I tried again and used the `test.sqlite` file from the data directory. The DockerImage booted and it seems like it would be working, but the problem is that it seems that the database isn't writeable.

 @tobybatch Thanks for the idea with the file.

As before I'm trying to run the whole thing through Plesk, with the Docker Extension.
![image](https://user-images.githubusercontent.com/5011101/48222266-76a52700-e394-11e8-82d3-49d29a708c18.png)
As you can see there are several Port-Mappings, the working one is 8080 to 32772 which is then proxied through Plesk to a TLD.
Also you can see the Volume-Mapping at the bottom, which links the sqlite-File.

The Error:
![image](https://user-images.githubusercontent.com/5011101/48222387-ca177500-e394-11e8-8609-7c2f2bac1ca3.png)

I think it has to be something with the file permissions. `chmod`ing it to 777 works, but I don't think that should be the solution.

tobybatch:
@Haidy777 You may want to use the container that attaches to a SQL DB.  https://github.com/tobybatch/kimai2/tree/master/prod

kevinpapst:
Can you show us an shell output from the kimai directory: `ls -la var/data/`

Haidy777:
@tobybatch Thanks for your response. As it's only a testing setup I thought I'll give `sqlite` only a try.
It works when you chown the file to `www-data`. Not sure why I didn't test that :)

@kevinpapst I made two screenshots for you, the first one shows the host system, the second one the docker image.
![image](https://user-images.githubusercontent.com/5011101/48269385-bd982880-e437-11e8-8a5a-04c18c4dc2c2.png)
![image](https://user-images.githubusercontent.com/5011101/48269398-c688fa00-e437-11e8-84e6-b9c882d9be88.png)

Maybe this step should be added to the docs.

kevinpapst:
@Haidy777 It was added today: https://github.com/kevinpapst/kimai2/blob/master/var/docs/docker.md#developing-against-the-docker

@tobybatch I tested the ID 33 (as per new docu) on three systems. 

Mac OS:  
`uid=33(_appstore) gid=33(_appstore) groups=33(_appstore),12(everyone),61(localaccounts),701(com.apple.sharepoint.group.1),100(_lpoperator)`

[...]

PR #398, opened 2018-11-09
Fix for #397

tobybatch:
## Description

The sqlite DB needs be writable by the web user.

## Types of changes

### DIFF
--- a/var/docs/docker.md
+++ b/var/docs/docker.md
@@ -50,7 +50,7 @@ To install the fixtures:
 
 ## Developing against the docker
 
-It is possible to mount your source tree and sqlite DB into the container at run time.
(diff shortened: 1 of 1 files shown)


var/docs/docker.md
```diff
@@ -50,7 +50,7 @@ To install the fixtures:
 
 ## Developing against the docker
 
-It is possible to mount your source tree and sqlite DB into the container at run time.
+It is possible to mount your source tree and sqlite DB into the container at run time.  **N.B. The sqlite database needs to writable by the www-data user.** Use ```chown 33:33 /path/to/file.sqlite``` on the host machine.
```
kevinpapst: I'd prefer if the /path/to/file.sqlite example would either match the command example below `chown 33:33 $(pwd)/var/data/kimai.sqlite` or just `chown 33:33 kimai.sqlite`

tobybatch: I switched it to just kimai.sqlite as the other path does not match the error generated by Hay777

PR #435, opened 2018-11-22
Two-factor authentication

eljuanchosf:
## Description

Adding 2FA option via https://github.com/scheb/two-factor-bundle with options for email and Google Authenticator.
Still not working, PR per @kevinpapst request to discuss next steps.

## Types of changes

## Checklist
- [X] My code follows the code style
- [X] All files have a license header 
- [X] All methods have a doc header with type declarations


config/packages/security.yaml
```diff
@@ -17,6 +17,10 @@ security:
             anonymous: true
             stateless: false
 
+            two_factor:
+                auth_form_path: 2fa_login    # The route name you have used in the routes.yaml
```
kevinpapst: comments can be removed

config/packages/security.yaml
```diff
@@ -36,13 +40,15 @@ security:
                 path: fos_user_security_logout
                 target: homepage
 
+
```
kevinpapst: empty line can be removed

src/Entity/User.php
```diff
@@ -301,4 +302,31 @@ public function __toString()
     {
         return $this->getAlias() ?: $this->getUsername();
     }
+
+    /**
+     * @ORM\Column(name="googleAuthenticatorSecret", type="string", nullable=true)
+     */
+    private $googleAuthenticatorSecret;
```
kevinpapst: is that only for the google authenticator or can other products be used as well?

kevinpapst (changes requested):
The tests fail mainly because the test database doesn't have the new column. 
I pushed the updated database, but I guess there are still tests that need to be adapted.

lduer:
IMO this should not be part of the kimai-core (see #437) to follow the KISS principle.

Additionally, I've concerns of adding this, because of the different types of 2FA (TOTP - not only Google Authenticator, HOTP, E-Mail, 3rd Party Providers etc.)
If we start to add 2FA to the kimai core, where do we stop with implementing providers?

The two-factor-bundle wants to split the bundles into small bundles for each provider: https://github.com/scheb/two-factor-bundle/pull/148#issuecomment-410522213

src/Entity/User.php
lduer: I think it's copied from here https://github.com/scheb/two-factor-bundle/blob/master/Resources/doc/providers/google.md#basic-configuration and the `Scheb\TwoFactorBundle\Model\Google\TwoFactorInterface` is written for google authenticator. 

But as far as I can see, it is here to store TOTP/HOTP secrets. The use of other TOTP apps (e.g. https://github.com/andOTP/andOTP) should be possible too.

composer.json
```diff
@@ -29,6 +30,7 @@
         "symfony/monolog-bundle": "^3.1",
         "symfony/orm-pack": "^1.0",
         "symfony/profiler-pack": "^1.0",
+        "symfony/requirements-checker": "^1.1",
```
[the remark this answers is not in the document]
eljuanchosf: I actually didn't add it, I think it was an automated addition when I ran `composer check-platform-reqs`

composer.json
```diff
@@ -81,7 +83,8 @@
     "scripts": {
         "auto-scripts": {
             "cache:clear": "symfony-cmd",
-            "assets:install": "symfony-cmd"
+            "assets:install": "symfony-cmd",
+            "requirements-checker": "script"
```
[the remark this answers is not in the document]
eljuanchosf: Ok, thank you!! I think it was added automatically when used the requirement checker.

src/Entity/User.php
eljuanchosf: @lduer is right. I added this property and column per the bundle's instructions, but the bundle also support email 2fa as well.

config/packages/scheb_two_factor.yaml
```diff
@@ -0,0 +1,8 @@
+# See the configuration reference at https://github.com/scheb/two-factor-bundle/blob/master/Resources/doc/configuration.md
+scheb_two_factor:
+    security_tokens:
+        #- Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken
+        # If you're using guard-based authentication, you have to use this one:
+        - Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken
```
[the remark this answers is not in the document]
eljuanchosf: Fixed in da106761d80571b417f881ca40406559d09f4525

src/Entity/User.php
```diff
@@ -301,4 +302,31 @@ public function __toString()
...
+
+    // [...]
+
+    public function isGoogleAuthenticatorEnabled(): bool
+    {
+        return $this->googleAuthenticatorSecret ? true : false;
```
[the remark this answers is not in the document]
eljuanchosf: Fixed in f844300c8401e67c06b1988c56b8b2375376cab0

src/Migrations/Version20181120120000.php
```diff
@@ -0,0 +1,60 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace DoctrineMigrations;
+
+use App\Doctrine\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Initial database structure of Kimai 2.
```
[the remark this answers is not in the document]
eljuanchosf: Fixed in 623b6f9a4e4a4c14551409772ab07ee67c024be5 and e2ac657dfab8a3d60c25cb4c02569e78b87a895e

config/routes.yaml
```diff
@@ -4,3 +4,10 @@ home:
         _controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction
         route: homepage
         permanent: true
+2fa_login:
```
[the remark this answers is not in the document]
eljuanchosf: Fixed in e54efab18aa6ec4c4aaf30086e6dbd59f166629f

config/packages/security.yaml
eljuanchosf: Fixed in 9944d94d01421fd5ea527806a2d3654727c45af8

config/packages/security.yaml
eljuanchosf: Fixed in 9944d94d01421fd5ea527806a2d3654727c45af8

src/Entity/User.php
kevinpapst: then please rename the column accordingly, something like `twoFactorSecret`?

composer.json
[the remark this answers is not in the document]
kevinpapst: ok, but please remove it

src/Entity/User.php
eljuanchosf: @kevinpapst - I'm afraid not. The author of the bundle states in the docs that specific interfaces should be used per each authenticator module. However, per https://github.com/scheb/two-factor-bundle/blob/master/Resources/doc/providers/custom.md we could implement a generic one...

config/routes.yaml
```diff
@@ -3,4 +3,4 @@ home:
     defaults:
         _controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction
         route: homepage
-        permanent: true
+        permanent: true
```
kevinpapst: can you please revert that one, so the file ends with a newline

composer.json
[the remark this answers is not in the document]
kevinpapst: then please remove it

composer.json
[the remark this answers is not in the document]
kevinpapst: Thanks for fixing.

composer.lock
```diff
@@ -2685,16 +2752,16 @@
         },
         {
             "name": "phpoffice/phpspreadsheet",
-            "version": "1.4.1",
+            "version": "1.5.2",
```
kevinpapst: Could you please have a look at the composer.lock as well. Looks as if you executed a general composer update and all available package updates were pulled in. I like to do that in a seperate PR, to have the chance to review the updated packages (if they are directly used by Kimai) and to see if they changed anything important. 
So I'd like to see only changes in composer.lock which are dedicated to the 2FA packages and not stuff like Phpoffice or Doctrine.

kevinpapst: The simplest approach should be: revert composer.json / composer.lock / symfony.lock to their original state, then execute a `composer install` => the new packages should be replaced by the old ones. Then execute a `composer require scheb/two-factor-bundle`.

HeinzWuert:
Would'nt it be a good idea to include the newest way for authenfication WebAuthn?
https://www.golem.de/news/webauthn-standard-fuer-passwortloses-anmelden-verabschiedet-1903-139794.html

hebrides:
Could this be added as a plugin?

kevinpapst:
Thanks a lot for this PR @eljuanchosf - it was a great template for the two factor implementation that will finally ship with the next major release #2902

PR #466, opened 2018-12-05
support line breaks in docx

### LINKED ISSUE #464: Line breaks for address and payment information fields
I took a look at invoice.md and didn't find any information. At the moment line breaks in the mentioned text areas aren't considered in the document (docx in my case). I tried to add a html break but this broke my document I also tried markdown. Is there a way to insert line breaks in those fields?

For even more formatting options something like markdown support would be great. But for the moment I'd be happy to be able to insert simple line breaks.

kevinpapst:
Fixes #464


Sch-Tim:
Works great!

PR #528, opened 2019-01-29
Improve allowed running records

kevinpapst:
If the "create" action was triggered with a complete entry (end date also set) the check was executed as well and all data was lost. 

Any objection @lduer ?

### DIFF
--- a/src/API/TimesheetController.php
+++ b/src/API/TimesheetController.php
@@ -194,10 +194,13 @@ class TimesheetController extends BaseApiController
                 }
             }
 
-            $this->repository->stopActiveEntries(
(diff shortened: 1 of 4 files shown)


lduer:
OK. looks good. 

Maybe we need much more tests when adding entries on systems where `hard_limit=1`? E.g. inserting entries in the past, where already an entry exists. 

But that's part of another issue which I'll try to phrase as soon I've more time ;)

kevinpapst:
Sounds good! 
I want to move all these checks to timesheet validators, as the ruleset is already quite complex and spread around several places.

PR #538, opened 2019-02-01
added export module

kevinpapst:
Ignore the failing tests, I'll fix them later.

Do not use/test this branch in your existing installation, but only in a new dev environment - the database structure changed and that would be incompatible with your own installation.

You can also test this PR (while its open) at the following URL with the regular [demo accounts](https://v2.kimai.org/demo/):
https://demo-branch.kimai.org/


infeeeee:
Very nice, thank you, I installed the export-module branch. 

- The text in the mark all dialog is this: export.mark_all
- I see the summary in the pdf export, why it's only there? I think it should make sense in the print export as well
- It feels strange that the export buttons are below the table if I select a lot of entries. Of course I can push the End button on the keyboard, but it's a little bit inconvenient. 

But the module is very nice, thank you again

kevinpapst:
Thanks or your feedback @infeeeee !
I added the missing translation and the summary to the print view

blochwitz:
Hi Kevin,

is the export module still a part of the timezone branch?

kevinpapst:
@blochwitz no and it has never been, wach PR has its own branch

PR #540, opened 2019-02-03
Localized date-inputs and daterange-picker

kevinpapst:
- localized date input fields
- added daterange picker for toolbars


config/packages/kimai.yaml
```diff
@@ -97,38 +97,74 @@ kimai:
...
+            date: 'yyyy-MM-dd'
+            date_picker: 'YYYY-MM-DD'
             date_short: 'Y-m-d'
             duration: '%%h:%%m hours'
             duration_short: '%%h:%%m h'
         hu:
+            date_time: 'yyyy.MM.dd HH:mm'
+            date_time_picker: 'YYYY.MM.DD HH:mm'
+            date: 'yyyy.MM.dd'
+            date_picker: 'YYYY.MM.DD'
```
infeeeee:
```suggestion
            ddate_time: 'yyyy.MM.dd. HH:mm'
            date_time_picker: 'YYYY.MM.DD. HH:mm'
            date: 'yyyy.MM.dd.'
            date_picker: 'YYYY.MM.DD.'
```
In Hungarian we always put a dot after the number of the day. 
If you don't fix it now, I will fix it when I update the translation with the new words.

infeeeee:
Wow! really nice! I wanted to create a feature request for this, but you were quicker! It's very user friendly, thank you! I don't see any bugs.

Minor problem: I see that the start of the week depends on the selected language. If I change to english it's from sunday to saturday, in hungarian it's from monday to sunday. Is it possible to change this independently from the language? It can create confusion in a multi language environment, where the users use different languages. Maybe an option in the settings "start of week" or in a .env or similar file where I can select from "use locale" and "force monday/sunday". If it's not fixed it's ok, it's not a common feature, e.g. it's not  implemented in nextcloud neither. If it's implemented it should affect the calendar view as well.

kevinpapst:
Thanks for the test and format feedback @infeeeee

Yes, there is a setting for the first day of the week in the JS date-picker. I am not sure about the calendar, but every calendar component should support this.
Currently it is hardcoded to be 1 (Monday) in the datepicker. And the calendar seems to be bound to the language. I never use the calendar, so I don't recognize these things ^^

Having this setting in a config sounds as wrong as binding it to the users language. Kimai supports full locales as well, but unless we don't really need them, I'd like to avoid the configuration overhead.
As I'd really like to stick to KISS and avoid as many user settings as possible we stick with your approach - put it in a config file for now => can you please open a new issues for that?

config/packages/kimai.yaml
kevinpapst: There is a typo in `ddate_time` so I can't commit it directly here.

PR #542, opened 2019-02-07
Use Symfony formatter for currency symbol placement

sanjitlpatel:
## Description
Use symfony's formatCurrency to place the currency symbol in the correct location based on ISO-4217.

## Types of changes

## Checklist
- [Y] My code follows the code style
- [Y] All files have a license header 
- [Y] All methods have a doc header with type declarations
- [Y] I have updated the documentation accordingly
- [N] I have added tests to cover my changes


kevinpapst:
Thanks @sanjitlpatel 👍 I didn't knew that component.
I'll have a look at the tests, f you don't want to fix them.

sanjitlpatel:
I have the unit test update ready for this change.  If you wish, you can reject the pull request and I'll resubmit the unit test and the code change together.

kevinpapst:
You can push it to your `patch-1` branch and the PR will be updated, no need to close this one and open another one :-)

kevinpapst:
Almost done ;-) Here are some more errors: https://travis-ci.org/kevinpapst/kimai2/jobs/490197126

sanjitlpatel:
I'll fix it in 2 hours.

kevinpapst:
That looks very good.Much better then before!
But you brought something to my attention. Currently the money gets formatted by the users locale and therefor the output changes by the user locale (test cases for dollars as examples):
```
[expectedFormat, amount,              currency, locale],
['$13.75',                13.75,                  'USD',       'en'],
['13,75 $',               13.75,                  'USD',       'de'],
['1.234.567,89 $',  1234567.8912,   'USD',       'de'],
```
Sometimes the dollar sign is in front, sometimes behind. But this can't be fix that by now, as the users locale is the only available setting we can use to determine the formatting.

src/Twig/Extensions.php
```diff
@@ -254,15 +254,15 @@ public function money($amount, $currency = null)
 
         if ($this->locale !== $locale) {
             $this->locale = $locale;
-            $this->numberFormatter = new NumberFormatter($locale, NumberFormatter::DECIMAL);
+            $this->numberFormatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
         }
 
         $fractionDigits = Intl::getCurrencyBundle()->getFractionDigits($currency);
```
kevinpapst: I know you didn't write these lines ... but your PR made me curious and I had a look in `Symfony\Component\Intl\NumberFormatter\NumberFormatter` and it seems, as if there is already a `roundCurrency()` applied inside `formatCurrency()` so these lines could probably be deleted.
I need to do some testing with and without the original PHP intl extension.

sanjitlpatel:
I might not understand you correctly.
I believe the unit tests are consistent in the placement of the currency symbol on different OS platforms. It depends on the locale specified in the API, so the unit tests are reproducible.  However, what is not consistent is the currency symbol.  The following data in ExtensionTests.php passes on the github server but fails on my test machine.
['13,75 руб.', 13.75, 'RUB', 'ru']
on my test machine, this is the Ruble symbol that passes the test:
['13,75 ₽', 13.75, 'RUB', 'ru']

Are you experiencing different behavior in your tests?

src/Twig/Extensions.php
sanjitlpatel: The reason why I left the rounding behavior is because the money function defaults to $currency=null.

kevinpapst:
Well, actually I meant something else initially: there is still a flaw in the way Kimai formats the money, due to the fact that the users locale is used for the currency symbol placement (NumberFormatter gets the users locale as constructor parameter). 
Lets say: a German user prints an invoice for a US customer with 50 Dollar => the result would be `50 $` where it instead should be `$50`.
Don't get me wrong: with your changes it's better in most cases, but it can still be wrong for some (edge?) cases.

And about your test case, I had to change a couple of lines and will push a change later on.
I am having the same problem that you have and found another small flaw in the invoice currencies which I fixed as well.

kevinpapst:
Maybe the failing test is related to the outdated ICU version, see Travis output:
https://travis-ci.org/kevinpapst/kimai2/jobs/490556272#L913

Locally I get on my Mac:
```
intl

Internationalization support => enabled
version => 1.1.0
ICU version => 62.1
ICU Data version => 62.1
ICU TZData version => 2018e
ICU Unicode version => 11.0
```
I'll try to find a workaround.

kevinpapst:
Okay, fixed it. Test run locally and on Travis some are skipped (due to the old ICU version).

sanjitlpatel:
Thanks for the quick response, and sorry for all the mess.  I didn't think 2 lines of code change would uncover so much.
I now understand what you mean about user locale vs customer locale.  I think that opens another can of worms - if the customer is in the US and the user is in France, and the user has an hourly rate specified in Euros, when an invoice is printed for the customer, is there a currency conversion performed?
I also read your note on the ICU version.  Is this something that can be updated on Travis?

Have a good weekend!

kevinpapst:
I read some issues about ICU on Travis, seems the version is bound to their virtual-machine images. And its almost impossible to update or at least only with way to much overhead for this case. The Symfony Intl package has some checks like the one I added as well in their tests, thats where I got the inspiration from ;-) so it seems to be a common problem with Travis. 

Invoicing in Kimai doesn't have currency conversion and never was meant to handle such complicated cases. Its impossible as one-man show to get all these things correct, so I am really thankful for every help and input 👍 

Using this as is, as I don't want to open another "can of worms" for now ;-)

PR #547, opened 2019-02-08
fix daterange-picker for fr, hu and ar

### LINKED ISSUE #546: New time picker not working on languages without updated translation
**Describe the bug**
The new time picker not working with languages which didn't translated the new time picker yet. It works with english or german languages.

**To Reproduce**
1. Change the language to something which is not english or german
2. Go to My times or Export.
3. It shows the entries from the current month correctly as the default timeframe, but if you click on anything in the time picker nothing changes. Other filters are working as expected.

**Expected behavior**
It should work with translations which are not updated yet, with fallback to english.

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: chrome, firefox

**Smartphone (please complete the following information):**
I didn't check it on phone, but it should exist there.

**Additional context**
This bug is reproducible on the demo installation as well. 

P.S. Sorry for not finding this bug when I tested this before merge.

kevinpapst:
Fixes #546


infeeeee:
On arabic the year selectors are not working, on the left pane it shows something (I can't read arabic letters, but I assume it's the year), but on the right year selector is empty. The dropdown doesn't show anything on both year selectors.

Everything else seems ok for me.

kevinpapst:
Thanks, true. I just tested the one-click ranges.I never had any feedback to arabic besides the person who contributed the translations.
I guess I am unable to fix that right now, someone from the arabic dev community needs to contribute here in the future...

PR #549, opened 2019-02-08
scss fixes: year selector width, dropdown menu width, navbar refactoring

infeeeee:
## Description
Small scss fixes: 
- The width of the restart activities dropdown follows the content, checked on mobile, it's ok(max-width set) 
- Half of the last character of the time selector was hidden for me, now it doesn't cut that.
- I also refractored the navbar.scss, so it more looks like a proper scss file.

## Types of changes

## Checklist
- [ ] All methods have a doc header with type declarations


infeeeee:
It's on firefox 65 on linux. On windows, it was ok as well.

I created screenshots to demonstrate the problem: [imgur](https://imgur.com/a/9DaAODI). These images are from the demo install.
These images are from gnome on ubuntu, but on my main PC it's kde plasma on ubuntu, and the problem was the same.

I checked that this width:unset doesn't affect other browsers, so it solves this minor problem somehow

kevinpapst:
Okay, Firefox 65 here as well, but on Mac ... I'll test with a couple of browsers and merge afterwards.
Thanks 👍

PR #552, opened 2019-02-09
Daily stats in timesheet

### LINKED ISSUE #495: Visual Grouping of entry from same day or same week with some stats of that group
Not really. Just convenience.

In Kimai v1 on the timesheet, there have been black lines between the days. This is what I currently missing on kimai v2. 
On kimai v1 I always missed a micro-overview of any day or a week on the timesheet: Worked hours that day/that week in sum.

-

**Additional context**
-

### LINKED ISSUE #526: Better user expierence on "My Times"
Save Filter / or Button Today in "My Times" + Times entered today

Save the defined Filter in "My Times". Our users only want to see the current day and the total amount of hours for today

kevinpapst:
This PR adds a new user preference "Show daily stats in timesheet", which is deactivated b default.
If activated, the timesheet view will contain a visually highlighted statistic row between days

If a day has activities for multiple customers with different currencies, each currency will be summed up on ints own and displayed on a new line:
bildschirmfoto 2019-02-12 um 14 02 13

Fixes #495 
Maybe fixes #526


infeeeee:
- No summary for the day on the bottom of the page. If only one page needed to show the times, no summary for the first day of the timespan. If pagination needed and pagination cuts a day in half, no summary for the times on the first page, and the summary on the second page shows only the times on that page.
- Different currencies summarized to one value, it doesn't make sense. No exchange rates in kimai, and I think it's not a necessary feature, so maybe different currencies should be added separately, and show up as something like this: PHP594.18 + AZN1,139.60, or each on a new line

kevinpapst:
Thanks for the detailed report:
- Added stats for the last entry in each table
  - I can live with "no report for the day that is cut in half" - too complicated to work around that one. But due to the point above you get a report for the entries which can be seen.
- Different currencies: I can try to count each currency individually, but then is the question: how often does that happen in reality? When using the demo with the Data-Fixtures yes, but in real live?

infeeeee:
Currencies: Yes, it's not a very common use case, and if I filter by customer they can't show up, but I think even showing nothing would be better then adding them as one currency.

kevinpapst:
True, good argument. I will update the PR later on.

infeeeee:
It's much better now, and on a new line they look good, thank you!

PR #563, opened 2019-02-12
allow to query other users timesheets via api

### LINKED ISSUE #562: Export other users' timesheets using the API
I'm looking for a way to export timesheets of other / all users via the API.

Currently we are using Kimai 1 and import all timesheets into our invoicing system via an ODBC connection to the database on a regular basis. We are thinking about switching to Kimai 2 and start using the API, but I couldn't find a suitable method in the API. It looks like /api/timesheets can only access timesheets for the current user. Is there a way to get all timesheets or query by user id?

kevinpapst:
Fixes #562


mhaustein:
Looking great for me after a quick test. I tested specific ids, "all" and no parameter (defaulting to current user).

I noticed that the description is not included in the result of /api/timesheets. The description is very import for us, because that's what we send to our clients. Currently we'd have to query for every single timesheet entry (/api/timesheets/{id}) to get the description. Could you add it to the "TimesheetCollection" result?

kevinpapst:
Timesheet description is now available in collections as well.
The API docs have been updated to reflect the real type.

PR #603, opened 2019-02-26
Edit and create timesheet records in modal

kevinpapst:
Also fixed while working on it:
- [x] added loading indicator for datatables
- [x] fix multiple reloads of datatable content when using the customer/project toolbar filter

bildschirmfoto 2019-02-26 um 02 00 44
bildschirmfoto 2019-02-26 um 02 01 58

Someone around who could give this a try and help with ideas and input? 
@infeeeee @hmr-it-jr probably?


infeeeee:
Everything works correctly!
One small thing, the loading sign shows up on the center of the table vertically. If the length of the table is longer than the height of the screen * 2, the loading sign is offscreen.

kevinpapst:
Yep, I am not sure how to position it, so its always visible and doesn't look stupid. The only other solution I can come up with is to place it always at 100px below the start of the data-table. 
Something like that:
bildschirmfoto 2019-03-01 um 19 29 51
What do you think about that one?

hmr-it-jr:
@kevinpapst finally weekend and I had time to test it.

I want to add that we should use always border-radius for input or never:
grafik

Regarding the preloader problem:
frameworks I know handle this by not display the preloader in the specific table but fullscreen infront of everything - and then it's much easier to to center it.

infeeeee:
I was thinking the same about positioning, but 40vh or 30vh instead of 100px. In both cases it would be hard to write queries for this on mobile. The nicest solution would be with js, jquery has[ scrollTop()](https://api.jquery.com/scrollTop/), so the position can be calculated.

kevinpapst:
I played with `vw` and `vh` but that doesn't work with tables that have only a couple of entries, the loader would be positioned outside the table.
Positioning the loader above the complete content would make it easier to position it, but that might look weird, but I'll give it a try. And the next I will try is to use JS, the loader is already added via JS.

Regarding the border-radius: Actually I never saw that, but you are right, it should be consistent. Weird thing: it is already `border-radius: 0` for all elements, don't know why the selects still have rounded corners.

kevinpapst:
Okay, the loading indicator above the full content is the better solution. The horizontal alignment is not perfect, due to the sidebar. 
bildschirmfoto 2019-03-01 um 21 36 46

kevinpapst:
Btw: big thanks for your support 👍 It's so much easier and better to do stuff together!

infeeeee:
About the border-radius: on windows they show up correctly in firefox, chrome and edge. On android in firefox they look good as well.

kevinpapst:
**Mac OS here**

Firefox:
bildschirmfoto 2019-03-02 um 18 53 02
Chrome:
bildschirmfoto 2019-03-02 um 18 53 08
Safari:
bildschirmfoto 2019-03-02 um 18 53 18

The stack overflow reveals the solution:
https://stackoverflow.com/questions/39917000/reset-select-border-radius-on-chrome-mac
I'll add `-webkit-appearance: none;`  to the CSS, that fixes the problems in Chrome and Safari for me.

@hmr-it-jr what are you using?

hmr-it-jr:
@kevinpapst Firefox 65.0.2 on Windows 10

PR #631, opened 2019-03-07
fix random order in recent activities

### LINKED ISSUE #609: Some activites don't show up in last activities
**Describe the bug**
On my production server some projects don't show up on last activities dropdown. 
The database is originally imported from kimai 1. After I first discovered this bug I changed all of the imported activities to global activities. This was before the only 10 last activities update. This bug remained, and I still have some activities that missing from that dialog.
I reinstalled the whole server last week, and reimported the database, but some of them still don't show up there.
I can't find any pattern why they are not showing. They are not containing any special character. Some show up from the same customer, same activities show up for different customer etc, they are totally random.

My question is where should I look for reasons? If you want to check it, I can give access to my server, or send a mysqldump.

**To Reproduce**
I cannot reproduce this bug, exists only on my installation

**Additional context**
 - Kimai 0.8
 - PHP 7.2
 - mysql: 10.1.37-MariaDB-0+deb9u1
 - db charset: utf8mb4

kevinpapst:
Fixes #609

### DIFF
--- a/src/Repository/ActivityRepository.php
+++ b/src/Repository/ActivityRepository.php
@@ -20,9 +20,6 @@ use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Pagerfanta\Pagerfanta;
 
-/**
- * Class ActivityRepository
- */
 class ActivityRepository extends AbstractRepository
 {
     /**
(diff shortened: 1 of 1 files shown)


infeeeee:
It's fixed, thank you!

Issue #632, opened 2019-03-10
Installing kimai via Git-Remote-Repository


fatnmad:
Hi, 
I use Kimai since the beginning of time and love it. It saved my ass sooo many times 
... but I am so excited to use the latest version as it looks incredibly awesome to use.

As my Webspace-Provider now lets me use the [Git Repositorys](https://www.plesk.com/extensions/git/) via Plesk, I tried to install kimai2, but failed.

_But_ as it seems, it doesn't work as easy as that: My screen is left blank when I head to kimai2/public/

System: Linux, PHP 7.3.2, MySQL

kevinpapst:
I do not know what Plesk does. Did you follow the installation docs?

fatnmad:
Plesk installs the kimai2-git and "installs" it to the server.

as a sidenote: I also used the "FTP installation"-Guide without success

kevinpapst:
Plesk will very likely "only" clone the git repository. But It does not execute the installation steps.

> Did you follow the installation docs?

fatnmad:
> Did you follow the installation docs?

You mean the Installation-Docs for FTP installation? Yes. 
Still blank page :/

kevinpapst:
You said you installed it via Plesk & Git.
If you want my free help, you should start posting clear information: help without logs and understandable step-by-step instruction what you did is impossible.

fatnmad:
I am sorry, but your answer confused me... 

My first try was: using the git extension which clones kimai2 onto my webspace and that's it.
So my second step was: Using XAMPP and composer to locally install kimai2 (as in the installation doc for kimai).
Afterwards I re-uploaded all files and tried to log in / register through kimai2/public/ - but there is nothing happening. And there is also no log-file created or any message.

I am sorry, but that's all I have... Hope it's enough for your free help ;)

kevinpapst:
No server logs? No PHP logs? Nothing in var/logs/prod.log?
Did Kimai work locally?

kevinpapst:
Did you read the sentence?
> Kimai needs its own sub-domain or you need to recompile the frontend assets for usage in a sub-directory

fatnmad:
No server logs?
> it's a shared webhosting ...
No PHP logs?
> it's a shared webhosting ...
Nothing in var/logs/prod.log?
> no.
Did Kimai work locally?
> no.

fatnmad:
> No server logs?
it's a shared webhosting ... (
>No PHP logs?
[...]

Issue #494, opened 2019-01-10
FTP Installation Fatal Error


Regnalf:
I've  tried the FTP installation process because i have only a shared server.

Went through all the normal installation process, not dev!
I used mysql not sqlite, worked so far (databases and user are created).
Pointed the domain entry point to /public/ and got this error:

Warning: require(): open_basedir restriction in effect. File(/data/web/e24498/html/kimai2-gg6/public../../vendor/autoload.php) is not within the allowed path(s): (/data/web/e24498/html/kimai2-gg6/public:/data/web/e24498/phpsession:/data/web/e24498/phptmp:/tmp:/usr/bin) in /data/web/e24498/html/kimai2-gg6/public/index.php on line 9

Warning: require(/data/web/e24498/html/kimai2-gg6/public../../vendor/autoload.php): failed to open stream: Operation not permitted in /data/web/e24498/html/kimai2-gg6/public/index.php on line 9

Warning: require(): open_basedir restriction in effect. File(/data/web/e24498/html/kimai2-gg6/public../../vendor/autoload.php) is not within the allowed path(s): (/data/web/e24498/html/kimai2-gg6/public:/data/web/e24498/phpsession:/data/web/e24498/phptmp:/tmp:/usr/bin) in /data/web/e24498/html/kimai2-gg6/public/index.php on line 9

Warning: require(/data/web/e24498/html/kimai2-gg6/public../../vendor/autoload.php): failed to open stream: Operation not permitted in /data/web/e24498/html/kimai2-gg6/public/index.php on line 9

Fatal error: require(): Failed opening required '/data/web/e24498/html/kimai2-gg6/public../../vendor/autoload.php' (include_path='.') in /data/web/e24498/html/kimai2-gg6/public/index.php on line 9

kevinpapst:
Do you have the file `vendor/autoload.php`? 
Did you execute the `composer install` step?

Regnalf:
I did the "composer install" step and i do have the "autoload.php" file under /vendor

kevinpapst:
Oh and also: your poster seems to use the php setting `open_basedir` but limits access to files inside the document root, which is wrong. It seems that the allowed paths are:
`/data/web/e24498/html/kimai2-gg6/public:/data/web/e24498/phpsession:/data/web/e24498/phptmp:/tmp:/usr/bin`
[...]

Issue #469, opened 2018-12-06
setup for Synology or STRATO


HeinzWuert:
Anybody with more infos since the last trial?

Is an installation on Strato Webspace with MariaDB possible? (not Strato server!)

hmr-it-jr:
Kimai2 works perfectly on shared web hosting servers of netcup. So chances are high that it will work on shared hosting servers of strato, too.

kevinpapst:
@hmr-it-jr I guess the problem is, that there is no proper documentation on how to install it in a shared web hosting environment. I am using root server since more than 10 years, so I cannot share insights how to install Kimai and which problems might arise.

There are FTP installation steps available, but I wrote them down without testing, so if you have any input how Kimai can be installed or how the documentation can be improved, please let me know and I would happily add it to the docs!

hmr-it-jr:
@kevinpapst I followed the exact steps of your guide and it worked perfectly. The only thing I had to adjust was this:
![grafik](https://user-images.githubusercontent.com/36242595/52900414-b9796500-31f5-11e9-9e3c-3e78248c5081.png)

but I don't know if this has to be set on other webhosting providers.

kevinpapst:
Please see this issue #642 which adds a FTP package. 
There is a package for the upcoming version 0.9 linked which can be used for FTP installation.
I need feedback from FTP users, as I cannot test it properly.

kevinpapst:
https://www.kimai.org/documentation/synology.html

Issue #199, opened 2018-07-04
More simple deployment needed - one package


climb4fun:
Hi all,

what was a success factor of Kimai V1?
It's simplicity of deployment. Only PHP and MySQL. This was a reason for me to choose Kimai. I'm really a big fan of Kimai.

Due to constraints on my server, I cannot deploy Symfony; unless the fact I didn't have knowledge to do so.
I'd like to be an early adopter to V2, but due to the new deployment process, I will stay on V1 and hope it will survive as long as possible. It works fine on PHP 7.2.x.

If you have a more simple deployment - just dropping on the server and unzip - then I'll be able to migrated to V2.

Sorry to say that.

But thanks for Kimai V1 - it's a great piece of software.

Best regards
climb4fun

kevinpapst:
Hey @climb4fun!
Thanks for your feedback, I wasn't considering that it might be a problem for someone to install it like its mentioned in the documentation.
I can write some documentation on how to install Kimai if only FTP is available. But thats nothing I would recommend, unless you have no other choice. Its much much faster to install it with Git and composer.
Symfony is only a PHP Framework like Zend which is used is in Kimai 1.
Do you have bash access or only FTP available?

climb4fun:
Hi Kevin,

I only have FTP and it's behind a firewall. In my daily work it means, I can upload a file, unzip and then start a php installer. For sure I can run phpMyAdmin.

Best regards
Klaus (climb4fun)

NicklasAndersson:
It is really common for cheap web hosts to only provide ftp and some preinstalled tools.

kevinpapst:
Basically all you have to do in that case: executing the already documented installation steps on your local machine, then zipping the folder, uploading it and unzipping it on your server (or even uploading all files unpacked). 
That might not be obvious for unexperienced users, so I will write some documentation to make it clearer

climb4fun:
Hi Kevin,

much appreciated.

Regards
Klaus

kevinpapst:
I added instructions for "FTP only" installations: 
https://github.com/kevinpapst/kimai2/blob/master/var/docs/installation.md

I'd appreciate if one of you could try it out and give me some feedback, as I can't try it out myself. 
There is also a new ticket for a web-based installer, which will simplify this process in the future: #209

climb4fun:
Hi Kevin,
first of all thanks for your accommodation and support.
[...]

PR #642, opened 2019-03-13
added command to create release packages (for FTP user)

kevinpapst:
Added a command used to create a release package with pre-installed composer, SQLite database and user.

**The package**
The contents of this new package (zip) can be uploaded via FTP and should work out-of-the-box.
This will NOT replace the recommended installation via SSH/git/composer but it can help the people with shared-webhosting.

If you want to help testing, I uploaded a package for `v0.9_dev` to Google drive (~ 81,5 MB): 
https://drive.google.com/open?id=1zbt-X7IkboOTVvzF8fUxj3YnBFsJWAR5

By default it uses a SQLite database, but you should reconfigure your DB connection in `.env` before uploading the files.

:exclamation: ATTENTION :exclamation: You have to point your Webserver to /public/ directory!!!!!!!!
If you see `public/` in the URL do NOT use Kimai, this is a huge security risk, as your database and credentials can easily be downloaded.

**Maintainer only**
This new command is not meant to be used by end-users, but for maintainer only.
The created package could be uploaded to GitHub, to simplify installation process for the "shared Webhosting users" without SSH, Git ... 

Possible commands could looks like this:
```
bin/console kimai:create-release --release 0.8.1 --directory /tmp/
bin/console kimai:create-release --release master
```

**TODO**
- [ ] Write installation documentation

Related to:
- #632
- #494
- #469 
- #199


NicklasAndersson:
Tried to download the file but did not manage to. Perhaps you could upload it to github?

kevinpapst:
@NicklasAndersson  Unfortunately not, thats what I tried first... GitHub allows 10MB max (unless its a release, what its not).
For the G-Drive download: you have to wait quite a while until you get an error message saying something like "the file could not be scanned for viruses, really want to download?" - then you can download it.

NicklasAndersson:
I have tried the package now. Seems to work great, the only thing i did not like is that it requires a "special" web root setup in apache (on most webhosts).

https://kimai.livsfiloso.fi if you want to do some testing I could setup an account for you

fatnmad:
@kevinpapst thank you so much! 

1. Uploaded zip to web-dir (via plesk)
2. Unpacked (via plesk)
3. Redirected subdomain to /public/
4. Works!

Anything I need to do afterwards? (deleting anything or so?)

kevinpapst:
Thanks for the feedback @NicklasAndersson and @fatnmad 

No cleanup required, but I would highly recommend, that you change your database from SQLite (which is used out of the box) to Mysql => this can be configured in `.env`.

fatnmad:
> No cleanup required, but I would highly recommend, that you change your database from SQLite (which is used out of the box) to Mysql => this can be configured in .env.

sadly it won't work after returning to mysql :/ 
`Authentication request could not be processed due to a system problem.`

kevinpapst:
Only changing the connection string is not enough:
https://www.kimai.org/documentation/installation.html#use-mysql-database
But I see that this documentation is not complete. You would have to change the .env locally and run the `bin/console doctrine:schema:create` command against your hosted database (which requires your DB account to have connection permissions from %).

climb4fun:
Hi Kevin,
first let me thank for this special version. Here my first test results for setup:
1. Setup with SQLite works perfect.
2. Setup with MySQL does not work "500: Internal Server Error". May be a wrong entry from my side in '.env'. MySQL is important for me, have a long history of time recording, which should be migrated to MySQL in Kimai V2. I didn't done what you wrote above for setting up database.
First feedback on application:
1. It is very simple and intuitive. Well done in in every respect.
2. I would like to have lines in Timesheet more narrow like in Kimai V1. I have to set whole month in one screen.
3. I need durations in decimal view - as was in Kimai V1.
Thanks for your outstanding support.
Best
Klaus

Issue #692, opened 2019-04-10
Add floor and ceil rounding


simone-gasparini:
Add floor and ceil rounding to round all up or down.
PR will follow.

PR #695, opened 2019-04-10
added floor and ceil rounding with tests (#692)

simone-gasparini:
## Description
Added floor and ceil rounding with tests

## Types of changes

## Checklist
- [x] I have executed `bin/console kimai:phpcs --fix` to make sure my changes adopt the correct code style

### DIFF
--- a/tests/Timesheet/Rounding/CeilRoundingTest.php
+++ b/tests/Timesheet/Rounding/CeilRoundingTest.php
@@ -1,0 +1,152 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Tests\Timesheet\Calculator;
+
+use App\Entity\Timesheet;
+use App\Timesheet\Rounding\CeilRounding;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \App\Timesheet\Rounding\CeilRounding
+ */
+class CeilRoundingTest extends TestCase
+{
+    /**
+     * @dataProvider getTestData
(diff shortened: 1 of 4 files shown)


tests/Timesheet/Rounding/CeilRoundingTest.php
```diff
@@ -0,0 +1,152 @@
...
+        $record->setDuration($record->getEnd()->getTimestamp() - $record->getBegin()->getTimestamp());
+
+        $sut = new CeilRounding();
+        $sut->roundBegin($record, $roundBegin);
+        $sut->roundEnd($record, $roundEnd);
+        $record->setDuration($record->getEnd()->getTimestamp() - $record->getBegin()->getTimestamp());
+        $sut->roundDuration($record, $roundDuration);
+
+        $this->assertEquals($expectedStart->getTimestamp(), $record->getBegin()->getTimestamp());
+//        $this->assertEquals($expectedEnd->getTimestamp(), $record->getEnd()->getTimestamp());
```
kevinpapst: why are these two lines commented?

simone-gasparini: My bad.. I fix it

simone-gasparini:
Added documentation PR
https://github.com/kimai/www.kimai.org/pull/24

simone-gasparini:
> Thanks for your PR! 👍
> 
> The new modes should be added that to the documentation:
> https://github.com/kimai/www.kimai.org/blob/master/_documentation/timesheet.md#rounding-of-begin-end-and-duration
> 
> I wonder why the tests don't have different numbers. There is something wrong, either the tests (what I guess) or the rounding mode.

The tests are based on the closest one, but with the correct results for the new implemented rounding

PR #703, opened 2019-04-13
Load export type ids dynamically

### LINKED ISSUE #702: Review export lifecycle for plugin injection
Add ExportSubscriber to dispatch event for plugin based export module.
PR will follow.
Help needed 😄

simone-gasparini:
## Description
Load export type ids dynamically

Fixes #702 

## Types of changes

## Checklist
- [x] I have executed `bin/console kimai:phpcs --fix` to make sure my changes adopt the correct code style


kevinpapst:
Fixing a test by deleting it is the worst possible solution. Can you please try to find tests for the new logic and check if you can fix the old one.

simone-gasparini:
I've restored the old test and added a foreach to add types. For the new implementation i think that the existing test it's already ok.
https://github.com/kevinpapst/kimai2/blob/master/tests/Controller/ExportControllerTest.php#L59

src/Controller/ExportController.php
```diff
@@ -44,6 +49,21 @@ public function __construct(TimesheetRepository $timesheet, ServiceExport $expor
     {
         $this->timesheetRepository = $timesheet;
         $this->export = $export;
+        $this->query = $this->createExportQueryWithExports();
+    }
+
+    /**
+     * @return ExportQuery
+     */
+    protected function createExportQueryWithExports()
```
kevinpapst: I am not completely happy with this method being in the controller.
Maybe it could be moved as factory function to the ServiceExport. Registering the Query as a service could work as well and then registering the renderer type in a compiler pass.

I'll give it another thought and maybe push a change later on.

kevinpapst:
I decided to remove the type validation from the query object, as it doesn't belong there.
And there was already another check in place in the controller... will merge now.

PR #752, opened 2019-05-03
dynamically update title and visible durations for running records

### LINKED ISSUE #420: Refresh of Duration time
There is no problem/bug - just a nice feature from my point of view, like it was with kimai v1.

Like in kimai v1 it would be great if the duration time will be updated automatically. Currently, using Firefox it will not be updated until I use the Browser refresh button. I also like the feature we have with kimai v1 that the duration time was shown in the Tab Label of Firefox. If this is not possible, a refresh button behind the stop,delete,edit button would be great.

Regards,

Ralf

### LINKED ISSUE #660: Show on the title if there is an active recording
In kimai1 on the titlebar of the website the current counter was visible. I want a similar feature, so if there are a lot of tabs open I can see if I started the timer or not.

In kimai1 a user could have only 1 active record, so it was straightforward which counter to show. In kimai 2 I recommend a static sign instead: the [play button emoji ](https://emojipedia.org/black-right-pointing-triangle/)looks cool in the titlebar: ▶ 
image
This should only appear if at least one recording is active, and disappear if no active.

I think showing a counter would be harder to implement, and personally I didn't like it, I just liked that it's visible that the recording is going. There would be a problem if there are multiple recordings active, which one to show?
Changing the favicon to an other color would an other option, I don't know if it's possible. But I like the emoji better.

kevinpapst:
## Description

Added a javascript which will update all durations for running records:
- on the browser title
- on the "active entries" dropdown

Fixes #420 
Fixes #660

**TODO** 
- [x] update timesheet tables
- [x] limit title to the first 3 entries

## Types of changes

## Checklist


infeeeee:
Very nice, thank you. My notes on this:

- One digit mins and hours show up without an additional 0. On 'my times' 2 hours and 2 minutes looks like this: 02:02. On the title it looks like this: 2:2. This is the same for the active records, however there the 0s disappear when the time shows up on the title.
- I don't really like the delay when the timer shows up, but I can live with it, I understand the reason(to show the name 'kimai' in the titlebar?) , it's just strange, maybe because on kimai1 the timer started immediately on the title. Maybe replace the 'time tracking' text with the clock, so it will look like this: 
image

- Should they update automatically on the my times as well? Active records updates, titlebar updates, but my times only when I reload: 
image
- Small bug: If I change the start time of a running timer, the times update on My times, however on the active recordings and title still shows the old timer started from the original start time. This is also true if I delete an active recording from 'my time'.

kevinpapst:
Hey, thanks for the quick and detailed feedback - awesome!

No, I don't care about "Kimai in the title" at the beginning.
I just registered a script which runs every 10 seconds, thats where the delay came from. 

I knew about the missing zero and pushed the following changes, if you want to do a re-test:
- update all durations and title as soon as the page is loaded
- added leading zero to duration units
- added duration updates on timesheet tables
- show a max amount of 3 records in title
- show only the own durations in title

infeeeee:
The title and the active records don't update when I modify the start of a running record or delete a running record, because the page doesn't reload like when I stop it. But it's a rare use case.

Everything else is superb!

kevinpapst:
Ah, forgot about that... I'll have a look.

kevinpapst:
Okay, I didn't read properly. This is about the "active entries" dropdown ... hmmmm, the bug was there before, but now its much more visible. 
The information in the title is taken from the dropdown, so its not refreshed after updating an entry. 

I want to stop entries in the future without reloading the page and therefor this dropdown needs to be rewritten to use the API. But thats a little bit more work and I don't want to mix it with this issue. I will write a new issue for that.

PR #757, opened 2019-05-04
added more entity actions

bjoerne2:
Adds actions to data records resp. table rows

Customers
* Activitites
* Timesheet
* Create project

Projects
* Timesheet
* Create activity

Activities
* Timesheet
* Create timesheet

Timesheet and My times
* Equal actions:
  * Stop (if running)
  * Edit
  * Start again (if finished)


src/Controller/ActivityController.php
```diff
@@ -82,7 +83,13 @@ public function indexAction($page, Request $request)
      */
     public function createAction(Request $request)
```
kevinpapst: Please use autowiring:
```
public function createAction(Request $request, ProjectRepository $repository) {
...
}
```
or even better via an additional route:
```
   /**
     * @Route(path="/create/{id}", name="admin_activity_create_with_project", methods={"GET", "POST"})
     */
    public function createAction(Request $request, ?Project $project = null)
```

src/Controller/ProjectController.php
```diff
@@ -83,7 +83,13 @@ public function indexAction($page, Request $request)
      */
     public function createAction(Request $request)
```
kevinpapst: same here

templates/macros/actions.html.twig
```diff
@@ -211,6 +226,10 @@
             {% set actions = actions|merge({'edit': {'url': path('timesheet_edit', {'id': timesheet.id}), 'class': class}}) %}
         {% endif %}
 
+        {% if timesheet.end and is_granted('start', timesheet) %}
```
kevinpapst: I like the shorter if, but I would move this back above edit, as edit is already available via the table row click

templates/macros/actions.html.twig
```diff
@@ -258,8 +281,8 @@
             {% set actions = actions|merge({'edit': {'url': path('admin_timesheet_edit', {'id': timesheet.id}), 'class': class}}) %}
         {% endif %}
 
-        {% if not timesheet.end and is_granted('stop', timesheet) %}
-            {% set actions = actions|merge({'stop': path('admin_timesheet_stop', {'id' : timesheet.id})}) %}
+        {% if timesheet.end and is_granted('start', timesheet) %}
```
kevinpapst: Why would you restart a timesheet from the team view?

translations/actions.de.xliff
```diff
@@ -46,6 +46,18 @@
...
+                <source>create-project</source>
+                <target>Projekt erstellen</target>
+            </trans-unit>
+            <trans-unit id="create-activity">
+                <source>create-activity</source>
+                <target>Tätigkeit erstellen</target>
+            </trans-unit>
+            <trans-unit id="create-timesheet">
+                <source>create-timesheet</source>
+                <target>Zeiterfassung erstellen</target>
```
kevinpapst: Sounds strange ;-)

templates/macros/actions.html.twig
```diff
@@ -23,6 +23,12 @@
             {% endif %}
             {% set actions = actions|merge({'edit': {'url': path('admin_activity_edit', {'id': activity.id}), 'class': class}}) %}
         {% endif %}
+        {% if is_granted('view_other_timesheet') %}
+            {% set actions = actions|merge({'timesheet': path('admin_timesheet', {'customer': activity.project ? activity.project.customer.id : null, 'project': activity.project ? activity.project.id : null, 'activity': activity.id})}) %}
```
kevinpapst: what happens to the URL if you pass null for a parameter?

src/Controller/TimesheetControllerTrait.php
```diff
@@ -158,6 +160,16 @@ protected function create(Request $request, $redirectRoute, $renderTemplate)
             }
         }
 
+        if ($request->query->get('project')) {
```
kevinpapst: I would say same here, but I am not sure if this would lead to code duplication?!?

translations/actions.de.xliff
bjoerne2: @kevinpapst True :) "Zeit erfassen" ? Other suggestions?

src/Controller/ActivityController.php
bjoerne2: Thanks for this suggestion! I autowired the repositories, but decided against a different route.

src/Controller/ProjectController.php
bjoerne2: Thanks!

templates/macros/actions.html.twig
bjoerne2: @kevinpapst Would it be OK to restart from the team view if the user of the last timesheet entry is the current user? That would be useful for single user scenarios when freelancers are using kimai just for themselves (that's my case). In such a scenario there is no difference between "My times" and the team view.

templates/macros/actions.html.twig
bjoerne2: @kevinpapst As far as I tested it the parameter is ignored and the URL doesn't contain it.

templates/macros/actions.html.twig
kevinpapst: Having one more action item there doesn't hurt. I am just questioning if it is necessary here, but then "why not".

translations/actions.de.xliff
kevinpapst: Zeit erfassen 👍

templates/macros/actions.html.twig
bjoerne2: OK

translations/actions.de.xliff
bjoerne2: Done

kevinpapst:
@bjoerne2 please update your branch, I pushed a change and merged the latest master.
I find the route approach much cleaner, so I changed that for activities and project. 
And I added some checks for the visibility, to hide the `create` buttons for "deactive/invisble" customer and projects.

kevinpapst:
Thanks @bjoerne2 for your contribution 👍

PR #798, opened 2019-05-20
lock exported timesheets

### LINKED ISSUE #686: Lock exported objects
As a "manager", i want to lock certain time objects from editing, as they were exported (or invoiced to a client) and edit should be prohibited.

We wold like to have an option to disable editing of time entries that are exported. Because from the moment we created an invoice (or we export the time entries to another system), any change in entries results in differences between the invoice or the information in the 3rd party system.

There are no alternatives.

**Additional context**
None provided.

kevinpapst:
## Description

Exported records are locked to prevent further updates by regular users.
This limitation can be circumvented by users with the new permission `edit_exported_timesheet`, which by default is given to `ROLE_ADMIN` and `ROLE_SUPER_ADMIN`.

These users will see a warning instead:
Bildschirmfoto 2019-05-20 um 14 29 47

A new API method was added to switch the export state of timesheet records. 

Fixes #686

## Testing

Can be tested at http://demo-branch.kimai.org/ 

Please try it with both an admin and a regular user. The regular user cannot switch the "exported" state, so you have to either use the export screen or switch it manually in the timesheet administration with the admin user.
Regular user should not be able to edit the exported timesheets.
Admin user should see a warning.

## Types of changes

## Checklist
- [x] I ran `bin/console kimai:codestyle --fix` to verify the correct code style
- [x] I have updated the [documentation](https://github.com/kimai/www.kimai.org/tree/master/_documentation) accordingly


hmr-it-jr:
I've tested it.

So for exported objects the user can't edit it since there is not edit button:
grafik

But why can the user still delete the entry? This doesn't make sense to me. If it's exported it should be read only for the user - except he has special privilegs.

kevinpapst:
Pretty good argument, I simply did not think about it at all :-D
I think I will re-use the permission `edit_exported_timesheet` or shall we create another permission for deletion?

hmr-it-jr:
I think reusing

> edit_exported_timesheet

would be the best.

kevinpapst:
added "prevent deletion of exported timesheets"

PR #810, opened 2019-05-25
enhance tag query to show all tags of an item

### LINKED ISSUE #769: Additional status indicators for a field (tag field) more like lables which are globally assigable and appear for each task a user is working on or has started
Currently the tags filed does not exist, if it is created an option to add label to it can really help. In some cases certain tasks the user is working on has to be prioritied or in some cases have to be stopped. Having an option of a label can meet this requirement. Where one label can indicate something an another something else. If its labelled as priority the user can try to complete it faster and so on.

Ability to add a label to a tag. It more like additional information for a field, or actions to be performed for a value in a tag. Something more like that is present in git.

We use the desciption field to update the tag value and then export the data. Post which we identify the user working on the task and inform them via email.

**Additional context**

kevinpapst:
## Description

- enhance tag query to show all tags of an item, even if not queried

Fixes #769 

## Types of changes

## Checklist

### DIFF
--- a/src/Controller/TimesheetAbstractController.php
+++ b/src/Controller/TimesheetAbstractController.php
@@ -84,10 +84,11 @@ abstract class TimesheetAbstractController extends AbstractController
             $query->setUser($this->getUser());
         }
 
-        if ($query->hasTags()) {
+        $tags = $query->getTags(true);
+        if (!empty($tags)) {
             $query->setTags(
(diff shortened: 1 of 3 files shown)


rarvin:
I have tested this and could see that even when i query with one tag name all tags against the said timer task are listed. The functionality is working well, thanks for this.

PR #815, opened 2019-05-27
add ldap support

### LINKED ISSUE #193: Add Ldap support
Hi
It would be nice to have Ldap support for the app.

### LINKED ISSUE #449: LDAP - Active Directory
Is there a integration to log in using AD users on kimai2? Like kimai1?
Or a doc explaining how to use.

Ty.

### LINKED ISSUE #795: connect kimai to ldap
hi.

how i can connect kimai v2 to ldap?

kevinpapst:
## Description

Documentation can be found at: https://www.kimai.org/documentation/ldap.html

Features:
- Authenticate against an LDAP
- New users will be created on the fly
- Configured attributes will be synced on every login
- Mapping LDAP groups to Kimai roles is possible

**TODO**
- [x] prevent setting an internal password
- [x] get rid of the service class name parameter
- [x] convert ldap container parameter to config class
- [x] add group mapping query

**How to test**
- checkout the branch
- `composer install`
- adjust `config/packages/local.yaml` to your needs (see [documentation](https://www.kimai.org/documentation/ldap.html))
- `bin/console cache:clear`

**Open questions for EVERY reader**

- How do other apps handle LDAP support?
- Should an app still offer registration when LDAP is activated? 
  - Leave that configurable or turn it off always?
  - I guess we need to avoid problems like a local account with the same account name in LDAP, then the LDAP account would never be checked?
- Should groups be transferred?
  - How? Is there a general approach?
  - I read something about a "memberof" field/attribute which seems to only a OpenLDAP feature?
- The login needs to be checked against LDAP every time?
- What should happen to the local account if an account in LDAP is deleted?
- Can LDAP accounts be deactivated?

If anything else comes to your mind, please share it. This is one of the features I have to develop without having a proper LDAP setup/knowledge, so I am "flying blindly" here and need YOUR input. 

Fixes #193 
Fixes #795
Fixes #449 

## Types of changes

## Checklist


army1349:
Thanks for your work on this, this is my view on your questions:
>     * How do other apps handle LDAP support?
Some apps I use with LDAP:
[Nextcloud](https://docs.nextcloud.com/server/stable/admin_manual/configuration_user/user_auth_ldap.html)
[MediaWiki](https://www.mediawiki.org/wiki/Extension:LDAP_Authentication/Examples)
Support of multiple servers and StartTLS/LDAPS are always good to have.
>     * Should an app still offer registration when LDAP is activated?
>       * Leave that configurable or turn it off always?
>       * I guess we need to avoid problems like a local account with the same account name in LDAP, then the LDAP account would never be checked?
Probably no. Local accounts coexisting with LDAP accounts are with exception of some kind of admin account uncommon.
>     * Should groups be transferred?
If groups are important part of an app, then it is always nice to have support for LDAP groups.
>       * How? Is there a general approach?
>       * I read something about a "memberof" field/attribute which seems to only a OpenLDAP feature?
People which want to use LDAP generally know how to use it, so I would let them configure it. Group base, group name attribute, group search filter, membership attribute and optionally reverse membership attribute (memberof) could do it. Example:
`group base: ou=groups,dc=example,dc=com`
`name attribute: cn`
`group search filter: (objectClass=groupOfNames)`
`member attribute: member`
`reverse member attribute: memberOf`
>     * The login needs to be checked against LDAP every time?
Yes, correct way is to try to bind to LDAP with user credentials.
>     * What should happen to the local account if an account in LDAP is deleted?
Some kind of auto cleanup option is nice to have feature, but not without confirmation, or time delay, or other foolproof mechanism.
>     * Can LDAP accounts be deactivated?
Again, I would leave that to admin. User search filter should do it all. Example:
`(&(objectClass=posixAccount)(memberof=cn=kimai,ou=groups,dc=example,dc=com))`

kevinpapst:
Thanks @army1349 for your input!
After more digging and thinking about your input, that leaves me with just one minor open point, before we can start testing a first version:
- transfer the permission from LDAP to Kimai (some kind of group mapping, still unsure how)

Some feedback to your points:
- deleting user: won't do anything here, normally you don't want to delete timesheet records for users how left the company
- users who were imported via LDAP are flagged (empty password, DN is cached) and login attempts will be rejected if they can't be found in the LDAP (AKA deactivated)
- duplicate local vs ldap accounts: won't do anything here, as this is up to the admin to handle (worst case scenario: rename the user in the DB)
- SSL and StartTLS seem to be supported, even though I don't know how to test it locally.

A couple of things can't be added easily, as I am using another library for the groundwork. Especially the group / role transfer for permissions is unclear to me. What I can provide as solution, is mapping LDAP attributes to the user and from there to Kimai groups. 
For example if a user has the field `groupId` with value `13` I could offer a configurable mapping `13: ROLE_ADMIN`. Would something like that help?

army1349:
Glad I can help.
> A couple of things can't be added easily, as I am using another library for the groundwork. Especially the group / role transfer for permissions is unclear to me. What I can provide as solution, is mapping LDAP attributes to the user and from there to Kimai groups.
> For example if a user has the field `groupId` with value `13` I could offer a configurable mapping `13: ROLE_ADMIN`. Would something like that help?
I would not use groupId or gid attribute. Let's split this problem into few smaller ones.
1. Which LDAP groups should be visible/added?
Admin could specify that like this:
`group base: ou=groups,dc=example,dc=com`
`name attribute: cn`
`group search filter: (&(objectClass=groupOfNames)(|(cn=myKimaiGroup1...)(cn=myKimaiGroup2...)))`
2. Who is member of this groups?
- If admin can specify membership attribute like:
'member attribute: member'
then you can query group for that membership attribute and you will get all members.
- If admin can specify reverse membership attribute like:
`reverse-member attribute: memberOf`
then you can query user for that reverse membership attribute and you will get all groups, which user is member of.

kevinpapst:
@army1349 Could you maybe setup and share such an example configuration as LDIF file? I am missing the knowledge how to administrate an LDAP server or setup such a structure. 
But at least I am able to import the data and build the code around it ;-)

army1349:
Here is simple example.
I use OpenLDAP with _memberOf_ overlay, so _memberOf_ attributes are maintained automatically according to member attributes of groups.
```
cn=group1,ou=groups,dc=example,dc=com:
objectClass: groupOfNames
member: uid=user1,ou=users,dc=example,dc=com
member: uid=user2,ou=users,dc=example,dc=com
...

cn=group2,ou=groups,dc=example,dc=com:
objectClass: groupOfNames
member: uid=user2,ou=users,dc=example,dc=com
member: uid=user3,ou=users,dc=example,dc=com
...

uid=user1,ou=users,dc=example,dc=com:
objectClass: inetOrgPerson
memberOf: cn=group1,ou=groups,dc=example,dc=com
...

uid=user2,ou=users,dc=example,dc=com:
objectClass: inetOrgPerson
memberOf: cn=group1,ou=groups,dc=example,dc=com
memberOf: cn=group2,ou=groups,dc=example,dc=com
...

uid=user3,ou=users,dc=example,dc=com:
objectClass: inetOrgPerson
memberOf: cn=group2,ou=groups,dc=example,dc=com
...
```

kevinpapst:
Thanks for sharing, I could import it after some modifications. 

For clarificaton: I can't use your suggested approach above, because I would have to write all the necessary code myself. And I'd like to stick to a tested bundle as far as possible. So we have to find a solution that works with the bundle logic: 
the bundle uses a bind to authenticate the user. But if a new user authenticates the first time, it is performing a search to fetch all fields that should be imported into Kimai. These fields can now be mapped to the user and this mapping is under my control.

I don't see the `memberof` field with that query:
```
ldapsearch -LLL -x -b "ou=people,dc=test,dc=local"
```
But only when adding these (operational search attributes?):
```
ldapsearch -LLL -x -b "ou=people,dc=test,dc=local" "+" "*"
```

I could patch the used search query to use those attributes, but for now hard-coded, with such an result:
Bildschirmfoto 2019-05-29 um 14 44 15

And now my follow-up questions...

Are these search attributes OpenLDAP specific? 
Do they need to be configurable (e.g. when using an AD, would that fail)?
How do we map the memberof to Kimai groups. In the example from the screenshot, I could imagine a simple config like this:
```
"cn=group1,ou=groups,dc=kimai,dc=org" => ROLE_TEAMLEAD 
"cn=group2,ou=groups,dc=kimai,dc=org" => ROLE_ADMIN
```
Could that work or am I completely on the wrong track here?

army1349:
> the bundle uses a bind to authenticate the user. But if a new user authenticates the first time, it is performing a search to fetch all fields that should be imported into Kimai.

If this is how it works, then I would leave LDAP groups idea for now.
Group membership can change over time, so onetime LDAP search for membership is not an option.
Groups would have to be checked periodically using dedicated LDAP user.
For example my Nextcloud is checking groups using user:
`uid=nextcloud,ou=system,ou=users,dc=example,dc=com`
This would require more work and need for this is questionable, so I propose using internal groups with LDAP users for now a and wait for feedback. What do you think?

kevinpapst:
Thanks for all your input! 

I finally found a way to sync all fields on each login (abusing the framework a bit, but who cares, it works), updated the branch and added extensive documentation: https://www.kimai.org/documentation/ldap.html

I wouldn't mind some feedback on that.

army1349:
Don't mention it.
Is it OK for Kimai, that group membership will update from LDAP only when user logs in?
Do Kimai roles have fixed names like ROLE_TEAMLEAD, so there has to be a group-role mapping?

tobybatch:
@kevinpapst I'm back in the game.  I don't know how far you have got but this drupal 8 module https://www.drupal.org/project/ldap uses LDAP to sync users and is written on top of the symfony framework (although not fosuser).  The source code is here: https://git.drupalcode.org/project/ldap

I'll take a better look today.

tobybatch:
I agree with @army1349, pretty much totally

I think Peter's most salient point is LDAP admins tend to know what they are doing, so Kimai could mandate a group name for each of it's roles and we could add a roadmap entry to adding mapping latter if the LDAP uptake seems popular.

We don't use the memberOf to include users into groups but instead use memberuid in the group to indicate who is in a given group, e.g. 

```
dn: cn=webdev,ou=groups,dc=foo,dc=bar
cn: webdev
gidnumber: 12345
memberuid: rupert
memberuid: harry
memberuid: rob
...
```

I've just seen @army1349's comment about fixed names and the answer is yes. ROLE_TEAMLEAD is fixed, https://github.com/kevinpapst/kimai2/blob/master/src/Entity/User.php#L33

If it works with your bundle then expecting a ROLE_TEAMLEAD to be in an LDAP group named ```cn: role_teamlead``` or ```kimai_teamlead``` would work for us.

kevinpapst:
> Is it OK for Kimai, that group membership will update from LDAP only when user logs in?

Not sure that I understand the question. Kimai uses roles exclusively to check permissions while the user is performing tasks. I could add a command that you can use in a cron job, but not in this PR. If it  would be a requirement for you to sync roles also in between (e.g. when you configure very long session times or use the "remember me" feature), open up a new feature request later on.

@tobybatch thanks, I'll have a look at the Drupal stuff, but I think I am too far with the other bundle. What I documented (see link in the initial posting) should already work.

> I think Peter's most salient point is LDAP admins tend to know what they are doing, 

Hahaha, I hope so! The thing is: I don't in this PR (at least for the LDAP part), thats why I need more input.

I don't need exact Kimai role names to be returned, I will leave the mapping as is. So you configure a mapping. This can be `kimai_teamlead => ROLE_TEAMLEAD` or an arbitrary string like `cn=webdev,ou=groups,dc=foo,dc=bar => ROLE_TEAMLEAD`.

As I already added my own query to sync attributes during each login, I can add another search to find the user groups. But I'd need detailed information how that works, including:
- LDIFs with test data
- query that will match the data
- explanation how to filter the groups

I guess for the start you need to be able to define one query for the groups?!?
But how is the resultset structured? In your example @tobybatch is seems as if I get the usernames and have to iterate over them to check if one entry matches the user that authenticates. Right? Or will the username be included in the query and I get all groups of that user?

- Will I get the exact groups of the user and map them to Kimai roles?
- Will I get a an array of all users and iterate over the resultset to check if the user is included...

Please provide as many input as possible! This could be also a working LDIF example and a working config from another software. I should be able to extract the logic from there as well (but words make it even clearer).

kevinpapst:
Ok, thanks for being with me until here :-D

I went back to the whiteboard and read a lot of docs:
- then I removed the previously added "broken group sync via attributes"
- afterwards I added configs and support for another `ldap->search()` for the group sync as proposed by both of you @army1349 and @tobybatch 
- and finally I updated the docs: https://www.kimai.org/documentation/ldap.html

This PR is ready for real world testing, there are no more open points from my side.

Is one of you capable of testing it?
Or reading the docs and letting me know if that is what you would expect?

tobybatch:
I will test this today.

kevinpapst:
Any feedback is highly appreciated!
This PR is huge and code is not like wine: it doesn't get better without usage ;-)

tobybatch:
@kevinpapst I'm having trouble authenticating when I bind as a user.  This is to do with our ldap server setup not your code directly.  @army1349 Do you know how I allow user bind searches?

This returns the users correctly:

    ldapsearch -D cn=admin,dc=neontribe,dc=net -w XXXX -p 389 -h 192.168.21.97 -b "ou=users,dc=neontribe,dc=net" -s sub "(&(uid=tobias))"

This returns no such object:

    ldapsearch -D "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -w XXXXX -p 389 -h 192.168.21.97 -b "ou=users,dc=neontribe,dc=net" -s sub "(&(uid=tobias))"

In other apps we use, owncloud, openvpn, drupal etc we have an option bind as admin rather than as user to do the searches.

army1349:
@tobybatch your setup seems to be legit. Basic user often doesn't have permissions to crawl through LDAP structure. I have similar setup myself.
Login bind should work with base DN: dc=neontribe,dc=net or user object: cn=Toby Batch,ou=users,dc=neontribe,dc=net.

tobybatch:
thanks @army1349 

@kevinpapst  The bind to check the user does work, but the user object does not get retrieved.  Who are we binding as when we search for the user object?

```[2019-06-03 13:19:51] app.DEBUG: ldap_search(dc=neontribe,dc=net, (&(uid=tobias)), [array]) {"action":"ldap_search","base_dn":"dc=neontribe,dc=net","filter":"(&(uid=tobias))","attributes":["+","*"]} [] 
[2019-06-03 13:19:51] app.INFO: User tobias found on LDAP {"action":"loadUserByUsername","username":"tobias","result":"found"} [] 
[2019-06-03 13:19:51] php.INFO: User Deprecated: Calling "Symfony\Component\Security\Core\User\UserChecker::checkPreAuth()" with an AdvancedUserInterface is deprecated since Symfony 4.1. Create a custom user checker if you wish to keep this functionality. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Calling \"Symfony\\Component\\Security\\Core\\User\\UserChecker::checkPreAuth()\" with an AdvancedUserInterface is deprecated since Symfony 4.1. Create a custom user checker if you wish to keep this functionality. at /home/tobias/workspace/kimai/pr-815/vendor/symfony/security/Core/User/UserChecker.php:36)"} [] 
[2019-06-03 13:19:51] app.DEBUG: ldap_bind(tobias) {"action":"ldap_bind","bindDn":"tobias"} [] 
[2019-06-03 13:19:51] app.DEBUG: ldap_search(dc=neontribe,dc=net, (&(uid=tobias)), [array]) {"action":"ldap_search","base_dn":"dc=neontribe,dc=net","filter":"(&(uid=tobias))","attributes":["+","*"]} [] 
[2019-06-03 13:19:51] app.DEBUG: Zend\Ldap\Exception\LdapException: 0x20 (No such object):
... 
```

kevinpapst:
Depends on your config. 
Did you set 
```
        connection:
            host: 127.0.0.1
            #port: 389
            #useSsl: false                      # Enable SSL negotiation
            #useStartTls: false                 # Enable TLS negotiation
==>     username: foo
==>     password: bar
```

tobybatch:
Yes.  I just wanted check it was using the administrator login to search for the user objects.  I'll keep looking into it.

kevinpapst:
This docu explains all the connection settings: https://docs.zendframework.com/zend-ldap/api/
Maythe these attributes cause it `"attributes":["+","*"]`? I read that they are required for stuff like the `memberof` overlay. But know that I kind of removed support for it, they maybe could be removed?

tobybatch:
OK, so it seems to be re-binding as the user when it tries to look up the user object.  There is the log from my LDAP when I first try to log in.

```
ldap_1                   | 5cf52f51 conn=89110 fd=20 ACCEPT from IP=192.168.21.148:49552 (IP=0.0.0.0:389)
ldap_1                   | 5cf52f51 conn=89110 op=0 BIND dn="cn=admin,dc=neontribe,dc=net" method=128
ldap_1                   | 5cf52f51 conn=89110 op=0 BIND dn="cn=admin,dc=neontribe,dc=net" mech=SIMPLE ssf=0
ldap_1                   | 5cf52f51 conn=89110 op=0 RESULT tag=97 err=0 text=
ldap_1                   | 5cf52f51 conn=89110 op=1 SRCH base="ou=users,dc=neontribe,dc=net" scope=2 deref=0 filter="(&(uid=tobias))"
ldap_1                   | 5cf52f51 conn=89110 op=1 SRCH attr=+ *
ldap_1                   | 5cf52f51 conn=89110 op=1 SEARCH RESULT tag=101 err=0 nentries=1 text=
ldap_1                   | 5cf52f51 conn=89110 op=2 SRCH base="dc=neontribe,dc=net" scope=2 deref=0 filter="(&(uid=tobias))"
ldap_1                   | 5cf52f51 conn=89110 op=2 SRCH attr=dn
ldap_1                   | 5cf52f51 conn=89110 op=2 SEARCH RESULT tag=101 err=0 nentries=1 text=
ldap_1                   | 5cf52f51 conn=89110 op=3 BIND anonymous mech=implicit ssf=0
ldap_1                   | 5cf52f51 conn=89110 op=3 BIND dn="cn=Toby Batch,ou=users,dc=neontribe,dc=net" method=128
ldap_1                   | 5cf52f51 conn=89110 op=3 BIND dn="cn=Toby Batch,ou=users,dc=neontribe,dc=net" mech=SIMPLE ssf=0
ldap_1                   | 5cf52f51 conn=89110 op=3 RESULT tag=97 err=0 text=
ldap_1                   | 5cf52f51 conn=89110 op=4 SRCH base="ou=users,dc=neontribe,dc=net" scope=2 deref=0 filter="(&(uid=tobias))"
ldap_1                   | 5cf52f51 conn=89110 op=4 SRCH attr=+ *
ldap_1                   | 5cf52f51 conn=89110 op=4 SEARCH RESULT tag=101 err=32 nentries=0 text=
ldap_1                   | 5cf52f51 conn=89110 op=5 UNBIND
ldap_1                   | 5cf52f51 conn=89110 fd=20 closed
```

When I try do the search from the cli I can't get the user object:
```
tobias@tobias:kimai2-app $ ldapsearch -D "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -w XXXXX -p 389 -h 192.168.21.97 -b "ou=users,dc=neontribe,dc=net" -s sub
# extended LDIF
#
# LDAPv3
# base <ou=users,dc=neontribe,dc=net> with scope subtree
# filter: (objectclass=*)
# requesting: ALL
#

# search result
search: 2
result: 32 No such object

# numResponses: 1
```

The re-bound user can only find the user object if we point directly to it in the search base:

```
tobias@tobias:kimai2-app $ ldapsearch -D "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -w XXXXX -p 389 -h 192.168.21.97 -b "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -s sub
# extended LDIF
#
# LDAPv3
# base <cn=Toby Batch,ou=users,dc=neontribe,dc=net> with scope subtree
# filter: (objectclass=*)
# requesting: ALL
#

# Toby Batch, users, neontribe.net
dn: cn=Toby Batch,ou=users,dc=neontribe,dc=net
cn: Toby Batch
gidNumber: 10000
givenName: Toby
homeDirectory: /home/users/tobias
initials: tobias
loginShell: /bin/bash
mail: tobias@neontribe.co.uk
mobile: 07968505897
objectClass: inetOrgPerson
objectClass: posixAccount
objectClass: top
sn: Batch
uid: tobias
uidNumber: 10001
userPassword:: e2NyeXB0fSQxJGwuMEVWTDFyJGZLaXFmd2ZoYnN1VjFMT0tYQTRNLjA=

# search result
search: 2
result: 0 Success

# numResponses: 2
# numEntries: 1
tobias@tobias:kimai2-app $ 
```

tobybatch:
Here is the log output from an owncloud successful login, it also seems to rebind as the user but does pass the full path to the directory object used to extract the user object:
```
ldap_1                   | 5cf5340b conn=89251 fd=20 ACCEPT from IP=172.21.0.1:35368 (IP=0.0.0.0:389)
ldap_1                   | 5cf5340b conn=89251 op=0 BIND dn="cn=admin,dc=neontribe,dc=net" method=128
ldap_1                   | 5cf5340b conn=89251 op=0 BIND dn="cn=admin,dc=neontribe,dc=net" mech=SIMPLE ssf=0
ldap_1                   | 5cf5340b conn=89251 op=0 RESULT tag=97 err=0 text=
ldap_1                   | 5cf5340b conn=89251 op=1 SRCH base="dc=neontribe,dc=net" scope=2 deref=0 filter="(&(|(objectClass=inetOrgPerson))(|(uid=tobias)(|(?mailPrimaryAddress=tobias)(mail=tobias))))"
ldap_1                   | 5cf5340b conn=89251 op=1 SRCH attr=dn uid samaccountname mail cn entryuuid jpegphoto thumbnailphoto
ldap_1                   | 5cf5340b conn=89251 op=1 SEARCH RESULT tag=101 err=0 nentries=1 text=
ldap_1                   | 5cf5340b conn=89252 fd=21 ACCEPT from IP=172.21.0.1:35372 (IP=0.0.0.0:389)
ldap_1                   | 5cf5340b conn=89252 op=0 BIND dn="cn=toby batch,ou=users,dc=neontribe,dc=net" method=128
ldap_1                   | 5cf5340b conn=89252 op=0 BIND dn="cn=Toby Batch,ou=users,dc=neontribe,dc=net" mech=SIMPLE ssf=0
ldap_1                   | 5cf5340b conn=89252 op=0 RESULT tag=97 err=0 text=
ldap_1                   | 5cf5340b conn=89252 op=1 UNBIND
ldap_1                   | 5cf5340b conn=89252 fd=21 closed
ldap_1                   | 5cf5340b conn=89251 op=2 SRCH base="cn=toby batch,ou=users,dc=neontribe,dc=net" scope=0 deref=0 filter="(objectClass=*)"
ldap_1                   | 5cf5340b conn=89251 op=2 SRCH attr=primarygroupid
ldap_1                   | 5cf5340b conn=89251 op=2 SEARCH RESULT tag=101 err=0 nentries=1 text=
ldap_1                   | 5cf5340b conn=89251 op=3 SRCH base="dc=neontribe,dc=net" scope=2 deref=0 filter="(&(&(|(objectClass=posixGroup))(|(cn=gnomes)(cn=staff)))(uniqueMember=cn=toby batch,ou=users,dc=neontribe,dc=net))"
ldap_1                   | 5cf5340b conn=89251 op=3 SRCH attr=cn dn
ldap_1                   | 5cf5340b <= mdb_equality_candidates: (cn) not indexed
ldap_1                   | 5cf5340b <= mdb_equality_candidates: (cn) not indexed
ldap_1                   | 5cf5340b <= mdb_equality_candidates: (uniqueMember) not indexed
ldap_1                   | 5cf5340b conn=89251 op=3 SEARCH RESULT tag=101 err=0 nentries=0 text=
ldap_1                   | 5cf5340b conn=89251 op=4 UNBIND
```

kevinpapst:
Can you post your Kimai `ldap` config?

tobybatch:
local.yaml
```
kimai:
    ldap:
        active: true  # default: false
        
        # more infos about the connection params can be found at:
        # https://docs.zendframework.com/zend-ldap/api/
        connection:
            host: 192.168.21.97
            #port: 389
            #useSsl: false                      # Enable SSL negotiation
            #useStartTls: false                 # Enable TLS negotiation
            username: cn=admin,dc=neontribe,dc=net
            password: XXXXXXXXXXXX
            bindRequiresDn: true
            baseDn: dc=neontribe,dc=net
            #accountFilterFormat: (&(uid=%s))   # %s will be replaced with the username
            #allowEmptyPassword: false
            optReferrals: true
            #tryUsernameSplit: 
            #networkTimeout: 
            #accountCanonicalForm: 3
            #accountDomainName: HOST
            #accountDomainNameShort: HOST

        user:
            baseDn: ou=users,dc=neontribe,dc=net
            #filter: (&(ObjectClass=Person))      # extended search filter for users
            #usernameAttribute: uid               # field used in the LDAP bind for the given "login username" 

            # Mapping LDAP attributes to user entity
            attributes:
                - { ldap_attr: uid, user_method: setUsername }
            #    - { ldap_attr: mail, user_method: setEmail }
            #    - { ldap_attr: cn, user_method: setAlias }
```

I've used drupal settings as there is a config export function:

```
tobias@tobias:drupal $ drush cget ldap_servers.server.neontribe
uuid: 3252ed7e-641d-48bb-86e0-8866e4901eaf
langcode: en-gb
status: true
dependencies: {  }
id: neontribe
label: Neontribe
type: openldap
address: 192.168.21.97
port: 389
timeout: 10
tls: false
followrefs: null
weight: null
bind_method: service_account
binddn: 'cn=admin,dc=neontribe,dc=net'
bindpw: XXXXXXXXXX
basedn: 'ou=users,dc=neontribe,dc=net'
user_attr: uid
account_name_attr: cn
mail_attr: mail
mail_template: ''
picture_attr: ''
unique_persistent_attr: uid
unique_persistent_attr_binary: false
user_dn_expression: ''
testing_drupal_username: tobias
testing_drupal_user_dn: 'cn=Toby Batch,ou=users,dc=neontribe,dc=net'
grp_unused: true
grp_object_cat: ''
grp_nested: false
grp_user_memb_attr_exists: false
grp_user_memb_attr: ''
grp_memb_attr: ''
grp_memb_attr_match_user_attr: ''
grp_derive_from_dn: '0'
grp_derive_from_dn_attr: ''
grp_test_grp_dn: ''
grp_test_grp_dn_writeable: ''
search_pagination: false
search_page_size: null
```

kevinpapst:
I thnik I know what the problem is, but I'd like to rephrase it as question. 
The current flow:
- service account bind
- perform a "simple search" for the user by 
- if the account is found and a DN is returned
- a bind is executed to authenticate the user
- **a new search is executed to fetch the user attributes by username**
- another search is executed to fetch the user groups by DN

So I guess, I have to adjust the bold **query for user attributes** to use the DN as well, right?

A full log on my side looks like this:
```
5cf55177 conn=1001 fd=11 ACCEPT from IP=127.0.0.1:61148 (IP=0.0.0.0:389)
5cf55177 conn=1001 op=0 BIND dn="" method=128
5cf55177 conn=1001 op=0 RESULT tag=97 err=0 text=
5cf55177 conn=1001 op=1 SRCH base="dc=kimai,dc=org" scope=2 deref=0 filter="(&(uid=user2))"
5cf55177 conn=1001 op=1 SRCH attr=dn
5cf55177 conn=1001 op=1 SEARCH RESULT tag=101 err=0 nentries=1 text=
5cf55177 conn=1001 op=2 BIND dn="uid=user2,ou=users,dc=kimai,dc=org" method=128
5cf55177 conn=1001 op=2 BIND dn="uid=user2,ou=users,dc=kimai,dc=org" mech=SIMPLE ssf=0
5cf55177 conn=1001 op=2 RESULT tag=97 err=0 text=
5cf55177 conn=1001 op=3 SRCH base="ou=users,dc=kimai,dc=org" scope=2 deref=0 filter="(&(uid=user2))"
5cf55177 conn=1001 op=3 SRCH attr=+ *
5cf55177 conn=1001 op=3 SEARCH RESULT tag=101 err=0 nentries=1 text=
5cf55177 conn=1001 op=4 SRCH base="ou=groups,dc=kimai,dc=org" scope=2 deref=0 filter="(&(&(objectClass=groupOfNames)(|(cn=group1)(cn=group2)(cn=kimai_admin)))(member=uid=user2,ou=users,dc=kimai,dc=org))"
5cf55177 conn=1001 op=4 SRCH attr=cn + *
5cf55177 conn=1001 op=4 SEARCH RESULT tag=101 err=0 nentries=3 text=
5cf55177 conn=1001 op=5 UNBIND
5cf55177 conn=1001 fd=11 closed
```
You can see the group query already utilizes the users DN. I haven't noticed the problem on the second query, as it always worked on my local LDAP.

Thanks, thats exactly the feedback I need 👍 I will push a change later on, if you verify my assumptions

kevinpapst:
Riiiight, I could use an LDAP book, that would make it much easier^^
After having a fifth look at the owncloud example, I saw that I always compared the groups query, instead of the user attributes query. But you were talking about this one, didn't you:
```
ldap_1                   | 5cf5340b conn=89251 op=2 SRCH base="cn=toby batch,ou=users,dc=neontribe,dc=net" scope=0 deref=0 filter="(objectClass=*)"
```
So the search base is not the baseDn, but the users DN?
Is that a generic aproach that will work on every LDAP setup?
And the filter `(objectClass=*)` can also be used in every case?

Just asking because I need to know if I need new configs to define this query or if I can safely assume that it is always ok to use the userDn as search base and the filter `(objectClass=*)` as static setting.

tobybatch:
Yes that's the query I mean.  For me when I'm bound as a user rather than admin I need to target the full object, cn=name, dc=foo etc.

The object class is just a top level (parent) object.  Each gets more specific as drill down.  e.g. objecClass=* will match inetPerson or posixUser or anything.

In my experience that filter is fine left as is if you are targeting a known object such as the user object.

I'm here to help whenever I can.  Honestly getting LDAP into kimai will be epic.

kevinpapst:
Ok, you can test it once more. 

But I have one more question... I may have optimized it a bit too much:
Is it possible that a users DN changes? It is now cached and will be re-used on following logins. 

So the initial query e.g. `SRCH base="ou=users,dc=kimai,dc=org" scope=2 deref=0 filter="(&(uid=user3))"` is not executed again. 
Following logins will always query the DN directly: `SRCH base="uid=user3,ou=users,dc=kimai,dc=org" scope=2 deref=0 filter="(objectClass=*)"`

So: shall I add a method `updateDn()` using the initial filter `(&(uid=user3))` which will be called on every login before fetching user attributes?

tobybatch:
w00t!  I have a login!  I'll poke into the corners today.

tobybatch:
Roles don't work for us at the moment, but that's ok.  Can I see your LDIF export for a sample subset of groups and an user?
```
dn: cn=adm,ou=groups,dc=neontribe,dc=net
cn: adm
gidnumber: 4
memberuid: tobbias
memberuid: ...
memberuid: ...
memberuid: ...
...

dn: cn=Toby Batch,ou=users,dc=neontribe,dc=net
cn: Toby Batch
gidnumber: 1000
givenname: Toby
homedirectory: /home/users/tobias
loginshell: /bin/bash
mail: toby@neontribe.co.uk
objectclass: inetOrgPerson
objectclass: posixAccount
objectclass: top
sn: Batch
uid: tobias
uidnumber: 10000
userpassword: XXXXXXXXXXXXXXXX
```

And I have the same permissions issue with searching as a user:

```
tobias@tobias:pr-815 (ldap) $ ldapsearch -D "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -W -p 389 -h 192.168.21.97 -b "cn=adm,ou=groups,dc=neontribe,dc=net" -s sub "(&(memberuid=tobias))"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <cn=adm,ou=groups,dc=neontribe,dc=net> with scope subtree
# filter: (&(memberuid=tobias))
# requesting: ALL
#

# search result
search: 2
result: 32 No such object

# numResponses: 1

```
But if I search as admin:
```
tobias@tobias:pr-815 (ldap) $ ldapsearch -D cn=admin,dc=neontribe,dc=net -W -p 389 -h 192.168.21.97 -b "cn=adm,ou=groups,dc=neontribe,dc=net" -s sub "(&(memberuid=tobias))"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <cn=adm,ou=groups,dc=neontribe,dc=net> with scope subtree
# filter: (&(memberuid=tobias))
# requesting: ALL
#

# adm, groups, neontribe.net
dn: cn=adm,ou=groups,dc=neontribe,dc=net
cn: adm
gidNumber: 4
memberUid: rupert
memberUid: katja
memberUid: rob
memberUid: neil
memberUid: oliver
memberUid: falconx
memberUid: andy
memberUid: mattlayt
memberUid: tobias
memberUid: steph
memberUid: anthony
memberUid: frank
memberUid: charles
objectClass: posixGroup
objectClass: top

# search result
search: 2
result: 0 Success

# numResponses: 2
# numEntries: 1
```

kevinpapst:
I used the (adjusted) example provided by @army1349  (password `kitten`)
```
dn: cn=group1,ou=groups,dc=kimai,dc=org
objectClass: groupOfNames
member: uid=user1,ou=users,dc=kimai,dc=org
member: uid=user2,ou=users,dc=kimai,dc=org

dn: cn=group2,ou=groups,dc=kimai,dc=org
objectClass: groupOfNames
member: uid=user2,ou=users,dc=kimai,dc=org
member: uid=user3,ou=users,dc=kimai,dc=org

dn: uid=user1,ou=users,dc=kimai,dc=org
objectClass: inetOrgPerson
cn: Blub 1
sn: blub1
userpassword: {MD5}baic0Jq3k3R4odR9IJOFNg==

dn: uid=user2,ou=users,dc=kimai,dc=org
objectClass: inetOrgPerson
cn: Blub 2
sn: blub2
userpassword: {MD5}baic0Jq3k3R4odR9IJOFNg==

dn: uid=user3,ou=users,dc=kimai,dc=org
objectClass: inetOrgPerson
cn: Blub 3
sn: blub3
userpassword: {MD5}baic0Jq3k3R4odR9IJOFNg==
```
The group query looks like this:
```
5cf62dd1 conn=1087 op=5 SRCH base="ou=groups,dc=kimai,dc=org" scope=2 deref=0 filter="(&(member=uid=user3,ou=users,dc=kimai,dc=org))"
```
with the config
```
        role:
            baseDn: ou=groups, dc=kimai, dc=org
#            filter: (&(objectClass=groupOfNames)(|(cn=group1)(cn=group2)(cn=kimai_admin)))
            userDnAttribute: member
```
So we need a new config to rebind as "service user" before fetching the groups?
Or should that always be done if a service account was configured under `kimai.ldap.connection`?

Can you briefly explain me how I can reconfigure my OpenLDAP to be more strict with the privileges, so that I see the same problems that you experience? Why can my user access the attributes and groups?

kevinpapst:
> Is it possible that a users DN changes?
> Question: shall I add a method updateDn() and use the initial filter (&(uid=user3)) to find the users DN on every login before fetching user attributes with the full DN?
Do you have an opinion on this one @army1349 ?

army1349:
> 
> 
> > Is it possible that a users DN changes?
> > Question: shall I add a method updateDn() and use the initial filter (&(uid=user3)) to find the users DN on every login before fetching user attributes with the full DN?
> 
> Do you have an opinion on this one @army1349 ?

It is not common, that user DNs change, but sometimes it happens. I did it once and in most apps I had to manually change cached values in SQL.

Nextcloud has admin interface for: Clear Username-LDAP User Mapping and Clear Groupname-LDAP Group Mapping, but I kind of like the idea, that user is authenticated via LDAP according to actual LDAP state and actual configuration and nothing more, no hidden caches.

kevinpapst:
Alright, I feared that ;-)
I have to use the cache due to the Symfony internal authentication flow. It is more by accident that it is now used to bind directly instead of fetching the users DN first. I'll see if I can fix that easily.

tobybatch:
I didn't deliberatley secure the LDAP.  It's just following this HOWTO https://help.ubuntu.com/lts/serverguide/openldap-server.html.en

tobybatch:
If we need to mandate this way of doing groups then I could add support for that to our schema, I really don't how unique out LDAP is.

kevinpapst:
@army1349 what is your experience with other apps here? Under which bind scope are group queries "normally" done?
I'd like to use the "standard" way if there is one, unfortunately I don't know which one it is...

army1349:
> 
> 
> @army1349 what is your experience with other apps here? Under which bind scope are group queries "normally" done?
> I'd like to use the "standard" way if there is one, unfortunately I don't know which one it is...

Not sure I understand question, but I will try.
Group base DN should be configurable and separate from user base DN.
Bind DN is usually the same for users and groups
Search scope use usually _sub_.
Group search filter should be configurable.

kevinpapst:
That usually happens when people with different knowledge levels talk about the stuff ;-)

My question was related to:
> Bind DN is usually the same for users and groups

Currently I query the user attributes after using `bind` as user as discussed above. I bind via the users full `DN` and then query something like (filter is configurable):
```
SRCH base="ou=groups,dc=kimai,dc=org" scope=2 deref=0 filter="(&(member=uid=user2,ou=users,dc=kimai,dc=org))"
```
Can the groups usually be retrieved by the user or shall I re-bind as admin (service account) before?
Or is it initially wrong that I query the user attributes as user instead of being bound as service account?

tobybatch:
The issue I have is that the user bind, cn=John Doe, dc=kimai, dc=org with the user password, kitten doesn't have the rights to search the sub tree.  I believe this is pretty standard.

> @tobybatch your setup seems to be legit. Basic user often doesn't have permissions to crawl through LDAP structure. I have similar setup myself.

So I think we need to bind as the admin user, cn=admin, dc=kimai, dc=org to look up the objects.  There is a separate question of groups enumerating a list of users or user objects listing groups for that user.

army1349:
> Can the groups usually be retrieved by the user or shall I re-bind as admin (service account) before?
Or is it initially wrong that I query the user attributes as user instead of being bound as service account?

I agree, groups should be definitely queried using service account.

Users probably too. Usually users have rights to see his own attributes, but necessarily all of them. For our needs it should suffice, but using service account is safe play.

>  There is a separate question of groups enumerating a list of users or user objects listing groups for that user.

Not all LDAP installations have reverse membership attributes, so I vote for querying groups for their members.

tobybatch:
@army1349 Our schemas have different group referrences, memberUid vs member.  I'm not adverse to changing ours if the member notation is better supported but is there a way that the filter in kimai could be set up to handle that without @kevinpapst  having to add lot of conditional code.

Ours: ```memberuid: tobias```

Yours: ```member: uid=user3,ou=users,dc=kimai,dc=org```

kevinpapst:
@tobybatch I added a config so you can change the group query. It is not yet documented, I will update the documentation once everything is in place.

Change your config like this:
```
        role:
            baseDn: ou=groups, dc=kimai, dc=org
            usernameAttribute: uid
            userDnAttribute: memberuid
            #groups:
             #   - { ldap_value: group1, role: ROLE_TEAMLEAD }
              #  - { ldap_value: group2, role: ROLE_ADMIN }
               # - { ldap_value: kimai_admin, role: ROLE_SUPER_ADMIN }
```
`userDnAttribute` will become the first part of the group filter (default `member`) and `usernameAttribute` is the field which is taken from the user search result (default `dn`) and injected into the filter as second argument.

The service-account bind is hopefully the last piece of the puzzle for a first release *fingers crossed*

tobybatch:
The query looks like it should work, but only if I'm bound as the admin user.  Kimai is still erroring on an unfound oject. 
```
tobias@tobias:pr-815 (ldap) $ ldapsearch -D "cn=Toby Batch,ou=users,dc=neontribe,dc=net" -W -p 389 -h 192.168.21.97 -b 'ou=groups,dc=neontribe,dc=net' -s sub "(&(memberuid=tobias))"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <ou=groups,dc=neontribe,dc=net> with scope subtree
# filter: (&(memberuid=tobias))
# requesting: ALL
#

# search result
search: 2
result: 32 No such object

# numResponses: 1
tobias@tobias:pr-815 (ldap) $ ldapsearch -D cn=admin,dc=neontribe,dc=net -W -p 389 -h 192.168.21.97 -b 'ou=groups,dc=neontribe,dc=net' -s sub "(&(memberuid=tobias))"
Enter LDAP Password: 
# extended LDIF
#
# LDAPv3
# base <ou=groups,dc=neontribe,dc=net> with scope subtree
# filter: (&(memberuid=tobias))
# requesting: ALL
#

# adm, groups, neontribe.net
dn: cn=adm,ou=groups,dc=neontribe,dc=net
cn: adm
gidNumber: 4
memberUid: rupert
memberUid: ...
memberUid: ...
memberUid: ...
memberUid: ...
memberUid: ...
memberUid: ...
memberUid: ...

# ops, groups, neontribe.net
dn: cn=ops,ou=groups,dc=neontribe,dc=net
cn: ops
gidNumber: 10003
memberUid: tobias
memberUid: ...
memberUid: ...
memberUid: ...
memberUid: ....
memberUid: ...
```

kevinpapst:
One step after another ;-) I am trying to figure out a way how to do the admin bind without having to bind 3 times in a row ... the symfony authentication system is a powerful nightmare...
In other words: I havent worked on that part yet ;-) will push my changes later

tobybatch:
@kevinpapst Sorry, yes.  The query looks good for the alternative schema structure.

kevinpapst:
@tobybatch I pushed a quick solution. Please give it a try. 
I can still "polish" the code once everything is working as intended.

kevinpapst:
btw, I think this is the config you need:
```
        role:
            baseDn: ou=groups,dc=neontribe,dc=net
            usernameAttribute: uid
            userDnAttribute: memberuid
            groups:
                - { ldap_value: ops, role: ROLE_SUPER_ADMIN }
```

tobybatch:
You are a wizard.  That works perfectly.  On re-login the tobias user gained the teal lead role.

```
kimai:
    ldap:
        active: true  # default: false
        
        # more infos about the connection params can be found at:
        # https://docs.zendframework.com/zend-ldap/api/
        connection:
            host: 192.168.21.97
            #port: 389
            #useSsl: false                      # Enable SSL negotiation
            #useStartTls: false                 # Enable TLS negotiation
            username: cn=admin,dc=neontribe,dc=net
            password: XXXXXX
            bindRequiresDn: true
            baseDn: dc=neontribe,dc=net
            #accountFilterFormat: (&(uid=%s))   # %s will be replaced with the username
            #allowEmptyPassword: false
            optReferrals: true
            #tryUsernameSplit: 
            #networkTimeout: 
            #accountCanonicalForm: 3
            #accountDomainName: HOST
            #accountDomainNameShort: HOST

        user:
            baseDn: ou=users,dc=neontribe,dc=net
            #filter: (&(ObjectClass=Person))      # extended search filter for users
            #usernameAttribute: uid               # field used in the LDAP bind for the given "login username" 

            # Mapping LDAP attributes to user entity
            attributes:
                - { ldap_attr: uid, user_method: setUsername }
                - { ldap_attr: mail, user_method: setEmail }
                - { ldap_attr: cn, user_method: setAlias }

        role:
            baseDn: ou=groups,dc=neontribe,dc=net
            usernameAttribute: uid
            userDnAttribute: memberuid
            # filter: (&(objectClass=memberUid))  # additional filter to query the groups
            # nameAttribute: cn                      # the group name
            # userDnAttribute: uid                # field to look in for the users dn 
            
            # Convert LDAP group name (nameAttribute) to Kimai role
            groups:
               - { ldap_value: kimai-admin, role: ROLE_TEAMLEAD }
            #    - { ldap_value: kimai_admin, role: ROLE_ADMIN }
```

army1349:
> @army1349 Our schemas have different group referrences, memberUid vs member. I'm not adverse to changing ours if the member notation is better supported but is there a way that the filter in kimai could be set up to handle that without @kevinpapst having to add lot of conditional code.
> 
> Ours: `memberuid: tobias`
> 
> Yours: `member: uid=user3,ou=users,dc=kimai,dc=org`

Hm, AFAIK posix group style memberUid is not widely recommended, but it is supported in Nextcloud for example. Member attribute interpretation has to be handled in a code, so either we expect its value to follow common schema rules:
member, uniqueMember - full DN
memberUid - uid
or we introduce some kind of switch in configuration.

kevinpapst:
Hooray 🎉 ! Then: happy edge case testing_

- delete (deactivate) a user from LDAP
- deactivate a user in Kimai
- demote a user from a group in LDAP
- promote a user to a group in LDAP
- change attributes in LDAP
- change users DN
- change passwords (remember to deactivate that in Kimai)

The `dn` cache might be the last open issue, I'll see if I can fix in in a couple of minutes or will write a follow up ticket for it.

kevinpapst:
> Hm, AFAIK posix group style memberUid is not widely recommended, but it is supported in Nextcloud for example. Member attribute interpretation has to be handled in a code, so either we expect its value to follow common schema rules:
> member, uniqueMember - full DN
> memberUid - uid
> or we introduce some kind of switch in configuration.

I think I already solved that. You can freely configure the left side of the filter and for the right side you can configure the user attribute that should be used. 
Default for the filter is: `member: userDn`, but you could also use `memberUid: uid` as in the config example above.

I hope that covers most of the cases, as I am against any more config switches unless someone actively asks for it.

army1349:
> > Hm, AFAIK posix group style memberUid is not widely recommended, but it is supported in Nextcloud for example. Member attribute interpretation has to be handled in a code, so either we expect its value to follow common schema rules:
> > member, uniqueMember - full DN
> > memberUid - uid
> > or we introduce some kind of switch in configuration.
> 
> I think I already solved that. You can freely configure the left side of the filter and for the right side you can configure the user attribute that should be used.
> Default for the filter is: `member: userDn`, but you could also use `memberUid: uid` as in the config example above.
> 
> I hope that covers most of the cases, as I am against any more config switches unless someone actively asks for it.

Great work! That should do it.

kevinpapst:
Hopefully last feature added: cached user DNs are updated before bind.

@army1349 are you going to test it as well on an existing setup? Just asking, so I know how long I should wait before merging.

army1349:
> 
> 
> Hopefully last feature added: cached user DNs are updated before bind.
> 
> @army1349 are you going to test it as well on an existing setup? Just asking, so I know how long I should wait before merging.

Sure, on it.

army1349:
I have created fresh installation, but LDAP auth is not working so far.
I didn't find any useful logs. Can you enlighten me about Kimai logging options?

kevinpapst:
If you run in `dev` environment (check `.env`) then the LDAP calls should be logged to `var/log/dev.log`.

army1349:
```
app.INFO: User test found on LDAP {"action":"loadUserByUsername","username":"test","result":"found"} []

php.INFO: User Deprecated: The "Symfony\Component\Security\Core\Encoder\BCryptPasswordEncoder" class is deprecated since Symfony 4.3
, use "Symfony\Component\Security\Core\Encoder\NativePasswordEncoder" instead. {"exception":"[object] (ErrorException(code: 0): User Deprecated: The \"Symfony\\Component\\Security\\Core\\Encoder\\BCryptPasswordEncoder\" class is deprecated since Symfony 4.3, use \"Symfony\\Component\\Security\\Core\\Encoder\\NativePasswordEncoder\" instead. at /usr/local/www/kimai2/vendor/symfony/security-core/Encoder/BCryptPasswordEncoder.php:14)"} []

security.INFO: Authentication request failed. {"exception":"[object] (Symfony\\Component\\Security\\Core\\Exception\\BadCredentialsException(code: 0): Bad credentials. at /usr/local/www/kimai2/vendor/symfony/security-core/Authentication/Provider/UserAuthenticationProvider.php:85, Symfony\\Component\\Security\\Core\\Exception\\BadCredentialsException(code: 0): The presented password is invalid. at/usr/local/www/kimai2/vendor/symfony/security-core/Authentication/Provider/DaoAuthenticationProvider.php:58)"} []

security.DEBUG: Authentication failure, redirect triggered. {"failure_path":"fos_user_security_login"} []
```
User account found in LDAP a then failed on DaoAuthenticationProvider.
Do I have to create user in Kimai first, or why is LdapBindAuthenticationProvider not used?

kevinpapst:
Show me your LDAP config please

army1349:
```
kimai:
    ldap:
        active: true

        connection:
            host: ldap.example.local
            port: 389
            useStartTls: true                 # Enable TLS negotiation
            username: uid=kimai,ou=system,ou=users,dc=example,dc=ch
            password: ******************
            bindRequiresDn: true
            baseDn: dc=example,dc=ch

        user:
            baseDn: ou=common,ou=users,dc=example,dc=ch
            filter: (&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch))      # extended search filter for users

            # Mapping LDAP attributes to user entity
            attributes:
                - { ldap_attr: uid, user_method: setUsername }
                - { ldap_attr: mail, user_method: setEmail }
                - { ldap_attr: cn, user_method: setAlias }
```

kevinpapst:
Config looks good, but how did you install Kimai?
You are using Symfony 4.3 and Kimai is restricted to use Symfony 4.2.x. There is something odd with your installation. 

The LdapUserProvider is triggered, but not the AuthenticationProvider.
Did you apply changes to config/security.yaml? This part is important:
```
    providers:
        chain_provider:
            chain:
                providers: [fos_userbundle, kimai_ldap]
        kimai_ldap:
            id: App\Ldap\LdapUserProvider
        fos_userbundle:
            id: fos_user.user_provider.username_email
...    
    firewalls:
        secured_area:
            # activate the LdapAuthenticationProvider
            kimai_ldap: ~
```

kevinpapst:
No, you don't need to create the user in Kimai first, that would defeat the purpose of LDAP somehow ;-) But if you have an existing "test" user in your database with a password, the FOSUserBundle will take over and the authentication will stop before reaching the Ldap part

kevinpapst:
Can you look for the `"name": "symfony/security"` part in composer.lock. It should look like this:
```
        {
            "name": "symfony/security",
            "version": "v4.2.8",
            "source": {
                "type": "git",
                "url": "https://github.com/symfony/security.git",
                "reference": "9895679e1b90f975c84916749ec01707d10b5ecf"
            },
            "dist": {
                "type": "zip",
                "url": "https://api.github.com/repos/symfony/security/zipball/9895679e1b90f975c84916749ec01707d10b5ecf",
                "reference": "9895679e1b90f975c84916749ec01707d10b5ecf",
                "shasum": ""
            },
```

army1349:
> Config looks good, but how did you install Kimai?
You are using Symfony 4.3 and Kimai is restricted to use Symfony 4.2.x. There is something odd with your installation.

From what I can see in composer.json, only symfony/security-bundle is restricted to 4.2.x.
Installation:
```
git clone ...
cd kimai2
git checkout ldap
composer install --no-dev --optimize-autoloader
```

> Can you look for the "name": "symfony/security" part in composer.lock. It should look like this:

There is no symfony/security in composer.lock, only:
```
symfony/security-bundle - v4.2.9
symfony/security-core - v4.3.0
symfony/security-csrf - v4.3.0
symfony/security-guard - v4.3.0
symfony/security-http - v4.3.0
```

> But if you have an existing "test" user in your database with a password, the FOSUserBundle will take over and the authentication will stop before reaching the Ldap part

There is no user "test" in my DB.

kevinpapst:
Symfony Flex should limit all packages to Symfony 4.2, see https://github.com/kevinpapst/kimai2/blob/master/composer.json#L132

army1349:
> 
> 
> Symfony Flex should limit all packages to Symfony 4.2, see https://github.com/kevinpapst/kimai2/blob/master/composer.json#L132

I finally managed to downgrade all Symfony packages to 4.2:
```
composer global require symfony/flex ^1.0
rm -rf vendor && composer up --no-dev
```
but nothing has changed.

army1349:
I think user filter is not correctly used when searching for DN:
```
Failed to retrieve DN for account: test [0x1: Unexpected result count (26) for: (&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch))] at /usr/local/www/kimai2/src/Ldap/LdapDriver.php:99)"} []
```

army1349:
One more thing. Is there a way to specify multiple LDAP hosts?
I can´t find a way to do it.

kevinpapst:
The user query must return one single result, your filter should look something like:
```
(&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch)(&(uid=%s)))
```
And multiple LDAP servers are not supported. 

Do you know hat fixed your last "not connecting"  problem? Maybe someone else runs into the same issue and I can document it right away.

army1349:
> The user query must return one single result, your filter should look something like:
> 
> ```
> (&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch)(&(uid=%s)))
> ```

There is accountFilterFormat and user filter and I don't understand how those two options work.
When I set accountFilterFormat like:
```
(&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch)(uid=%s))
```
and don't set user filter, everything is ok.
What is then user filter good for?
When I leave accountFilterFormat to its default and set user filter like:
```
(&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch))
```
then it fails. It seems like there are to searches and one will pass, because (uid=_USERNAME_) is added to search filter and the other one fails, because nothing is added to filter and therefore there are too many results.
> And multiple LDAP servers are not supported.

What a shame. This is limitation of Symfony framework? LDAP as single point of failure is horrible thing, therefore LDAP replication is a must in production systems. Even PHP ldap_connect function has support for it.

> Do you know what fixed your last "not connecting" problem? Maybe someone else runs into the same issue and I can document it right away.

It was all about that search filter. Only I didn't notice that search error because of many Symfony 4.3 deprecation logs.

kevinpapst:
Thats what testing is for, to find parts which are unclear or don't make a lot of sense.

Stick with the accountFilterFormat if everything works as expected. The other one is used internally, if you query the LDAP functions for other purposes (eg. search a user by another field). 
But I understand that it is confusing and I will remove it, as such feature does not yet exist.

I am pretty sure I could get it working with multiple servers, which would take me probably one more day. But I am one person building a time-tracker and I have to kick out a couple of features from initial versions. If I would try to build the best possible solution in the first step, I would not ship anything... And as a side note: if someone wants to use Kimai in a multi LDAP setup, it is very likely a bigger company. And that company can always ask for a contract to built such a feature or sent in a PR.

Thanks for the feedback so far, I will push a smaller change with the config update and sent a ping once the docs are updated.

army1349:
> Edit: tooooo late, I was wrong and I will post an update tomorrow

Sure, I will be waiting.

> I am pretty sure I could get it working with multiple servers, which would take me probably one more day. But I am one person building a time-tracker and I have to kick out a couple of features from initial versions. If I would try to build the best possible solution in the first step, I would not ship anything... And as a side note: if someone wants to use Kimai in a multi LDAP setup, it is very likely a bigger company. And that company can always ask for a contract to built such a feature or sent in a PR (or an admin in that company sets up a HA proxy or any other solution for failover).

I understand. I am just pointing out LDAP best practice staff, so you can make informed decision :-)

kevinpapst:
> I understand. I am just pointing out LDAP best practice staff, so you can make informed decision :-)

Very much appreciated! The outcome of the PR without yours and @tobybatch help would have been ... far from being ideal^^

I updated the docs and pushed some minor config changes, to hopefully simplify the configuration for the next users.

Now I could tell you, that you should ONLY set the `user.filter` and that it needs to include the user identifier filter eg. `(uid=%s)`, but then I would give away the chance to ask you for a look at [the new configuration docs](https://www.kimai.org/documentation/ldap.html#configuration) :-D

Maybe you find it out by yourself if you forget which fields were documented before?

army1349:
> Now I could tell you, that you should ONLY set the `user.filter` and that it needs to include the user identifier filter eg. `(uid=%s)`, but then I would give away the chance to ask you for a look at [the new configuration docs](https://www.kimai.org/documentation/ldap.html#configuration) :-D

I have pulled latest commit, updated composer, removed accountFilterFormat from config and now when I set:
`filter: (&(objectClass=inetOrgPerson)(uid=%s))`
it ends with:
`ParameterNotFoundException The service "App\Configuration\LdapConfiguration" has a dependency on a non-existent parameter "s))(uid=".`

Have I done anything wrong?

kevinpapst:
This error looks selfmade. Does you `git status` say anything about `config/services.yaml`?

army1349:
> 
> 
> This error looks selfmade. Does you `git status` say anything about `config/services.yaml`?

No, nothing. Same behavior even after clean install.
It is all about "%" character. If I replace **%s** with **test**, it is all working for test user.

I have read something about "%" escaping in symfony by doubling it, but that ends with:
`Uncaught PHP Exception ErrorException: "Warning: sprintf(): Too few arguments" at /usr/local/www/kimai2/vendor/zendframework/zend-ldap/src/Ldap.php line 513`

kevinpapst:
Hmpf, I don't understand what is going in your environment. 
You are changing your configs in config/packages/local.yaml?
@tobybatch could you test if the latest changes work for you?

tobybatch:
@kevinpapst will do.  @army1349 what OS, PHP version are you on?

tobybatch:
@kevinpapst I just pulled the PR, dropped/created my schema and it still works for me.

Ubuntu 18.04.2 LTS
PHP 7.2.17-0ubuntu0.18.04.1

Authing against openldap docker container: osixia/openldap:1.2.2

kevinpapst:
My bad @army1349 - awesome testing! You were totally right and I was testing with an old config, besides I didn't knew that "feature" of Symfony, parsing custom configs at that point in the code.
Learnt a lot of new things with the LDAP integration.
Please update and the error will be gone, I hope that finally solves all issues on your end.
@tobybatch if you update once more, you have to adjust your config/filter - I guess you currently don't have a user identifier like `(uid=%s)` in it? 
I didn't and therefor didn't see the error as well. Should have read my own documentation more carefully 😆

tobybatch:
Pulled and tested.  Still working.

You are right:  Filters are commented out.
```
        user:
            baseDn: ou=users,dc=neontribe,dc=net
            #filter: (&(ObjectClass=Person))      # extended search filter for users
            #usernameAttribute: uid               # field used in the LDAP bind for the given "login username" 

            # Mapping LDAP attributes to user entity
            attributes:
                - { ldap_attr: uid, user_method: setUsername }
                - { ldap_attr: mail, user_method: setEmail }
                - { ldap_attr: cn, user_method: setAlias }

        role:
            baseDn: ou=groups,dc=neontribe,dc=net
            usernameAttribute: uid
            userDnAttribute: memberuid
            # filter: (&(objectClass=memberUid))  # additional filter to query the groups
            # nameAttribute: cn                      # the group name
            # userDnAttribute: uid                # field to look in for the users dn 
            
            # Convert LDAP group name (nameAttribute) to Kimai role
            groups:
               - { ldap_value: kimai-admin, role: ROLE_TEAMLEAD }
            #    - { ldap_value: kimai_admin, role: ROLE_ADMIN }
```

army1349:
> @kevinpapst will do. @army1349 what OS, PHP version are you on?

I am on FreeBSD 11.2, PHP 7.2.19.

> My bad @army1349 - awesome testing! You were totally right and I was testing with an old config, besides I didn't knew that "feature" of Symfony, parsing custom configs at that point in the code.
> Learnt a lot of new things with the LDAP integration.
> Please update and the error will be gone, I hope that finally solves all issues on your end.

Great, "%" error is gone.

But there is still same issue with user search filter.
With `filter: (&(objectClass=inetOrgPerson)(uid=%s))`
I still get "Invalid credentials" and I see this in logs:
```
ldap_search(ou=common,ou=users,dc=example,dc=ch, (&(&(objectClass=inetOrgPerson)(uid=%s))(uid=test)), [array]) {"action":"ldap_search","base_dn":"ou=common,ou=users,dc=example,dc=ch","filter":"(&(&(objectClass=inetOrgPerson)(uid=%s))(uid=test))","attributes":["+","*"]} []
```
So it seems that there are two searches using user filter and one of them doesn't care about "%s" and just adds (uid=_username_) to the filter.

army1349:
I dug into code and have some good news for a change.
Multiple LDAP hosts are supported without code change.
If you set host like:
`host: "ldap://ldap.example.local ldap://ldap2.example.local"`
or
`host: "ldaps://ldap.example.local ldaps://ldap2.example.local"`
it will flow unchanged right into _ldap_connect_ and, as it was already said, _ldap_connect_ knows what to do with it.
I have already tested it and it works.

kevinpapst:
Thats good news! I will update the documentation accordingly. 
And thanks for staying with me @army1349 - I think I may have been working one day too much on this PR and lost focus while trying to fix that problem you reported. 

I can confirm that I re-introduced the bug with the last commit, while reverting the bugfix which was not properly working...
As first step I will write a unit test for this part of the code and then try to fix it. I'll update this PR later or tomorrow once more (this is likely the longest PR in Kimai history).

kevinpapst:
@army1349 I think you can try to set both configs and it should work for you as well:
- connection.accountFilterFormat: `(&(objectClass=inetOrgPerson)(uid=%s))`
- user.filter: `(&(objectClass=inetOrgPerson))`

I am working on improving the default configurations, so these problems go away.
I'd like to know if it is better to use:
- `posixAccount` with `posixGroup` or 
- `inetOrgPerson` with `groupOfNames`

as default settings? From what you wrote above I understand that the use of posixAccount is discouraged.

My proposal is to use the defaults:
- groups use `member`
- accountFilterFormat `(&(objectClass=inetOrgPerson)(uid=%s))`
- filter `(&(objectClass=inetOrgPerson))` (if a user is searched by name the `(uid=%s)` is automatically appended and that's what caused your problems)

I could also imagine for max compatibility:
- groups use `member`
- accountFilterFormat `(&(uid=%s))`
- filter ``

What would you say?

army1349:
> @army1349 I think you can try to set both configs and it should work for you as well:
> 
>     * connection.accountFilterFormat: `(&(objectClass=inetOrgPerson)(uid=%s))`
> 
>     * user.filter: `(&(objectClass=inetOrgPerson))`

Yes, it is working like that.

> I am working on improving the default configurations, so these problems go away.
> I'd like to know if it is better to use:
> 
>     * `posixAccount` with `posixGroup` or
> 
>     * `inetOrgPerson` with `groupOfNames`
> 
> 
> as default settings? From what you wrote above I understand that the use of posixAccount is discouraged.
> 
> My proposal is to use the defaults:
> 
>     * groups use `member`
> 
>     * accountFilterFormat `(&(objectClass=inetOrgPerson)(uid=%s))`
> 
>     * filter `(&(objectClass=inetOrgPerson))` (if a user is searched by name the `(uid=%s)` is automatically appended and that's what caused your problems)
> 
> 
> I could also imagine for max compatibility:
> 
>     * groups use `member`
> 
>     * accountFilterFormat `(&(uid=%s))`
> 
>     * filter ``
> 
> 
> What would you say?

It is hard to say.
I think, for OpenLDAP most common setup is:
`groupOfNames -> member -> inetOrgPerson`
but beside posix schema:
`posixGroup -> memberUid -> posixAccount`
there are other setups:
`groupOfUniqueNames -> uniqueMember -> inetOrgPerson`
And of course there is AD with:
`group -> member -> user`

I think there are too many possibilities to be smart here and default settings doesn't matter that much.
It is one-time setup of few values. One example for OpenLDAP and one for AD should be enough to make this easy enough.

kevinpapst:
So glad that it finally works...

> I think there are too many possibilities to be smart here and default settings doesn't matter that much.
> It is one-time setup of few values. One example for OpenLDAP and one for AD should be enough to make this easy enough.

Hmm, maybe you are right and I am overcomplicating things. 
I will update to use groupOfNames -> member -> inetOrgPerson and then update the docu.

army1349:
Great work. After last commit with accountFilterFormat not set and:
`filter: (&(ObjectClass=inetOrgPerson)(memberOf=cn=cloud,ou=groups,dc=example,dc=ch))`
everything works like expected.

kevinpapst:
Docu updated once again: https://www.kimai.org/documentation/ldap.html

@army1349 in that case the initial search will execute `SRCH base="yourBaseDn" scope=2 deref=0 filter="(&(uid=xxxx))"` - thats probably not what you want?!

army1349:
> 
> 
> Docu updated once again: https://www.kimai.org/documentation/ldap.html
> 
> @army1349 in that case the initial search will execute `SRCH base="yourBaseDn" scope=2 deref=0 filter="(&(uid=xxxx))"` - thats probably not what you want?!



> 
> 
> Docu updated once again: https://www.kimai.org/documentation/ldap.html
> 
> @army1349 in that case the initial search will execute `SRCH base="yourBaseDn" scope=2 deref=0 filter="(&(uid=xxxx))"` - thats probably not what you want?!

Why they are two searches and search filters anyway?

kevinpapst:
"Internals" - ZendLdap requires one query including the uid=%s for finding the users DN before the bind (user authorization).  And I need one without the %s as well for loading the user.
I tried to avoid that, but each of the tested solutions had some kind of drawback.
And then you wrote 
>It is one-time setup of few value

so I though "you are right, fuck it" - the users has to configure it properly once. And I already spent a lot of hours with that (failed) optimization. Some things sound so easy but then they aren't.
If I get a daily bug report about LDAP configuration, I will maybe re-think it and give it one more try. But for now all I want is to merge that thing :D

army1349:
> "Internals" - ZendLdap requires one query including the uid=%s for finding the users DN before the bind (user authorization). And I need one without the %s as well for loading the user.
> I tried to avoid that, but each of the tested solutions had some kind of drawback.
> And then you wrote
> 
> > It is one-time setup of few value
> 
> so I though "you are right, fuck it" - the users has to configure it properly once. And I already spent a lot of hours with that (failed) optimization. Some things sound so easy but then they aren't.
> If I get a daily bug report about LDAP configuration, I will maybe re-think it and give it one more try. But for now all I want is to merge that thing :D

Yeah, I can imagine, just one proposal:
What about taking user.filter and using it twice?
And just before pushing it to ZendLdap modify it like this:
`filterForZend = "(&$filter(uid=%s)"`
What do you think?

kevinpapst:
I have to use the configured "usernameAttribute" to build that. Do you know anything about ADs?
I didn't used that approach because I feared to break stuff for the Microsoft folks.

army1349:
> I have to use the configured "usernameAttribute" to build that. Do you know anything about ADs?
> I didn't used that approach because I feared to break stuff for the Microsoft folks.

Yeah, I know few things about AD. It is not that much different from OpenLDAP for basic usage like bind and searches. Just an LDAP implementation with few specifics and own preferred schemas.

kevinpapst:
One more "last" try ;-)

kevinpapst:
@army1349 your idea was implemented, don't know why it didn't work the last time, probably it was simply too late ;-)
If you guys give me a last ok, I will merge the LDAP integration.

And one more time: THANKS FOR ALL YOUR SUPPORT!!!!

army1349:
> @army1349 your idea was implemented, don't know why it didn't work the last time, probably it was simply too late ;-)
> If you guys give me a last ok, I will merge the LDAP integration.

Nice! It is working for me and correct search filter is in logs.
> And one more time: THANKS FOR ALL YOUR SUPPORT!!!!

My pleasure gentlemen. We did good :-)

kevinpapst:
Ha, 99 comments when it was finished while 99 files were changed:
Bildschirmfoto 2019-06-07 um 22 48 12

Nice one! Thanks @tobybatch thanks @army1349 !

army1349:
Very nice!
Feel free to write me, if you have any questions in future.

tobybatch:
@kevinpapst Thanks so much for this.I'm also here to support the LDAP stuff if needed.  Do you think we could have a 0.9.1 release including the LDAP auth?

kevinpapst:
@tobybatch Can't do that right now, as I found out (after merging) that the new ZendLdap package requires the php-ldap extension and therefor Kimail will not install on many systems, as the DLAP package is not included in default installs (my demo server included). I have to find a way around this.
Why do you need the tag? For testing?

tobybatch:
I'm only building containers against releases at the moment.  I could build us a container against a commit I suppose.  Let me have a think about the non-composer requirements too.

Do we have a list of supported architectures?  OSX (obviously). Linux, I guess supporting rpm and deb distros should be enough.  Do we try and support windows?

kevinpapst:
There are a couple of Kimai users on Arch (pacman?). But its, like Windows, not very high on my priority list. 
I would assume that it should run on any platform with PHP, there are no big problems nowawdays with cross platform compatibility. The only issue could arise from directory traversal with the wrong slash type. But I don't do that afaik outside of tests.
Just for the sake of testing you could fork Kimai and create your own release, but then you have to switch repo URLs and I doubt that its much easier than targeting a commit?!
Otherwise people have to wait for the next release, which will take presumably a couple of weeks... I have a couple of open cleanup tasks and possible BC breaks on my todo list for 1.0.

tobybatch:
@kevinpapst We'll probably wait for the 1.0 release.  I can test an arch / debian/ fedora release process if you want me to.  I can easily spin up VM for each of those and run from a clean install to running kimai.

army1349:
> 
> 
> @tobybatch Can't do that right now, as I found out (after merging) that the new ZendLdap package requires the php-ldap extension and therefor Kimail will not install on many systems, as the DLAP package is not included in default installs (my demo server included). I have to find a way around this.

This is normal. If someone wants LDAP functionality, php-ldap extension is required.
As long as it is documented and it works without extension when LDAP auth is disabled, there is no better alternative.

kevinpapst:
The code works without ldap extension. The problem is that the zend-ldap composer package requires php-zend extension instead of suggesting it. That leads to composer stop the installation if LDAP is not available ... one could use the `--ignore-platform-reqs` composer flag, but that is really bad pratice.

army1349:
Oh, I see. How about removing zend-ldap from kimai's composer?
One could simply:
`composer require zendframework/zend-ldap`
if he wants to use LDAP auth.

kevinpapst:
I will try, that would be the easiest way.
Lets hope the Symfony cache/autoloader will not complain about a missing dependency...

kevinpapst:
Found a solution which requires a bit more configuration for the initial setup.
May I ask you both again to test #846 ?

Issue #763, opened 2019-05-09
Docker refactor


tobybatch:
@kevinpapst I've had some spare time at work this week and have refactored the external docker repo to build against tags.  Also we now correctly export thr DATABASE_URL so the console works in the dev env.

Consequently the latest dev image will be stored here: https://github.com/tobybatch/kimai2/blob/master/Dockerfile

I hope that's OK.

kevinpapst:
Thanks for the update.
Could you please explain the consequences for the Dockerfile in this repo - is it outdated?

Oh and I see this line: https://github.com/tobybatch/kimai2/blob/master/Dockerfile#L16
If you find stuff like that, it would be nice if you report them - that's looks like a bug.

tobybatch:
The ownership of that file is a consequence of the user de-escaltion in the docker.  The phpunit is run by the www-data user.  I don't think it's actually a bug for you to worry about as you will run your tests as the same user that installed the stack.

And yes, the Dockerfile in the core repo is, or will, be out of step with changes in the docker reepo.  I don't expect these docker files to change very often.  Perhaps I can just ping you when they do?

tobybatch:
BTW Any help on where to build the bridge for this https://github.com/kevinpapst/kimai2/issues/193 would be appreciated.

The reason I've been a bit slack on the docker files is that my company needs LDAP auth in it's time keeping solution so I can't support the container in work hours if we aren't using it.  If I can help develop that bridge then I'll be able to justify time on issues.

kevinpapst:
I will have a look nevertheless, as nothing should ever write outside var/cache/.

And I will try to setup and LDAP and then we can discuss further in the other topic.

Yes please ping me on changes. But I still wonder what should happen to the Dockerfile in this repo? You want to manage the dev image also exclusively in your repo?
Shall I simply copy & paste your changes over here? And is there a benefit from using the Kimai base image of the Debian image used here?

tobybatch:
The bug only manifests when tests are run.

The creation of the base image was in response to one of the bugs in my repo where it was pointed out that I was basing the prod image on the dev image.  Basically I was relying on the dev image to set up the LAMP requirements.  This way I do the common set up in kimai_base.  That will only change when/if  I roll up a php version or os version.

I think that if you just pick up the dev docker image and drop that into the base repo for now.  We can see if there is a better way of doing this at a later date.

tobybatch:
[...]

PR #840, opened 2019-06-04
updated dockerfile

kevinpapst:
## Description

See #763 

@tobybatch can you please review?

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,40 @@
 #
 # For the full copyright and license information, please view the LICENSE
 # file that was distributed with this source code.
- 
-FROM kimai/kimai2_base
+FROM php:7.2.9-apache-stretch AS tmp_kimai2_base
+
+RUN apt update && \
+    apt install -y --allow-unauthenticated \
+        git \
+        haveged \
+        libicu-dev \
+        libjpeg-dev \
+        libldap2-dev \
+        libldb-dev \
+        libpng-dev \
+        mysql-client \
+        unzip \
+        wget \
+        zip \
+        && \
+    EXPECTED_SIGNATURE="$(wget -q -O - https://composer.github.io/installer.sig)" && \
+    php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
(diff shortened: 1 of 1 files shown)


tobybatch:
@bsumners Yes the refactor has happened.  The docker repo is now building against tagged releases, 0.9 is the latest.  https://github.com/tobybatch/kimai2 or if you just want the images https://cloud.docker.com/u/kimai/repository/docker/kimai/kimai2

@kevinpapst The dev docker in this repo will only change if we need to change the dependencies in it.  That will only happen if we change the OS/PGO set up.

kevinpapst:
So you are saying I shouldn't merge this, as it relies on your base image? 
Or are you saying this is likely the last change we will see in a while?

tobybatch:
Yes, that's fine.  The base image will get pulled from dockerhub if it's needed.  If you want to include it in the kimai2 repo I can add it.

kevinpapst:
I would totally like that!

tobybatch:
Sorry @kevinpapst my local kimai repo is all messed up with the LDAP stuff, here is a gist with the docker file that will build both stages of the dev docker.  I'll fix my repo to use this format for dev later.  If you drop that into your repo it will include the base build.

https://gist.github.com/tobybatch/902ee71f9814239e70680e6ba07b4c2a

kevinpapst:
Hm, I think we are talking about different things. I thought that you said in #763 that I should copy the new Dockerfile for the dev image into the Kimai repo.
But your gist looks like the Dockerfile which is already existing in here.

My assumption was, that I should copy & paste this Dockerfile:
https://github.com/tobybatch/kimai2/blob/master/Dockerfile
into the kimai 2 repository here. Was that a misunderstading?

tobybatch:
The gist should replace the one in your repo, totally.  It uses a build stage to in line the kimai base into the kimai dev image build process.  This means you have the whole build change in a single file and removes dependencies on external resources.

It is hand crafted for your repo at the moment.  I'll add a build process into the docker repos when I get a sec.

kevinpapst:
Alright, after having a look at the diff it is clearer, updated.

PR #846, opened 2019-06-10
fix LDAP install for systems without ldap extension

kevinpapst:
## Description

Fixes the requirement to the php extension "ldap".
LDAP users now have to execute `composer require zendframework/zend-ldap` in order to use LDAP authentication.

Updated docs online at: https://www.kimai.org/documentation/ldap.html#installation

See #815 

## Types of changes

## Checklist


army1349:
When I added security section to local.yaml, I got this:
```
Script cache:clear returned with error code 1
In PrototypedArrayNode.php line 313:
You are not allowed to define new elements for path "security.firewalls". Please define all elements for this path in one config file.
```

kevinpapst:
Hm, worked here flawlessly. Ok, I will research whats wrong and report back.

kevinpapst:
Seems to be a limitation of the Symfony framework. 
I could reproduce the problem, when placing the config in `configs/packages/local.yaml`.

Locally I store my custom config in the environment specific directory, so it doesn't cause problem when running my testsuite. The file `configs/packages/local.yaml` is always loaded, where the files `configs/packages/[dev|prod]/local.yaml` are loaded environment specific.

I yet don't understand why this should be any different, but could you please try to move your unchanged config to `config/packages/prod/local.yaml` and clear the cache again?

army1349:
> I yet don't understand why this should be any different, but could you please try to move your unchanged config to config/packages/prod/local.yaml and clear the cache again?

Yep, works fine like that.

kevinpapst:
Smells fishy to manipulate the loading order to get it to work, still searching for the "right way to do it".

This seems to be a longer story, as there is an open issue at Symfony regarding this topic:
https://github.com/symfony/symfony/issues/25021

Will work my way backwards through the Symfony discussions to find a solution...

kevinpapst:
should work now with `config/packages/local.yaml` as well

army1349:
> 
> 
> should work now with `config/packages/local.yaml` as well

Yes, I can confirm, it is working.

PR #860, opened 2019-06-16
export tags in spreadsheets

kevinpapst:
## Description

This PR adds a new column to all spreadsheet exports with a comma separated list of the records tags. ping @rarvin

If you used an Excel export with custom macros to calculate data, you have to adjust the columns, as a new column was added in the middle.

## Types of changes

## Checklist

### DIFF
--- a/src/Export/Renderer/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Renderer/AbstractSpreadsheetRenderer.php
@@ -14,6 +14,7 @@ use App\Repository\Query\TimesheetQuery;
 use App\Twig\DateExtensions;
 use App\Twig\Extensions;
 use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Style\Border;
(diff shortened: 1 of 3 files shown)


rarvin:
Thanks for doing this. This would be of great help I am able to see the data in the export separated by commas. 

Additionally this is on the timesheet html manual table export  -- 
Would it be possible for you to add a comma or a delimiter like semi colon in the html table that you generate in the timesheet view so that when one copy paste the data to an excel sheet the tag text would appear with a comma or some delimiter after each complete tag word. Since currently for tags text spaces are allowed and when one copies and pastes the data to an excel sheet its difficult to differentiate between one complete tag with space and the other and so on.

Example - 

Description | Tags
test arvin | arvin arvin2
Here arvin and arvin2 are two different tags hence they would not have any issues when we delimit the text 

Description | Tags
test 2 tag words with space in them | arvin with spacearvin_without_space

Here both tag words have merged together and also the tag data with space would delimit into multiple columns when they are to be considered as one. I hope you understand the problem one would face.

If you need a new issue to be raised I could do that too.

PR #862, opened 2019-06-18
fix/LDAP multivalues Email Aliases

jacoblanco:
## Description
This fix checks if LDAP Directory has multiple values into Email value, like multiple Email Aliases.
If detects multiple values, gets first value of the array.

## Types of changes

## Checklist

### DIFF
--- a/src/Ldap/LdapUserHydrator.php
+++ b/src/Ldap/LdapUserHydrator.php
@@ -66,7 +66,7 @@ class LdapUserHydrator
         if (is_array($user->getEmail())) {
             $user->setEmail($user->getEmail()[0]);
         }
-        
+
         if (null === $user->getEmail()) {
             $user->setEmail($user->getUsername());
         }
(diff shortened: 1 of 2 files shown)


src/Ldap/LdapUserHydrator.php
```diff
@@ -63,6 +63,10 @@ public function hydrateUser(User $user, array $ldapEntry)
 
         $this->hydrateUserWithAttributesMap($user, $ldapEntry, $attributeMap);
 
+        if (is_array($user->getEmail())) {
+            $user->setEmail($user->getUsername()[0]);
```
kevinpapst: that doesn't look right

kevinpapst:
Additionally this sentence
> I verified that my code applies to the guidelines (composer kimai:code-check)

should point you in the direction why the build failed ... `composer kimai:codestyle-fix`.

src/Ldap/LdapUserHydrator.php
jacoblanco: So sorry Kevin, you are right, typo mistake. 
Changes requested sent.
Thanks a lot !

kevinpapst:
Thanks @jacoblanco 👍

PR #864, opened 2019-06-18
Removed lines that skip plugins in test env

tobybatch:
## Description
This enables tests that are stored in plugin bundle.  I don't think t needs a documentation update.

## Types of changes

## Checklist

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -83,12 +83,6 @@ class Kernel extends BaseKernel
             }
         }
 
-        // do not load Kimai plugin in test environment, they may alter the default behaviour and create
-        // false-negatives in integration/system tests
-        if ('test' === $this->environment) {
-            return;
-        }
-
         $pluginsDir = $this->getProjectDir() . '/var/plugins';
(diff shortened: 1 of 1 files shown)


tobybatch:
@kevinpapst A clean checkout, composer install and run tests all pass with these lines removed.  I'm goinf to raise a second PR that has a switch to include/exclude pugins via a flag.

kevinpapst:
Oh yeah, that works. The problems shows up when your plugins alter specific features, eg. an existing tests check the amount of entity actions and your plugins adds one => the core test fails now...

But probably its better to rewrite the core (integration) tests and make them more stable for those cases.

kevinpapst:
If someone stumbles upon this issue: you can deactivate your plugin while tests are running with:
```php
class MyBundleExtension extends Extension implements PrependExtensionInterface
{
    public function load(array $configs, ContainerBuilder $container)
    {
        if ('test' === $container->getParameter('kernel.environment')) {
            return;
        }
        // otherwise load your services as usual
    }
}

```

PR #865, opened 2019-06-18
Switch for plugin tests

tobybatch:
## Description
Allows tests in plugins to be included/excluded using a switch in the phpunit.xml.dist

## Types of changes
- [ X Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -29,6 +29,9 @@
         <!-- ###+ nelmio/cors-bundle ### -->
         <env name="CORS_ALLOW_ORIGIN" value="^https?://localhost(:[0-9]+)?$"/>
         <!-- ###- nelmio/cors-bundle ### -->
+
+        <!-- Toggle this to include/exclude plugins in tests -->
+        <env name="PHPUNIT_INCLUDE_PLUGINS" value="0" force="true"/>
     </php>
 
     <testsuites>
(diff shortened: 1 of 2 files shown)


tobybatch:
@kevinpapst I'll close the othe rone, I mis-understood

tobybatch:
I think this is OK.  There is reliance on the plugin developer to accept passing failing tests in core if the plugin messes with them.  I think that's ok.

kevinpapst:
Yeah, you say that now ... that was my initial thought as well and then I wasted multiple times "hours" by searching for the damn bug, which wasn't really existing :-D

kevinpapst:
Btw. I'd rather start with removing it altogether and see where we go from there.
Having a PHPUNIT env check in the Kernel feels a bit awkward.
I'll try here with a couple of plugins and see if the errors still exist.

tobybatch:
I tested the other PR with your CSS plugin and my chrome extension one and they both passed OK.

kevinpapst:
Thanks for testing! I have a couple of none published plugins for customers and others which I plan to sell soon, I'll test them as well. Shall we re-open the other PR then? I really like deleting the if better than adding a new env.

tobybatch:
Yes, we should.  And close/delete this one if you want to.

kevinpapst:
Closing in favour of #864

PR #930, opened 2019-07-08
Support tag creation from query params

tobybatch:
## Description
Currently we can specify star and end times in the query params.  This adds tag support as a comma separated list.

## Types of changes

## Checklist

### DIFF
--- a/.github_changelog_generator
+++ b/.github_changelog_generator
@@ -1,4 +1,4 @@
 unreleased=true
-future-release=1.0
+future-release=1.0.1
 exclude-labels=duplicate,question,invalid,wontfix,release
 enhancement_labels=>enhancement,Enhancement,feature request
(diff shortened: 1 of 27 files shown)


tobybatch:
Sorry Kevin, I have no idea how to fix this?

> 1) App\Tests\Controller\TimesheetTeamControllerTest::testCreateAction
> ArgumentCountError: Too few arguments to function App\Controller\TimesheetAbstractController::create(), 4 passed in /home/travis/build/kevinpapst/kimai2/src/Controller/TimesheetTeamController.php on line 75 and exactly 5 expected

kevinpapst:
The AbstractController is extended by two Controllers:
https://github.com/kevinpapst/kimai2/blob/master/src/Controller/TimesheetTeamController.php#L73
you have to adjust that method as well.

Is it necessary to persist the tags before actually creating the record?

tobybatch:
I thought the record didn't get a id before it was persisted.  I'll try without.

PR #941, opened 2019-07-09
polyfills for old browser

### LINKED ISSUE #929: Activity drop-down not listing data in Kimai v1.0 for older browsers but works well in new
We had migrated to Kimai v1.0 from v0.9 today and in some of the systems using older browsers, after selecting the customer & project details, the activity dropdown field is coming as empty. This was not happening in the previous Kimai version(0.9) on the same old browsers. Having said this, for the newer version of browsers, all works fine. We're having to use old Mac systems running the maximum supported Firefox versions as they are loaded with old Adobe software which will not work on the new OS and is very expensive due to Adobe's subscription rates. As over 13 operators are unable to make use of Kimai due to this, after the upgrade today, we request you to kindly help us find a solution. 
      
While testing out different features in version 1.0, we added a new global activity. Strangely enough, that alone is getting populated in the activity list.

All the other existing functionalities along with the new ones like 'Tag' option are accessible to the operators. They're also able to edit the existing entries but are unable to add new ones due to this problem. Similarly, they are unable to restart the existing task also and it throws up an xml exception. Screenshot of both the errors are attached.

kimai-1
kimai-2

**Desktop/Smartphone**
Please complete the following information if this seems to be a frontend issue:
 - Device: MAC OSX 10.5
 - Browser [Mozilla Firefox 29.0]

kevinpapst:
## Description

a test to fix crappy old browsers

fixes #929

## Types of changes

## Checklist


rarvin:
@kevinpapst ran the test but it does not seem to help. Still not able to start and stop tasks. I think since the browsers are too old you could very well close this bug as supporting browsers that are really old might be very tough. 

We are trying to upgrade our OS from 10.5 to 10.6 which helps us move to Firefox 45. We have tried some testing on Firefox45 and could confirm that all works well without any buy fixes etc. This could also help give you a baseline from what browsers Kimai works well. It does work well Firefox 45. Sorry for the trouble

kevinpapst:
Yeah, I really don't want to support such old browsers, but if multiple users (especially the ones who care about the community)  have troubles I want to help... It just complicated, because I have no way of verifying this problem.

If you want we can still try, I don't think that it is too complicated to fix it.  Just need the errors from the browsers JS console.

rarvin:
Thanks Kevin, the users have the old browser have left for the day. I can only get in touch with them on Monday. The users are not too technical work from home which is like 300kms from where I am. So possibly I shall try my best to get the details from them on Monday. Kindly provide me some time.

rarvin:
Please find attached the logs.
[log_ff23.txt](https://github.com/kevinpapst/kimai2/files/3393424/log_ff23.txt)

kevinpapst:
Thanks @rarvin - I guess I need help from someone who knows these JS pitfalls with old browsers better. I'll leave this PR open, hopefully someone stops by and lends us a hand...

sideeffect42:
@rarvin Did you have a look at http://www.floodgap.com/software/tenfourfox or https://github.com/wicknix/Arctic-Fox or https://sourceforge.net/projects/leopard-webkit?

kevinpapst:
Closing, sorry but I donÄt want to invest any more time into that...

PR #996, opened 2019-07-31
Team permissions

### LINKED ISSUE #366: Role and access management on customers and projects
An in my  Opinion essential Feature would be to Assign different Role or Access Right per Project or Customer.

Letting ever Employe see all Customers und Projects could cause some serrios securut Issues

### LINKED ISSUE #765: Grouping users in teams
We would like to group users and create reports in the dashboard by the group the user is assigned to. Currently under the admin user it shows the summed up summary of all users within the enter kimai platform, what would be nice is an option to show a summary by also the various groups (like department, team, agency or internal or all) that are present on the platform. This filtering and summation summary can really help.

Summarised view by various groups in the dashboard as well as in the filtering pages that are already present. Currently in the dashboard a summarised view of all users on the platform are reported. Having an option to summarise and show by department, team or agency or internal can help. 
Similarly the filter option allows to filter by user, customer, project, activity. Having an option to filter by various teams that are present can really help. If nothing is chosen it would be all users on the platform. In addition there must be an option to assign a user to multiple teams. For example we have users who are company resources who work from office and some who work from home, in this case they are to be part of the company team and also have to be part of the work from home team or office worker team. Summarised view of the time for such team users can really help. I think currently you have the pervlige mapping similar to this for a user where is a user can be a team lead a user as well as admin etc all together and individually. The idea is similar where this is for teams or groups.

None as of now.

**Additional context**
None

### LINKED ISSUE #876: Admin Timesheets Toolbar: order users by username
The select input form field to filter users is order by id.

It would be much more convenient to order the users list by username.

**Additional context**
Affects administrators only.

### LINKED ISSUE #972: Minor: Currency upon NewProject
if
`Create customer - default values  >> CURRENCY != EUR`
then
Creating a new project will show wrong currency (EUR), (but will take the right one anyway)

Hope this is not annoying, I just protocol stuff I find. OK?
If I get the chance I'd do a checklist. But I'm on another mission right now.

### LINKED ISSUE #1017: Change/Set Currency Globally for Customers, Users, etc.
I work out of the US and while I have clients in different countries I only bill in USD. I can see how to set USD as the currency for Customers but not for a Users hourly rate. E.G. when I set a rate for a user I can only set it in EUR.

It would be nice to have a way to set the default currency for the entire system (all places where currency is relevant) so that we have control over the currency for Users and other options/areas as well as Customers.

Being able to do this through local.yaml would be sufficient for my needs as this has worked well for other settings/overrides.

Sorry if there is a way to do this that I have missed (if so how?).

kevinpapst:
## Description

Adds:
- Teams (group users in a team)
  - Create named teams 
  - Assign customer and projects to the team
- New screen to assign teams to a customer
- New screen to assign teams to a project
- Access to customers and projects can be limited to teams
- Customer & projects without team assignment are visible to every user
  - If a customer or project has at least one assigned team, only the assigned team members can see them
- Super-Admins see everything, nothing is limited
- New API endpoints to fetch and delete teams

Related issues:

- Fixes #366 
- Fixes #765
- Fixes #876 
- Fixes #972 
- Fixes #1017 

Can be tested at: https://demo-branch.kimai.org/ with the normal demo accounts.

## Open points (maybe new issues for later) 

- [ ] Check sqlite migrations
- [x] add tests for new controller methods
- [x] add unit tests
- [x] delete teams

## Create new issues for later

- [ ] New team filter in toolbars
- [ ] Team Reporting
- [x] Rework teamlead permissions

## Types of changes

## Checklist


HeinzWuert:
Right to change members of a team should have also the Admin.
Most work will be done by Kimai Admins!

rarvin:
If one is assigned as a team lead for a team then he should be able to see
only the records for the users associated with that specific team, like in
timesheets, invoices, exports etc. This would be a nice add on.--
Regards,
Aravind

BeckeBauer:
By introducing teams, a single user should no longer have the right to see the customers but should need to be grouped into a team first and gets its permissions according to the team permission (or at least, there should be an option 'only team permission').

For the creation of teams, there should be an option 'all customer'. So, the admin can decide if he /she creates a team with access to all customers and could then delete the right to access certain customers for this team or creates a team with no access rights and then adds customers to the team.

BeckeBauer:
I understand that the permissions 'view_customer' or 'start_own_timesheet' are still general permissions even if the user is assigned to a team. So, the user is still able to see all customers and can start records booked on any customer disregarding his/her status as team member.

If this understanding is correct, maybe it would be an idea to introduce another role 'ROLE_TEAM_USER' who has limited rights to act only within his team (see only team customers, start only recordings for team customers). The advantage is that the team restrictions could then be applied easily on the front-end.

kevinpapst:
Don't understand. The user can only see customer for the teams he is assigned to.
Please open a new issue

rarvin:
@kevinpapst currently I could see that for a team a teamlead is assigned. And when the teamlead logs in he would be able to see all of the users under that said team listed for filtering in timesheets export etc. Is my understanding correct. 

If so how could other team leads see for the same team. Can multiple teamleads be assigned to a team.

If we can assign only one team lead then, Could you please think of creating something called Groups which could be specific for users. Multiple users could be assigned to a group, and this group would be made the team lead of the team. When a user is created he is himself a part of his group so if one needs only one team lead then the group of that user needs to be mapped. Similary a different group which has various users could be assigned as the owner or team lead of the team.

This way all users under the group can log in and see respective users timesheets do export etc. 

I may be wrong in my understand, hence please guide. Right now I have not started assigning users to team after deploying 1.2 so thought I understand things before I proceed. If currently if I am able to assign only one team lead so for a different team lead to see the team records would mean that i created the very same team and assign a different team lead and so on.

kevinpapst:
Multiple team leads are not supported for now. 
If you need that, you have to create multiple teams, as you already guessed. 
Or assign the Admin role to users who need to see more users.
Use the demo to test how Kimai behaves when using teams.

PR #1002, opened 2019-08-01
Fix various minor English grammar translations

trentks:
Hi @kevinpapst,

Only very minor mistakes, don't feel bad. English is a horrendously complicated hodge-podge of a language that borrows from about 50 different other languages, and is woefully inconsistent and hard to navigate. It truly is probably impossible to understand unless it is your first language and you've been exposed to it all your life—and even then it's still not spoken or written well by the vast majority of speakers. I myself still learn new little things all the time that I've been doing wrong all my life.

## The following minor grammar errors were found and fixed

1. Administration/System Menu items pluralisation made consistent. Only some of them were pluralised, {"Timesheets", "Customer", "Projects", "Activities", "Tags", "User", "Plugins", "Settings"} ->  {"Timesheets", "Customer**s**", "Projects", "Activities", "Tags", "User**s**", "Plugins", "Settings"}
2. "Amount Users", "Amount Projects" "Amount Activities"
3. "All user" to "All users" on the dashboard.
4. "Records of all user" to "Records of all users" on the admin timesheet.
5. Exchanged "count up to" for "sum up to" in several places. There's nothing really wrong with "count up to" but when dealing with fractional groups of data which you're adding/summing, rather than physically counting in series, it's less awkward to say "add up to" or "sum up to" as you're performing the act of adding/summing rather than actually counting. Counting is the process of having independent, discrete objects and then tallying them... "1 apple, 2 apples, 3 apples, ... 10 apples". I hope this explanation makes sense, it's a really vague difference that's somewhat difficult to articulate. Counting and adding/summing are very similar acts, but are different.
6. "All registered user of your Kimai installation" -> "All registered user**s** of your Kimai installation" in admin: user
7. "Revenue total" to "Total revenue". I can't even/won't even attempt to explain this one, and I'm a native English speaker. Probably best explained as simply being conventional.
8. Changed "used to data export" to "used to export data". In english the verb "action/doing word" comes before the noun "object/place/activity". So "export" is a verb, and "data" is a noun. Confusingly however "data export" can be a noun itself—so you could say "I'm going to quickly do a data export". But in this context only "data" is the noun, so "export" is a verb, and so should precede the noun.
9. "Name Consultant" to "Consultant's name" and similar for "Customer". "Consultant" is always a noun, and in this case so is "name", however. English is a bit weird in that sometimes you can join two nouns and get another noun. For instance "dog bowl", or "computer bag" etc. In both of these instances it would be equally valid to also write "dog's bowl" or "computer's bag" if you were talking about a particular dog, or computer whose bowl, or bag it was. It would however never be acceptable to write "bowl dog", or "bag computer" unless you had a bowl that was a dog, or a bag that was a computer. So similarly here a consultant owns a name, a name does not own a consultant. Then you've got this really strange concept in english where suffixing with `'s` means ownership, where without the apostrophe it means pluralisation. So "Consultants name" would be nonsensical... but "Consultant's name" means there is one consultant and he has a name. This gets extra confusing when you realise that `'s` can also be suffixed to a word as a contraction for "is" or "has" for instance... "the consultant is changing his name" could be contracted to "the consultant's changing his name". Note: It would also be perfectly valid to write "Name of consultant" rather than "Consultant's name".

## Uncertain, so left unchanged

1. `label.tag` = "Tags", as well as `label.tags` = "Tags".
2. There exists both `label.hourly_rate` and `label.hourlyRate` which map to the same "Hourly rate". Is one of these two unused? Every other entry is in snake_case, not camelCase—should the camelCase one be refactored?
3. `label.color` is currently "Color", this is the American spelling of the world, and the near universal adoptation for programing variables. However, like a lot of American english spellings, this is not consistent with most english dialects—where it is spelled "Colour". Not sure if we're writing in American English or British/Australian/New Zealand/Canadian/etc English so I have left it.

## Types of changes

## Checklist


kevinpapst:
Super cool PR - I really appreciate the long explanation 👍 
You don't have to explain every change, even though I totally enjoy reading them! 
Much nicer than just changing them, so I can learn new stuff ... or lets say it brings back memories from my education. Most of the information isn't really new but there is a difference between passive knowledge and active usage. 
Reading is easy, but writing is much harder. Especially when there is not the chance for constant practice with the spoken language.

I'll leave that PR open for a couple of days, maybe you find more changeable translations in the other xlifff files as well?!

trentks:
Sorry, literally editing from my phone in bed on the github interface. Should've just run them on my own branch from my fork, rather than off your master. Bit funky in the branching there.

I think I fixed everything in the .en translations. I'm on holidays for the next week.

trentks:
The big one that needs explaining there is the em dashes. Commit message copied below.

In many languages, not just English you have em dashes—which are used pretty versatile to join thoughts into the same sentence but may be somewhat different thoughts. 

You have en dashes which are used to separate ranges, I.e 6–7 or Monday–Tuesday. En dashes are narrower than emdashes.

Then finally you have hyphens which are the narrowest of all, these are used to concatenate-words-to-form-conjuntions.

Unfortunately a standard keyboard only has hyphens, and using en or em dashes can be difficult in some systems (read: windows). On linux systems you have something called a 'compose' key which you can hit and then key in really easy to remember combinations. You might map right ctrl as your compose key, or perhaps caps lock. Emdash is `compose ---` endash is `compose --.`, the compose key is really good for make characters with different accents too, like ò would be `compose o<grave>` or ü would be `compose u.`.

kevinpapst:
Can you please change this line as well:
https://github.com/kevinpapst/kimai2/blob/master/tests/Controller/PluginControllerTest.php#L31

The test is bound to the english translation (which is a bit stupid...) and currently checks for `You have no plugin installed yet`. 
You changed it to `You have no plugins installed yet`, so you need to add the plural `s` in the test as well.

kevinpapst:
Please let me know when you are done and I will merge!

trentks:
Yeah should be done there @kevinpapst, if you're happy with it, then it's good to merge.

kevinpapst:
Absolutely! and a new PR can always be opened if you find more weird wordings ... which I strongly believe you will/could ;-)

Thanks 👍

PR #1030, opened 2019-08-13
Copy meta fields

### LINKED ISSUE #1015: auto copy current values if autocopy flag is set for fields created though metafields
Currently when the activity is stopped and subsequently restarted, the meta filed values are not copied automatically which is good. But there are instances where for certain fields the value entered needs to be retained. For example like location. It would be good if while creating the field itself a flag is provided which can indicate if the filed has to be auto-copied when restarting an activity.

This way one can have fields which autocopy on restart and ones which dont copy the values. More like what you have done for the description and tags fields.

This would greatly help since asking the user to enter all of the fields again while they are restarting an activity is tedious. Then again there are fields that need to be entered on each restart too.

kevinpapst:
## Description

Copy timesheet meta-fields on restart action (as with all fields).

Fixes #1015

## Types of changes

## Checklist

### DIFF
--- a/src/API/TimesheetController.php
+++ b/src/API/TimesheetController.php
@@ -409,8 +409,6 @@ class TimesheetController extends BaseApiController
      *      required=true,
      * )
      *
-     * @Security("is_granted('delete_own_timesheet') or is_granted('delete_other_timesheet')")
-     *
      * @param int $id
      * @return Response
      */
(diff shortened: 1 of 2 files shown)


rarvin:
@kevinpapst Thanks for delivering this. But could you please add an addition flag option when creating the metafield so that autocopy only happens if this additional flag is set. This way the user could choose which ones need mandatory copying and which ones need not be copied at restart. Would enhance the existing functionality greatly.

PR #1046, opened 2019-08-18
refactored invoice and export forms

### LINKED ISSUE #373: Mark time entries as invoiced
Every time-entry needs an "invoiced" flag, which can be set manually or while generating an invoice.

### LINKED ISSUE #649: Export - Mark all as cleared button doesn't update
**Describe the bug**
If you mark multiple entries as "cleared" (or "open") with the button in the upper right of the export view and then edit the filters to view another set of entries, the button does not update.
If e.g. you filter all entries of one customer last month, mark them all as "cleared" and then change the filters to this month (with all entries "open"), the button still has a "Mark as open" functionality.

**To Reproduce**
1. Go to kimai-demo
2. In export view set filters to e.g. one month and one customer
3. use "set all cleared" button
4. Change filters (e.g. to another month, containing open entries)
5. Now, if you want to "clear" them all, you have to press the button twice: First time it "opens" them all, second time it clears them.

**Additional context**
Maybe it's not really a bug, but more kind of unwanted behaviour (to me). Even if you change the filters from "only open" to "only cleared", the button still has a "clear" function.

I think it would be nice to check if the majority of the entries is open or cleared and set the button to the other function, every time the filter changes.
Alternatives would be: Two buttons (one for clear, one for open) or one button with "toggle" functionality

### LINKED ISSUE #882: Exporting - Minutes output or a Summery of hours when filtering
When using the export function, the way time is currently presented (##:## h) is very hard to use in excel (or similar). 

I'm proposing, another column is added (or can be set up) to just export minutes, and with a decimal point and not character suffix. This would mean the column could be selected and summed very easily.

An alternative, would be to have a summary of the hours/minutes performed once a filter has been added in the export (eg Project A, between date B and date C).

### LINKED ISSUE #887: Configurable default date-range for initial filter in export, invoice and team-timesheets
Currently when clicked on the export view it defaults to load the data present in Kimai for a month. Due to the presence of large number of records it takes quiet a lot of time for the page to load. I have seen this more with chrome than Firefox. In most cases we want to just see a days worth of data for exporting but can’t choose any options till the whole page loads which take like around 3 minutes. Could you please change the way export page works so that no data is displayed by default or it just shows only one days worth of data so that the page loads quickly. 

Also the filter options works or starts fetching data when one option is selected, in our case we would like to have multiple filtering by a specific user and for a specific day. The view starts fetching when one option is itself selected and one has to wait for the results to be shows before we could select the next filter criteria. Could you please change this since when there is considerable data in the system it could be very slow.

Change default export to just one day as default or not display anything till chosen.
As well as change the selection to accept all filter options before starting to fetch records. Even if the first option is done it would be good for us.

kevinpapst:
## Description

Export screen:
- use horizontal form for selecting filters
- do not show preview by default
  - added preview button/action (limited to 500 records)
- option to mark all filtered timesheets as exported when creating the export
- choose multiple users to filter timesheets for
- added option to filter running records 
- allow to choose all filters without data being reloaded - fixes #887 
- improved export documents - fixes #882 
  - added duration (seconds) column to export
  - using native date values and format (for Excel export)
  - added start date column without time
- removed ODS renderer, as it produces wrong data (due to missing formats) and the generated Excel is fully compatible with OpenOffice/LibreOffice

Invoice screen:
- use horizontal form for selecting filters
- do not show preview by default
  - added preview button/action (limited to 500 records)
- option to mark all filtered timesheets as exported when creating the invoice - fixes #373
  - using the same "export state" than the export screen
- choose multiple users to filter timesheets for
- allow to choose all filters without data being reloaded - fixes #887 

Can be tested at: https://demo-branch.kimai.org/
Usual [demo logins](https://www.kimai.org/demo/) can be used, eg. `susan_super` / `kitten`

Closes #649 

## Types of changes

## Checklist


kevinpapst:
@rarvin this is how horizontal forms look like
Bildschirmfoto 2019-08-19 um 14 28 45
Can you have a look at the demo installation: https://demo-branch.kimai.org/en/invoice/

I will apply the same changes for the export as well. It will solve your problem with all the data pre-loading and the reloading table when changing filters.

BeckeBauer:
I checked the excel export file: It is great that there is a separate "Date" field and that the fields "Date", "From" and "To" are now recognised as dates by excel. 

However, I do not understand why there is a field duration after the date fields which shows seconds (like 26280) and another field duration as second last row which shows the duration in hours (like 07:18 h). 

The first duration entry can be directly used for calculation though (in my case, I would add a row, configure it as ```hh:mm "h"``` type and apply the formula ```=D2/86400``` to show hours which can be used for calculation). 
The last field cannot be used for calculation by excel (in my case, I would add a row configure it as ```hh:mm "h"``` type and apply the formula ```=LINKS(N2;5)*1``` to show hours which can be used for calculation). 

An optimal solution would be do show the duration in hours which can be directly used for calculation but I do not know if it is possible to pass to excel the command to format a row as ```hh:mm "h"``` type. If not, maybe the division by 86400 can be already applied so that the first duration would not show seconds but hours (one would then only need to apply the row formatting).

kevinpapst:
> However, I do not understand why there is a field duration after the date fields which shows seconds (like 26280) and another field duration as second last row which shows the duration in hours (like 07:18 h).

The first row (seconds) is new, if you need them for calculation. The other one is mainly a left-over, as it was existing before. Its meant for a quick manual check without the need to do calculations.
I could replace it with a column as you suggested, using `[h]:mm "h"` and `=D2/86400`.

I can set formulas and formattings, thats not a problem. I already did it for the first three columns.

My biggest fear is, that Excel treats these formulas/formattings language dependent and some users might not be able to see anything. Not sure about it, becuase I don't have Excel, but I have read quite a lot about it.

All formulas must take into account, that a record can be longer than 24 hours, even weeks (for holidays). I couldn't get that working properly (maybe because LibreOffice and Numbers use a wrong interpretation of these Excel specific formattings?).

Can you please rephrase shortly what you suggest, like that:
- remove column x
- apply formatting abc to column d
- add formula to column d
Otherwise I might misunderstand you.

I don't use these exports at all, so I am really happy about any input. And about any tip about Excel, because I don't use that as well...

kevinpapst:
@BeckeBauer I installed the latest changes, can you please do a (hopefully) last re-test of the excel export and check the data and if it is now better usable for calculations?!

BeckeBauer:
Basically, it works (data beside names are recognised by Excel as data and not text anymore). Just some suggestions:
1. First row (date) is exported with the English format ```JJJJ-MM-TT``` no matter which language is used in Kimai - if it were possible, the date_type from kimai.yaml according to the chosen language should be used
2. Second and third row: Just the time would be sufficient ```hh:mm``` instead of ```JJJJ-MM-TT hh:mm``` at least in most user cases it would be problematic though if the time recording would comprise of 2 or more days)
3. Fourth row: For duration, I personally favor the "h" at the end (but this might be confusing for other languages)
4. Fifths row: The rate should have two decimal places (instead of 45,5 better 45,50) to have a uniform appearance with other entries like 79,17 - pretty cool would be if the currency shown in the sixth row would be applied to the rate directly without an extra-row, e.g.  79,17 EUR (Excel format ```#.##0,00 [$EUR];-#.##0,00 [$EUR]```)
5. Regarding the order: I preferred to have the rows "duration", "rate" (and "currency") after the "hourly rate" and "fixed rate" rows but I see that at least "duration" fits also next to the "dates" rows. 
6. The last two rows "hourly rate" and "fixed rate" should also have two decimal places (75,00 instead of 75) since they are also currency values. If the currency could be shown (see remark 4) it should be applied also to these values (e.g. 75,00 EUR).
7. If it were possible I would not need to see a row without data. If the exported data for instance are calculated on basis of an "hourly rate", the row "fixed rate" could be invisible (but maybe it gives the impression that the exported data are incomplete).
Do not get it wrong, these are only suggestions. The changes (from text values to data) already applied are really helpful!

kevinpapst:
Some feedback after a last round:
1. Please create a new feature request for that (too much work to cover it here, I want to finish this one first)
2. Changed. The date is still there are can always be shown. I guess for most cases, only showing the time is sufficient
3. I keep it for now as hh:mm
4. I applied a style where the currency is left aligned and the money amount on the right + two decimals
5. My thought is that the duration is more important to know at first glance, so I moved it further to the beginning
6. Done (same as 4) but empty values will produce completely empty cells
7. We could discuss if these two columns could be skipped completely (maybe make the exported columns configurable in the future => new feature request if someone wants that). But calculating if a fixed rate was applied in any of the exported rows is too expensive. Works for 40 entries, but not for 35000 rows.

Thanks for all the feedback!

BeckeBauer:
👍 Great

Stadly:
Nice work! Can you update https://demo-branch.kimai.org/ to include the latest commits? Then I can ask my co-worker to test if the changes she is looking for are covered here.

kevinpapst:
@Stadly done. I am going to merge today, so please use the slot for feedback if possible. Otherwise new feature requests have to be planned for the future.

PR #1059, opened 2019-08-22
prepare release 1.2

kevinpapst:
## Description

Last preparations for 1.2

Can be tested here: https://demo-branch.kimai.org/
Login with usual [test accounts](https://www.kimai.org/demo/) (eg. `susan_super` / `kitten`)

## Checklist


kevinpapst:
Everybody is invited to do some final testing on the 1.2 release (link above).
If we don't find any big issue, this will be released on the 28 August.

PR #1158, opened 2019-10-06
Bugfixes for Kimai 1.4.1

### LINKED ISSUE #1157: Maybe: Conflict as Teamleader with missing user list
**Describe the bug**
Not sure, but I think a teamleader who can "edit_team" is not given a full list of users in the Team page and therefore will overwrite membership if saved there.

Or I couldn't find out what would let me get a list of users.
(Or I didn't grasp the concept yet)

If members-change-not-possible is intended, then the teamleader should not be given the right to save those.

**Additional context**
V 1.4

kevinpapst:
## Description

Fixes #1157 
Fixes https://github.com/tobybatch/kimai2/issues/32

## Types of changes

## Checklist

### DIFF
--- a/src/Controller/DoctorController.php
+++ b/src/Controller/DoctorController.php
@@ -48,11 +48,8 @@ class DoctorController extends AbstractController
     /**
      * @var string
      */
-    protected $projectDirectory;
+    private $projectDirectory;
 
-    /**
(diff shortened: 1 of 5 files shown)


kevinpapst:
Awesome, then we will have a full feature set, thanks 👍

PR #1201, opened 2019-10-29
fix avatar path for subdirectory installations

### LINKED ISSUE #1196: Avatar url broken
**Describe the bug**
The avatar is a nice new feature. The problem is, when kimai is installed as subfolder of the top level domain, e.g example.com/kimai2. Then, avatars won't be shown, because the url will resolve in example.com/avatars/<hash>.png instead of example.com/kimai2/avatars/<hash>.png

**To Reproduce**
1. Go to example.com/kimai2/de/admin/user/
2. You won't see the avatars, only broken images of the users
2. Right-Click on one of the broken images
4. You will get to an url, like example.com/avatars/<hash>.png
5. Insert 'kimai2', like example.com/kimai2/avatars/<hash>.png
6. Avatar will show up.

**Additional context**
 - Kimai version kevinpapst/kimai2 | 1.4.2.0@de533e7031ae194b35f6a638cd95f5c94c516796
 - PHP version  7.3.1

kevinpapst:
## Description

Fixes #1196

## Types of changes

## Checklist

### DIFF
--- a/config/packages/assets.yaml
+++ b/config/packages/assets.yaml
@@ -1,3 +1,6 @@
 framework:
     assets:
         json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
+        packages:
+            avatars:
+                base_path: 'avatars'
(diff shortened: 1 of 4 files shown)


kevinpapst:
As there was no feedback, I can only hope that it works...

PR #1240, opened 2019-11-12
Update and delete multi timesheets and tags

### LINKED ISSUE #1100: Multi select and delete tags in one go
#1083 gave us the option to create tags via the tags page. Having an option to multi select and delete the tag entries in one go would be a good management feature for tags. Similarly multi select could be used for triggering other features like merge tags etc.

kevinpapst:
## Description

This PR adds the base logic for handling updates on multiple items at once.

- adds a "delete multiple" functionality to the tags management view
- adds a "delete multiple" functionality to the timesheet user + admin views
- adds a "update multiple" functionality to the timesheet user + admin views

Fixes #1100 

Can be tested at https://demo-branch.kimai.org/

Use `susan_super` / `kitten` or one of the [demo accounts](https://www.kimai.org/demo/)

## Types of changes

## Checklist


kevinpapst:
ping @wdenx @rarvin up for testing ?

Can be tested at https://demo-branch.kimai.org/
Use `susan_super` / `kitten` or one of the [demo accounts](https://www.kimai.org/demo/)

wdenx:
I can select several entries, but it is not clear where to select an "Action" - using one of the selected entries will then only operate on this.  Is it correct to assume that only the selection is implemented yet?

rarvin:
Carried out testing by creating two new enties and then checking the boxes and then using actions clicked on delete. I could see among the two entries the actions from where the delete is clicked is the only entry that is getting deleted.

rarvin:
allowed the entry to be active as well as stopped, the effect was the same, only one entry was getting deleted. Like what wdenx said since I couldnt find a common action button used the one from the existing entries and action for that single item is what is getting triggered and not for the multi selected items

kevinpapst:
The "Action" selection for "all checked items" is below the table, like in many other web-apps. So you need to scroll down.
I don't know how to communicate that differently. Any idea how to make it clearer?

rarvin:
:) in the tags view it i was able to see the action when i did a sear by the tagnames and i was able to do a test and it worked well with the multi delete.

rarvin:
I tried the one at the bottom and it worked well. Instead of the bottom can it be moved to the top, this way as soon as the check box is checked one would know its there

rarvin:
Can it be just like the existing action button but appears at the top above all entries so that look and feel wise its just like the existing screen and one does not have to scroll down also. If its difficult I could be happy with this too. Once users get used to it, it should be okay. Since i couldnt find it I was thinking since its multiselect etc any of the buttons should do the same action to all of them.

rarvin:
In the edit view, if i select something by mistake like a customer name and then I would like to change it back to the old one and to do this when i hit reset, nothing happens.

kevinpapst:
Putting it above the table would lead to a "jumping page" and considering you click a checkbox very deep down in the table, it would be unclear as well where it is.

> Can it be just like the existing action button but appears at the top above all entries so that look and feel wise its just like the existing screen and one does not have to scroll down also. 

But I like that idea 👍  Maybe always visible, but deactivated until the first checkbox is ticked. I will give it a try.

The "reset" button doesn't work on any form with the javascript selects. Thats a problem unrelated to this PR ;-)

rarvin:
@kevinpapst would it be possible to add the description field also to multi edit. Tested the multi edit functionality and it worked well, I am able to add tags, change customer, project, activity, user all works very well. What is missing is deletion of tags assigned, is there any option to clear all tags and then to use the multi edit option to add the required tags alone. This would be helpful along with the change description field to a new value (overwrite description to a common one)

wdenx:
Trying to edit 2 itens, changing only the "Project" entry, results in error: "Missing activity"

rarvin:
> Trying to edit 2 itens, changing only the "Project" entry, results in error: "Missing activity"

I dont think that is an error, since you are changing the project you must choose an activity from it. once you do this it should be okay. I tried it and it worked well. Not sure if this is related to a global activity. I have not tested that scenario.

rarvin:
Tried the global activity scenario also and it works well. I think all functionalities are working as needed. Please choose an activity for the project that you chose and the problem should go away. I trieid with a global activity where in i didnt change the project or customer and it updated the entries with the same customer and project and to the changed global activity. All looks good now.

kevinpapst:
Will leave the action box where it is for now. Moving it requires more effort than expected.
I hope ever user will eventually find it and get used to it after the first try. Everything can be improved later on...! Thanks for testing and feedback 👍 

@rarvin I added a radio button to choose whether to add or replace selected tags.

rarvin:
@kevinpapst i was carrying out the test and I couldnt find the radio button that you had added to add or replace selected tags. Please let me know if I am trying something wrong.

kevinpapst:
It was merged to master, so I didn't reinstall the demo. 
If a PR was merged, you can always use the "development version" demo at https://demo.kimai.org

rarvin:
thanks Kevin, I would keep note of this. I have done the testing also and it works very well.

Is there any chance of adding the description field also with an append and replace option. Atleast the replace option. This would indeed be very helpful.

kevinpapst:
You know how to create new feature requests 😀

PR #1262, opened 2019-11-19
Select2 search enhancement

choules:
## Description
The search of Select2 options now also matches the parent's name in a nested list. With this it is now possible, for example, to search the project list by the name of a customer – just as with the project list in Kimai1.

## Types of changes

## Checklist
- [ ] I verified that my code applies to the guidelines (`composer kimai:code-check`) – **just front-end changes**
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation)) – **not relevant**


kevinpapst:
Awesome 🎉  I already had that in mind and the matcher article open. Great contribution!

Will test later

choules:
After thinking a bit more about this functionality: Perhaps the select box for the customer may now be obsolete, since we are able to search for customers inside the project field.

We could think about a smarter matching function which allows multiple search params to enable combined searches for customer and projects, for example "CustomerXYZ ProjectABC".

kevinpapst:
That would be awesome, because the customer box tends to confuse some users. They think the customer could be assigned (especially when working with global activities). They don't understand that it is just a filter in most forms.

But in environments with many customers, it is easier to filter the projects by customer first, so I would make the box optional via configuration. 

What I would really appreciate: if the box could filter by multiple strings.
For example: having such a filtered result set
Bildschirmfoto 2019-11-19 um 14 11 17
I cannot type anything more, as even a space would now clear the result list completely.

But if the search term would be split by space, then we could match all single search terms to the result set. I guess thats what you proposed above as well?!
Do you want to have a look at that?

choules:
> But if the search term would be split by space, then we could match all single search terms to the result set. I guess thats what you proposed above as well?!

That's exactly what I wanted to suggest. I will have a look at it.

choules:
Some further enhancements are now implemented:
- support for multiple, space separated search params
- for nested option lists:
  - return parent and all children if parent contains all search params
  - return parent and those children, which contain all search params, even though the parent has no match at all
  - return parent and some children, when all search params are contained within the concatenation of parent and each child
- for not nested option lists:
  - return only those options which contain all search params in any order

The following examples should illustrate the functionality:

### Example Options from a Select2 input
- Company Ltd
  - Development Project X
  - Support Contract
  - Other
- Enterprise SE
  - Meta Project
  - Consulting
  - Other

### Queries and expected results
#### "Company"
- Company Ltd
  - Development Project X
  - Support Contract
  - Other

#### "Company Lt"
- Company Ltd
  - Development Project X
  - Support Contract
  - Other

#### "Other"
- Company Ltd
  - Other
- Enterprise SE
  - Other

#### "Project"
- Company Ltd
  - Development Project X
- Enterprise SE
  - Meta Project

#### "Enterp Project"
- Enterprise SE
  - Meta Project

#### "Project Meta"
- Enterprise SE
  - Meta Project

#### "Enterprise Development"
- (no results)

kevinpapst:
Like it very much! So much better than before 👍 

You can search for funny strings, but that is by design :)
Bildschirmfoto 2019-11-22 um 15 10 52

Can you merge master and rebuild the assets once more, then I will merge.

kevinpapst:
You denied push access for me in this PR. 
I'd like to release 1.6 this weekend, so it would be great if you could:
- merge & rebuild  & push
- or give me the permissions to push to your branch (all done locally)
- or let me know if you are currently not available, then I will merge it via a new branch

kevinpapst:
I added these changes via #1276 

Thanks for your contribution @choules 🎉

choules:
Sorry, I missed your comment - thanks for your work!

Issue #1080, opened 2019-09-04
Using the variable ${entry.duration_decimal} and ${invoice.duration_decimal} should always have 2 decimals 


JGrotepass:
With Change #1037 new variables for invoices were introduced. At least in my case the format of these variables are without 2 decimal characters. That means, that a value of "7.00" will be printed as "7", a value of "7.50" will be printed as "7.5". 
This results in a invoice report that does not look very nice. 
Is it possible to have both variables formatted with 2 decimal characters?
Thanks
Jochen

Here a screenshot:
![image](https://user-images.githubusercontent.com/54812367/64257635-fa349380-cf25-11e9-98fc-359dc869e809.png)

gises:
Hello, Kevin, 
There are different number formats depending on the country. Is it possible to take this into account as well?
For example, Siwtzerland and Germany are a bit different. 
![image](https://user-images.githubusercontent.com/54983006/64421038-efead480-d0a0-11e9-8aaa-c747e03d234e.png)

Thanks for your effort!

kevinpapst:
I think we can add `de_CH` as locale. Please open a new issues for that, as this is not related to invoices. Can you provide a PR / all the required information mentioned here (in the new issue):
https://www.kimai.org/documentation/translations.html#configure-locale-formats

kevinpapst:
You can simply change your invoice template.
Set the format for the column / cells to be a number with two decimals.

PR #1272, opened 2019-11-22
Translation: Add de_CH

aptevo:
## Description
This pull request adds de_CH or Deutsch(Schweiz) to the available languages.
I had not to change any de translations for this, these seem ok for Switzerland too.
This also fixes the issue with some number separators.

## Types of changes

## Checklist


kevinpapst:
Awesome, thanks 👍 

Did you try if the language files are really necessary? I haven't tried yet, but would assume that de_CH falls back to use de if not available.

labor4:
What... no dialect??
What about de_CH_zh_stereotypeedition?

just joking. ;-)

kevinpapst:
This PR does actually does "nothing". All translations are 100% like the German version. 
If I delete all Swiss translation files Kimai will use German as fallback.

Did you read https://www.kimai.org/documentation/translations.html#adding-a-new-language ?

I worked on the PR and pushed some changes, but all configs are also the same.

Besides that you can now select "Deutsch (Schweiz)" in the user profile, I don't see any benefit in this PR. Nothing changes compared to `de`. I see only one effect: the Javascript gets larger.

> This also fixes the issue with some number separators.

What do you mean? What is fixed and how is it fixed if there is no difference to German?

aptevo:
sorry for the "non-work". the cause for the added language is the java script. without having de-ch, we get wrong number formats, see  https://github.com/kevinpapst/kimai2/issues/1080#issuecomment-528800717 and your answer for that.

This changes the number format in invoices, and on the whole front end from ie. 1.637,90 to 1'637.90.

Reading the docs for me it seemed that I have to add a "complete" language to achieve this , sorry if i was wrong about that. if there is a better solution, just close this PR, and I will try again.

kevinpapst:
Aha, so the number formats are fixed. Well, thats a good argument!
Please have a look at the changed config - the date formats are the same than in de I assume?

aptevo:
Yes, the date formats are also the same. 
I just re-checked the translations again and found
messages.de.xliff:                <target>Schließen</target>
which should be "Schliessen" for de_CH. I did not find this yesterday, seems this was not my day at all ;-).
It is the only ß I found, so maybe just change it in de, so an extra messages.de_CH.xliff is not needed?

kevinpapst:
One can simply create a messages.de_CH.xliff file and only put that one translation into it. The nice thing is that the fallback to de still works. 

I just did, you can see how it works if you want to change other translations in the same manner:
https://github.com/kevinpapst/kimai2/pull/1272/commits/bdc5e41c099069cad760c6116600cb6dfd26b4b9

aptevo:
great, thanks for the information! the doc is not that clear about that, it states in "Adding a new language":
> Copy each translation from it’s english version translations/*.en.xliff and rename them to translations/*.xx.xliff.
> 
> Adjust the target-language attributes in the file header, as example for the new file exceptions.xx.xliff:

I will write an update to the doc and create another PR for this.
thanks again!

kevinpapst:
Well, thats a special case because of the fallback from de_CH to de. Technically I wouldn't consider de_CH a "new" language, even though I am sure you would say something else ... especially when it comes to the spoken words 😀 

Normally the fallback is always English. There are just a couple of translations were this special handling would be used (pt and pt_BR, de and de_CH/de_AT, some chinese "dialects" ...). 
The common case is the one described in the docs. 

But: yes, please go ahead with a docs PR! Its always good if someone actually using the docs reads and corrects it. When I write it, I already know what to do and therefor it might not include all necessary information.

aptevo:
I would describe de_CH not as a new language, but as a lets say "very special" variant ;-)

Is the "special handling" automatically used for every variant added, so if I added an fr_CH the fallback will be fr?

kevinpapst:
Yes, the fallback is always the language code (fr) without the region code (CH).

kevinpapst:
You can test the integration at https://demo-branch.kimai.org/de_CH/timesheet/ 
with username `susan_super` and password `kitten`

aptevo:
looks good. what's strange is that the number format is from ch-de even when the language setting is Deutsch and not Deutsch(Schweiz) - it seems as at least the number format on the frontend and the invoice derives from the browser language settings, if the browser language exists in in kimai JS - doesn't this make the settings in Kimai a bit inconsistent?

kevinpapst:
All formatting should be exclusively taken from the current requests locale.
And this is the one in the URL (not necessarily the one configured in the users profile).
Please check if you mistakenly used multiple tabs without changing the locale in the URL.
The users profile language is only used after the initial login, to redirect to the correct language.

It's harder to spot if the translations don't change.

At least this changes the money formatting correct for me:
- https://demo-branch.kimai.org/de_CH/timesheet/
- https://demo-branch.kimai.org/de/timesheet/

Can you please check again if you still experience the problem describe in detail what is wrong and in which place.

aptevo:
I checked again with lots of different settings of the browser and the application and you are absolutely right. It must have been that I accidentally reused an url from the browser history in one of the tests, all works perfectly as intended, sorry for the false alarm.

kevinpapst:
No need to be sorry ;) qualified feedback is always very welcome!

PR #1384, opened 2020-01-16
API methods to handle teams and users

### LINKED ISSUE #1331: User and Team API
Hi,

I would love to see Kimai being capable of not only exposing user and team data over API, but to also allow creation/modification of these entities through REST. I am aware of a very similar issue posted some time ago - #977. However the author wasn't able to point to any use-case, other than a preference of REST over e.g. LDAP of someone on his team.

As for myself I actually do have a couple of reasons for wanting to create users and teams with API:
1. The nature of the users of Kimai in my organization - they are not employees strictly speaking, so their information cannot not be kept in user directories we use (rather in CRM). Making a new directory just for them would be a possibility, but it's not so trivial if you take GDPR into consideration - we are trying to reduce the number of databases containing personal data, and creating an intermediate directory, used just for the integration purposes seems like an overkill. Still, we would love to automate it as much, as we can.
2. I haven't found any information on how to create/modify teams and maintain membership in them using LDAP. Also, afaik API allows only getting and deleting of teams - there are no methods for creating/modifying them. There are also no methods for adding and removing members. This one is crucial, because we need to control the access to different customers' data - specifically, a user having access to one project, should not be able to even see any other customers, projects etc.

I have thought about using OS-level commands for user creation, but this still leaves the issue of the teams, and not being able to maintain those in any way.

Would it be possible for the addition of the create/modify/delete methods for users and teams to be reconsidered as a new feature in one of the next releases?

Thanks!

kevinpapst:
## Description

Planned methods:
- [x] create team
- [x] update team
- [x] add member to team
- [x] remove member from team
- [x] grant access to customer/project
- [x] revoke access from customer/project
- [x] create user
- [x] update user

Fixes #1331 

## Types of changes

## Checklist


kevinpapst:
@kbancerz this PR is ready for testing, want to jump in?

PR #1418, opened 2020-01-28
added pagination to working time widget

kevinpapst:
## Description

added new widget: paginated working time with chart

Can be tested at http://demo-branch.kimai.org/  eg. with user: `susan_super` / pass: `kitten`

## Types of changes

## Checklist


lowrob:
I like it:
It was only showing 6 days but that was another problem and that is fixed here.

Noticed WORKING HOURS THIS MONTH goes to 0 when you navigate to then end of one month / start of the next month - I like the 0 / it is easy to explain that you are not on a month - you are really between months - this brings up a possible bit of a redesign of the Graph.

Consider removing the WORKING HOURS THIS MONTH and the WORKING HOURS THIS YEAR values from the graph widget.  Treat the graph as a weekly only view.   The by month and this year could be moved to a section of their own.

kevinpapst:
I fixed the 0 in split month views and the 0 in split year views. I think its better than the zero (which will definitely be reported sooner or later.

I agree, that all numbers besides the "this week" are less helpful now, but moving them to its own widget is a new feature request / change for me.

kevinpapst:
@lowrob re-deployed for a last test

lowrob:
> @lowrob re-deployed for a last test

Looks good, does what we need for our group of users to start with as we move to this version.  There may be feature request coming your way when we deploy to the broader user group (March 1st is the target) but this is what we asked for.

PR #1420, opened 2020-01-29
added symfony/http-client as requirement for composer

hmr-it-jr:
## Description
Added symfony/http-client as requirement for composer because I need this to make sure a new plugin from me will work properly.

## Types of changes

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
         "symfony/flex": "^1.0",
         "symfony/form": "^4.0",
         "symfony/framework-bundle": "^4.0",
+        "symfony/http-client": "4.3.*",
         "symfony/intl": "^4.0",
         "symfony/monolog-bundle": "^3.4",
         "symfony/orm-pack": "^1.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -44,6 +44,7 @@
         "symfony/flex": "^1.0",
         "symfony/form": "^4.0",
         "symfony/framework-bundle": "^4.0",
+        "symfony/http-client": "4.3.*",
```
kevinpapst: To strict, must be `"symfony/http-client": "^4.0",`

McBochi:
> do you need that for a public or private plugin? for private plugins, there is another way to achieve that.

I found the same requirement missing when I tried to enable transactional mails via Mailgun.

kevinpapst:
Oh, thats good to know. In that case the library will be included.
Thanks for the feedback!

McBochi:
> Oh, thats good to know. In that case the library will be included.
> Thanks for the feedback!

Not sure why, though... The included `symfony/mailgun-mailer` just states it as development dependency and I assumed this dependency would be resolved by composer during the initial installation. However, when enabling Mailgun as mail service everything stopped working and a manual install fixed it.

kevinpapst:
Closing. I fix it in #1545

PR #1424, opened 2020-01-30
improve rate rounding

kevinpapst:
## Description

This PR is work in progress, a lot of places need to tested and probably adjusted if that should go live.

**Changes**
Allow split cents when calculating rates -> rounding to the third place after the decimal separator instead of the second place (eg 12.346 instead of 12.35)

**Background info**
Currently a rounding problem can occur:
1 entry with 30 min (hourly rate = 27,50) = 13,75
but
2 entries with 15 min each (hourly rate = 27,50) = 6,88 * 2 = 13,76

With the change other problems rise up, like the calculation of the total sum in invoices.
Each single rate will be still rounded in the frontend to the second place and then we might get a total that differs from the sum of all singles. 
Why? Because the frontend would show each of the 15 min entries rounded as 6,88 but the total will be calculated from the internal values 2 * 6,875 => 13,65

I think this is worse than before, because currently all values are consistent (even though you could argue that the rounding is wrong in some cases). With the change the values within the invoice are inconsistent - I don't know which one is better or worse from a legal perspective. 

ping @wdenx @HeinzWuert

Not sure about the future of this PR, help is needed.

## Types of changes

## Checklist


wdenx:
Dear Kevin,

In message <kevinpapst/kimai2/pull/1424@github.com> you wrote:
> 
> Allow split cents when calculating rates -> rounding to the
> third place after the decimal separator instead of the second
> place (eg 12.346 instead of 12.35).

Does this really help?

> **Background info**
> Currently a rounding problem can occur:
> 1 entry with 30 min (hourly rate = 27,50) = 13,75
> but
> 2 entries with 15 min each (hourly rate = 27,50) = 6,88 * 2 = 13,76

This canot be avoided.  The same happens always when rounding.

> With the change other problems rise up, like the calculation of
> the total sum in invoices.
> Each single rate will be still rounded in the frontend to the
> second place and then we might get a total that differs from the
> sum of all singles ... which is worse than the current behavior.

It is not worse, the problem just shows up at a different place.

> Why? Because the frontend would show each of the 15 min entries
> rounded as 6,88 but the total will be calculated from the internal
> values 2 * 6,875 => 13,65

I don't have a solution.  With standard rounding rules
("kaufmännisches Runden") this is unavoidable.  There is something
called "summenerhaltendes Runden", but I cannot find any hints that
this is actually used anywhere or standardized in any public rule set
(Kaufrecht or Handelsrecht or any DIN or EU norm or such).

> Not sure about the future of this PR, help is needed.

Sorry...

Best regards,

Wolfgang Denk

-- 
DENX Software Engineering GmbH,      Managing Director: Wolfgang Denk
HRB 165235 Munich, Office: Kirchenstr.5, D-82194 Groebenzell, Germany
Phone: (+49)-8142-66989-10 Fax: (+49)-8142-66989-80 Email: wd@denx.de
Nearly everyone is in favor of going  to  heaven  but  too  many  are
hoping  they'll  live  long  enough  to see an easing of the entrance
requirements. Never appeal to a man's "better nature." he  might  not
have one.

kevinpapst:
Thanks for your feedback! I added a sentence:
I think this is worse than before, because currently all values are consistent (even though you could argue that the rounding is wrong in some cases). With the change the values within the invoice are inconsistent - I don't know which one is better or worse from a legal perspective.

gises:
Hello Kevin

is an exciting topic and I handle it with my customers as follows:

If the client requests a billing according to individual time entries, these will be shown separately on the invoice and the total amount of the individual time entries will be invoiced. So 2x15 min a 6.88.

Usually, however, in the project environment it is more likely that at the end of the month the total hours of a project or activity are shown as one sum. The invoice then shows 1x30min a 13.75.

If the individual time entries have to be countersigned by the customer, no hourly rates and amounts are shown on the time recording sheet. In most cases, this is not appropriate anyway, because the signatories are often not allowed to know who is billing how much, or if this falls into the wrong hands, there will only be unnecessary discussions :-)

I would appreciate it if no change in the calculation is made at this point. You might consider to hide the field "Wage" in the function "My Times" via the settings and not user-dependent and thus also hide this information during the PDF export.
Or, if the function "My times" and the function "Export" are used and the function "Totals calculation" is activated like the function "Invoice", the time sheet can be generated identically to the invoice and this discussion does not exist. This would also be easier or less "confusing" for project controlling.

Regards Michael

kevinpapst:
Thanks for the feedback!

I won't change the rounding for now => closing

PR #1517, opened 2020-03-03
fixed sweet alerts blocking major ui elements

adrianrudnik:
## Description

The current SweetAlert setup did block major UI elements for an "annoyingly" long period. The upper right part became un-usable for every notification you trigger, like starting time and wanting to stop immediately.

## Types of changes
- [ ] Moved the alert two and a half element down on mobile devices.
- [ ] Moved the alert one and a half element down on desktop devices.

## Checklist

## Remarks

Did not run `composer kimai:code-check` as this is a scss change only.

## Screenshots

medium-to-large
small-1
small-2

### DIFF
--- a/assets/js/plugins/KimaiAlert.js
+++ b/assets/js/plugins/KimaiAlert.js
@@ -43,7 +43,7 @@ export default class KimaiAlert extends KimaiPlugin {
         Swal.fire({
             timer: 2000,
             toast: true,
-            position: 'top-end',
+            position: 'top',
             showConfirmButton: false,
             type: 'success',
             title: message,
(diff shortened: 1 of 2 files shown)


kevinpapst:
Seems we had the same idea: https://github.com/kevinpapst/kimai2/pull/1508/files

But I haven't thought about the top position. I'll give it a try the next days

adrianrudnik:
Yeah, it's somewhat bad to place it anywhere. It gets in the way of the most time. For mobile maybe a bottom-bar would be OK if the body gets a global margin/padding to allow to fully scroll to the bottom. On desktop though I'm kinda lost, just don't have it over buttons would be great!

Didnt see the other pull you already did but cool that a patch is coming!

kevinpapst:
thanks @adrianrudnik 👍

Issue #1512, opened 2020-03-02
Excel export with table format


yayitazale:
**The exported excel archive is not formatted as a table**

**Solution**
Format the excell as a table, with titles and so, This is the guide to do it manually: https://support.office.com/en-us/article/Format-an-Excel-table-6789619F-C889-495C-99C2-2F971C0E2370

Thanks

kevinpapst:
Which export exactly? 
And I have no Excel for testing, so I need dev help with that one.

yayitazale:
Export > Excel

![imagen](https://user-images.githubusercontent.com/6462849/75674284-102cd200-5c85-11ea-93ce-fd867d739df2.png)

This is hat we get:
[kimai-export-original.xlsx](https://github.com/kevinpapst/kimai2/files/4275207/kimai-export-original.xlsx)
And this is the same but with table format:
[kimai-export-titles.xlsx](https://github.com/kevinpapst/kimai2/files/4275208/kimai-export-titles.xlsx)

kevinpapst:
Maybe this is related: https://stackoverflow.com/questions/50934401/phpspreadsheet-multiple-tables-with-format-as-table-in-excel-worksheet

HeinzWuert:
Maybe this would solve the problem with loosing the autostart of the macro when creating the excel sheet.

yayitazale:
> 
> 
> Maybe this is related: https://stackoverflow.com/questions/50934401/phpspreadsheet-multiple-tables-with-format-as-table-in-excel-worksheet

You only need 1 table format as table.

kevinpapst:
@HeinzWuert this is the export, which creates the excel dynamically, instead of parsing an existing template file.

@yayitazale yes, but I have no idea how to apply that. if anyone knows how to set that style on the table, I am happy to integrate it into kimai.

yayitazale:
I found this, maybe it can be usefull:

https://github.com/PHPOffice/PhpSpreadsheet

kevinpapst:
Yes and no. 
Yes: this is the library I am using to generate the excel files.
No: I couldn't find any documentation how these table styles can be applied.

kevinpapst:
now available in master

PR #1518, opened 2020-03-03
xlsx improvements (#1512)

adrianrudnik:
## Description

Small improvements for the xlsx export as in part requested in #1512. Changes should be fitting to every export that contains a single header row.

## Types of changes

- [ ] Feature: Activated "Auto Filter" feature (as #1512 misnamed it as "table format")
- [ ] Feature: Freeze first row and column for scrolling through larger reports.
- [ ] Feature: Auto size the columns so content fits on screen.

## Checklist

## Remarks

The `composer kimai:code-check` command did not complete successful:

- `phpstan` did reach the memory limit during execution and had to be run with parameter `--memory-limit=512M`. Error message was `PHPStan crashed in the previous run probably because of excessive memory consumption. It consumed around 110 MB of memory.`
- `phpunit` did fail due to `Remaining self deprecation notices (35)` and `Remaining direct deprecation notices (9)`.

after that point I skipped running the code-check.

### DIFF
--- a/src/Export/Base/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Base/AbstractSpreadsheetRenderer.php
@@ -464,6 +464,12 @@ abstract class AbstractSpreadsheetRenderer
         $spreadsheet = new Spreadsheet();
         $sheet = $spreadsheet->getActiveSheet();
 
+        // Set default row height to automatic, so we can specify wrap text columns later on
+        // without bloating the output file as we would need to store stylesheet info for every cell.
+        // LibreOffice is still not considering this flag, @see https://github.com/PHPOffice/PHPExcel/issues/588
+        // with no solution implemented so nothing we can do about it there.
+        $sheet->getDefaultRowDimension()->setRowHeight(-1);
+
         $recordsHeaderColumn = 1;
         $recordsHeaderRow = 1;
 
--- a/src/Export/Base/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Base/AbstractSpreadsheetRenderer.php
@@ -288,7 +288,7 @@ abstract class AbstractSpreadsheetRenderer
                 if (!$isColumnFormatted) {
                     if (null !== $maxWidth) {
                         $sheet->getColumnDimensionByColumn($column)
-                            ->setWidth(40);
+                            ->setWidth($maxWidth);
                     }
                     $isColumnFormatted = true;
                 }
--- a/src/Export/Base/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Base/AbstractSpreadsheetRenderer.php
@@ -74,7 +74,10 @@ abstract class AbstractSpreadsheetRenderer
         'customer' => [],
         'project' => [],
         'activity' => [],
-        'description' => [],
+        'description' => [
+            'maxWidth' => 50,
+            'wrapText' => false,
+        ],
         'exported' => [],
         'tags' => [],
         'hourlyRate' => [],
@@ -264,8 +267,31 @@ abstract class AbstractSpreadsheetRenderer
         }
 
         if (isset($columns['description']) && !isset($columns['description']['render'])) {
-            $columns['description']['render'] = function (Worksheet $sheet, int $row, int $column, ExportItemInterface $entity) {
-                $sheet->setCellValueByColumnAndRow($column, $row, $entity->getDescription());
+            $maxWidth = array_key_exists('maxWidth', $columns['description']) ? intval($columns['description']['maxWidth']) : null;
+            $wrapText = array_key_exists('wrapText', $columns['description']) ? (bool) $columns['description']['wrapText'] : false;
+
+            // This column has a column-only formatter to set the maximum width of a column.
+            // It needs to be executed once, so we use this as a flag on when to skip it.
+            $isColumnFormatted = false;
+
+            $columns['description']['render'] = function (Worksheet $sheet, int $row, int $column, ExportItemInterface $entity) use (&$isColumnFormatted, $maxWidth, $wrapText) {
+                $cell = $sheet->getCellByColumnAndRow($column, $row);
+
+                $cell->setValue($entity->getDescription());
+
+                // Apply wrap text if configured
+                if ($wrapText) {
+                    $cell->getStyle()->getAlignment()->setWrapText(true);
+                }
+
+                // Apply max width, only needs to be once per column
+                if (!$isColumnFormatted) {
+                    if (null !== $maxWidth) {
+                        $sheet->getColumnDimensionByColumn($column)
+                            ->setWidth(40);
+                    }
+                    $isColumnFormatted = true;
+                }
             };
         }
 
--- a/src/Export/Base/XlsxRenderer.php
+++ b/src/Export/Base/XlsxRenderer.php
@@ -39,6 +39,18 @@ class XlsxRenderer extends AbstractSpreadsheetRenderer
             throw new \Exception('Could not open temporary file');
         }
 
+        // Enable auto filter
+        $sheet = $spreadsheet->getActiveSheet();
+        $sheet->setAutoFilter('A1:' . $sheet->getHighestColumn() . '1');
+
+        // Freeze first row
+        $sheet->freezePane('B2');
+
+        // Auto size columns to fit at least the headers
+        foreach (range('A', $sheet->getHighestColumn()) as $column) {
+            $sheet->getColumnDimension($column)->setAutoSize(true);
+        }
+
         $writer = IOFactory::createWriter($spreadsheet, 'Xlsx');
         $writer->save($filename);
 
--- a/src/Export/Base/XlsxRenderer.php
+++ b/src/Export/Base/XlsxRenderer.php
@@ -11,9 +11,12 @@ namespace App\Export\Base;
 
 use PhpOffice\PhpSpreadsheet\IOFactory;
 use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Style\Alignment;
 
 class XlsxRenderer extends AbstractSpreadsheetRenderer
 {
+    protected const COLUMN_DESCRIPTION = 'J';
+
     public function getFileExtension(): string
     {
         return '.xlsx';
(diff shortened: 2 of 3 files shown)


adrianrudnik:
lets hope this sails better with the ci :)

src/Export/Base/XlsxRenderer.php
```diff
@@ -39,6 +39,18 @@ protected function saveSpreadsheet(Spreadsheet $spreadsheet): string
             throw new \Exception('Could not open temporary file');
         }
 
+        // Enable auto filter
+        $sheet = $spreadsheet->getActiveSheet();
+        $sheet->setAutoFilter('A1:' . $sheet->getHighestColumn() . '1');
+
+        // Freeze first row
+        $sheet->freezePane('B2');
```
kevinpapst: I donÄt get why that is B2, but its seems to work^^

src/Export/Base/XlsxRenderer.php
```diff
@@ -39,6 +39,18 @@ protected function saveSpreadsheet(Spreadsheet $spreadsheet): string
...
+
+        // Auto size columns to fit at least the headers
+        foreach (range('A', $sheet->getHighestColumn()) as $column) {
+            $sheet->getColumnDimension($column)->setAutoSize(true);
```
kevinpapst: Did you test how that looks, when a (ver)y long description is used?

src/Export/Base/XlsxRenderer.php
adrianrudnik: It's this weird cell logic of spreadsheet apps.

You want to lock something? It wants to know the upper left corner that is locked. Never understood it mayself 100%. I hope B2 is okay with all possible exports.

src/Export/Base/XlsxRenderer.php
adrianrudnik: Ah good idea, nope not yet. Just squash it away, may be a better idea to lock it at the first row but I need to dig into that if excel supports that. As the implementing class is pretty generic it would be hard to inject proper "max width and wrap content" rules into it I think.

kevinpapst: I'd be very happy to get some help with that, as I have neither much experience with spreadsheets nor am I using it myself. 
I have implemented height/line-wrap for invoices [recently here](https://github.com/kevinpapst/kimai2/commit/72b73bc9a690435034e6613f44be0b7494eedbc0). Maybe something like this could be done for exports as well?

kevinpapst: `autosize` didn't work properly for me, maybe I did something wrong... or does it only works for the width and not for the height?

adrianrudnik: Sure, I don't see why not. You use a specific renderer from the looks of it for invoices, right? I must dig a bit deeper into it. Just remove it from this one when you squash and I will take a look on thursday, I got a bit more time and brain there. Never used the invoices feature myself, will take a look into it.

PhpOffice is quite complete, I did not find any case where it did not sport a specific feature. On the other hand creating xlsx exports are a pain as you either have to throw around letter-coordinates or x/y based coordinates that are not fully feature complete in all functions.

adrianrudnik: autosize is only for width sadly. If you want to handle height `$sheet->getDefaultRowDimension()->setRowHeight(-1);` might be a better solution.

adrianrudnik: Gotta go now, here are some hints I collected for myself, maybe they help or give some clues what you could do as well. Got my through localization issues we had with german / english exports.

```
// Force text
$sheet->getStyle(PHPExcel_Cell::stringFromColumnIndex($index))
	->getNumberFormat()
	->setFormatCode(PHPExcel_Style_NumberFormat::FORMAT_TEXT);

// Proper number
$sheet->getStyle(PHPExcel_Cell::stringFromColumnIndex($index))
	->getNumberFormat()
	->setFormatCode('#,##0_-');

// Proper date
$sheet->getStyle(PHPExcel_Cell::stringFromColumnIndex($index))
	->getNumberFormat()
	->setFormatCode(PHPExcel_Style_NumberFormat::FORMAT_DATE_XLSX14);

// Fill with color
$style->getFill()->setFillType(PHPExcel_Style_Fill::FILL_SOLID)->setStartColor(
	new PHPExcel_Style_Color('00FFA200')
);

// Join multiple columns in first row of headers to group them for the second row that is split into more detailed rows (like logistic 1st row, pake width / height / depth in 2nd row).
$style->getFill()->setFillType(PHPExcel_Style_Fill::FILL_SOLID)->setStartColor(
	new PHPExcel_Style_Color('00C3C3C3')
);
$style->getFont()->setColor(new  PHPExcel_Style_Color(PHPExcel_Style_Color::COLOR_BLACK));
$sheet->mergeCellsByColumnAndRow($col, $row, $col + 4, $row);
```

adrianrudnik:
Bit more work on it, take a look if it fits. Screenshot and example attached based on the dev fixtures. J12 has a break text, maybe implement some larger fixtures for the description column as well?

[kimai-export (52).xlsx](https://github.com/kevinpapst/kimai2/files/4294761/kimai-export.52.xlsx)
Selection_012

adrianrudnik:
> And I have several request for changing th export and allowing to dynamically select the exported columns, maybe that could be combined. Let me think about it for a while.

My simple thoughts about it: Keep it clean and simple.

Just from my mind I would love to work with something like this instead of fiddleling with `AbstractSpreadsheetRenderer` and `XlsxRenderer`:

```php
class ExcelExporter extends BaseExporter {
    public function export(array $in, ExportSettings $settings) {
        // Modifiers need to implement ExportModifierInterface
        $this->addModifier(new Excel\FormatDateTimeModifier([
            'column' => 'start_date',
            'format' => $settings->dateFormat
        )]));

        $this->addModifier(new Common\MarkdownToTextModifier([
            'column' => 'description'
        ]));

        $this->addModifier(new Excel\SetFormulateModifier([
            'add' => true,
            'column' => 'total',
            'formula' => '=SUM(%ROW%%COL%)'
        ])

        // this could also live in BaseExporter and we would just use
        // parent::export($in, $settings) somewhere at the start to apply
        // generic rules.
        if ($columns = $settings->getHiddenColumns()) {
            foreach($columns as $column) {
                $this->addModifier(new Common\ColumnVisiblityModifier([
                    'column' => 'responsible_person'
                ]));
            }
        })

        $spreadsheet = new Spreadsheet();
        $sheet = $spreadsheet->getActiveSheet();

        // this executes all registered modifiers and applys them in a huge
        // foreach row / col on each cell.
        $this->modify($sheet));

        // ...apply more generic rules like column formatting directly on the $sheet

        // we still can work with $in and still change the $result.

        $path = $this->convertResultToXslxVersion14($result);

        readfile($path);
        exit(0);
    }

    public function convertResultToXslxVersion14() { ... }
}

class FormatDateTimeModifier implements ExportModifierInterface {
    public function modify(Cell $source, object $target) {
        if (1$target instanceof Spreadsheet) return;

        $value = $source->getValue() ? $source->getValue()->format('c') : '')

        // We modify the target object via reference
        $target->styleCell($source->getX(), $source->getY());
        $target->setCellValue($soruce->getX(), $source->getY(), $value);

        // and return just the value for more common modifiers that need to be wrapped by the exporter
        // like markdown => text and excel needs to set the cell value with PhpOffice->setValue() or so.
        // that would happen in an overriden or extended ExcelExporter::modify().
        return $value;
    }
}
```

Its pretty easy to extend, you can integrate export settings (that could a combination of global settings, export settings and user settings).

Also every specific exporter has all the freedom to implement it's own modifiers that are very specific to the format or use common modifiers (`$this->modify()` needs to maybe wrap their behavior but that should be possible with).

It makes it testable, keeps it clean and readable. Dont be too abstract with the implementation. There is just one version of Excel 14, just one PDF version you want to implement and you want to be able to throw anything at it.

Thats just from the top of my mind right now, maybe your approach is similar to this :)

kevinpapst:
That looks great. I am just one developer, so I rarely get reviews... I appreciate every input that makes Kimai better and simpler to extend!
Do you want to propose a PR for that? It doesn't need to change all existing exporter obviously, just a start so we can talk about it in detail.
It only needs to be abstract enough to support any output format, not only spreadsheets. As I have plans for other exporters (eg. XML for import with finance applications).

kevinpapst:
I tried your changes and really like them!

What I wonder (maybe you know an answer):
I added an ODS renderer to export as copy&paste from your code and the result is still a plain table with wrong formatted data. The latter one probably comes from the format constants in `AbstractSpreadsheetRenderer`, which are currently targeting Excel... 

But when opening an Excel generated with your changes in OO, the fixed rows and filters are shown. Do you know if PHPSpreadsheet writer doesn't support these auto-filters in ODS?

kevinpapst:
@adrianrudnik did you have time to look into it?

adrianrudnik:
@kevinpapst Ah sorry, not this week, had some urgent tickets to relocate some workstations to home office places... hopefully monday if that is ok for you!

kevinpapst:
Sure, then just ping me whenever you found the time

adrianrudnik:
@kevinpapst Take a look at this, tried to make it a non-invasive options inside the abstract renderer. Just for the description column now, but I think it could be refactored into something to apply to all columns if `protected $columns` would be refactored into a proper configuration class.

src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -264,8 +267,31 @@ protected function getColumns(array $exportItems, TimesheetQuery $query, array $
...
+                // Apply wrap text if configured
+                if ($wrapText) {
+                    $cell->getStyle()->getAlignment()->setWrapText(true);
+                }
+
+                // Apply max width, only needs to be once per column
+                if (!$isColumnFormatted) {
+                    if (null !== $maxWidth) {
+                        $sheet->getColumnDimensionByColumn($column)
+                            ->setWidth(40);
```
kevinpapst: `->setWidth($maxWidth);`?

adrianrudnik: Yeah.... good idea.

kevinpapst:
> if protected $columns would be refactored into a proper configuration class

one step after another 😁 

I'll merge this one and then we can see what comes next.

PR #1528, opened 2020-03-06
[BUGFIX] Escape distinguished name in ldap search

sabbelasichon:
## Description
If you use special characters in the ldap "Distinguished Name" the search with a filter will break. 
We need to escape this before.

## Types of changes

## Checklist

### DIFF
--- a/src/Ldap/LdapManager.php
+++ b/src/Ldap/LdapManager.php
@@ -157,7 +157,7 @@ class LdapManager
 
         return $this->driver->search(
             $roleParameter['baseDn'],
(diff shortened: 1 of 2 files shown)


kevinpapst:
Thanks for the contribution @sabbelasichon ! 
I guess you already tried it in your live system? Looks good to me, I just have no test setup here, so I would like to get a last confirmation before merging.

sabbelasichon:
Thanks. Yes, after this change the login was successful again, before we had "Zend\Ldap\Exception\LdapException: 0x57 (Bad search filter)" exception.

PR #1536, opened 2020-03-08
Include pagination header for Timesheet list action

### LINKED ISSUE #2720: Return pagination related info in the api response
When I call /api/timesheets endpoint for fetching timesheet entries, by default Kimai 2 API returns 50 entries present on page 1(Default page size). But I have no means of knowing whether there are any more records and if there are, how many records/pages are there.

I would love to get a meta object along with timesheet entries so that I would know exactly how many pages/records are available.

Calling /api/timesheets GET endpoint with incrementing page param and collating responses until pagerfanta gives 404 exception because of the page out of range. But this is very very hacky, hence it would be good to know exact numbers of total pages.

**Screenshot**
Timesheets response object can look something like this :

`
{
   "data":[
      {
         "activity":0,
         "project":0,
         "user":0,
         "tags":[
            "string"
         ],
         "id":0,
         "begin":"2021-08-19T09:59:27.920Z",
         "end":"2021-08-19T09:59:27.920Z",
         "duration":0,
         "description":"string",
         "rate":0,
         "internalRate":0,
         "metaFields":[
            {
               "name":"string",
               "value":"string"
            }
         ]
      }
   ],
   "meta":{
      "currentPage":1,
      "totalPages":10,
      "itemsPerPage":50,
      "totalItems":490
   }
}
`

HonzaKopecky:
## Description
I noticed that when I am looping through the pages of Timesheets I never know what is the last page. To find that out I actually have to make a request that fails with 500. Which does not seem really the best way to me. How about Kimai would include a header saying whether the client should try to get next page? Once that header is missing, the client won't try to get any more data.

In dev environment, the 500 response actually says the reason for the failure is page out of index but in production, the message just says Internal Server Error so the client has no clue what caused the problem.

What do you think about this approach?

## Types of changes

## Checklist

### DIFF
--- a/src/API/BaseApiController.php
+++ b/src/API/BaseApiController.php
@@ -18,4 +18,5 @@ abstract class BaseApiController extends AbstractController
 {
     public const DATE_FORMAT = DateTimeType::HTML5_FORMAT;
     public const DATE_FORMAT_PHP = 'Y-m-d\TH:m:s';
+    public const NEXT_PAGE_HEADER = 'Kimai-Has-Next-Page';
 }
--- a/src/API/TimesheetController.php
+++ b/src/API/TimesheetController.php
@@ -226,6 +226,7 @@ class TimesheetController extends BaseApiController
 
         /** @var Pagerfanta $data */
         $data = $this->repository->getPagerfantaForQuery($query);
+        $hasNextPage = $data->hasNextPage();
         $data = (array) $data->getCurrentPageResults();
 
         $view = new View($data, 200);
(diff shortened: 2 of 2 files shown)


kevinpapst:
I would prefer a more generic / standard compliant way, which doesn't have to be added to every endpoint. I was thinking of embedding HAL or HATEOAS, but I don't have the time to look into it right now.

kevinpapst:
Next upgrade of NelmioApiDocBundle  will support the new OpenAPI specification, which includes support for linking other embedded resources/result sets ... maybe that can be used. Start reading here: 
https://github.com/nelmio/NelmioApiDocBundle/blob/master/UPGRADE-4.0.md

kevinpapst:
Thanks for your inspiration, sorry for keeping that open for sooooooo long.

Next version ships: https://www.kimai.org/documentation/2.0/api-pagination.html

PR #1542, opened 2020-03-11
fix timesheet permissions for non-admins

### LINKED ISSUE #1525: Accounting person cannot see Time Sheets
Situation:
Configured and running system with customers, projects, activities, some teams of customers with their team leaders. Everything is running fine.

Problem:
The time accounting is managed not by the team leader. The time accounting (evaluate and export the time sheets) managing is done by an separate accounting person. But this accounting person has no view privileges to see the "other" time sheets of all the other users in all other teams.

Workaround:
Put the accounting person into the role "ROLE_ADMIN". But the accounting person should not have privileges to change some customers, projects or activities!

other Workaround:
Set the accounting person as the team leader. But this is also not correct, the team leader is an another person!

Can anybody help me to solve this problem?
Is it possible to set a second team leader to a team?

kevinpapst:
## Description

This PR changes the way how timesheets of other users are read.

Considering a project or customer was not assigned to a team, but the current user is in a team, all timesheets related to globally available projects were excluded by default.

The result of such a query, eg. in the invoice or team-timesheets section was unpredictable and lead to very wrong results.

The new permission check is basically the same that is used for activities.

Resolves #1525 

## Types of changes

## Checklist

### DIFF
--- a/src/Repository/TimesheetRepository.php
+++ b/src/Repository/TimesheetRepository.php
@@ -527,24 +527,37 @@ class TimesheetRepository extends EntityRepository
             return;
         }
 
-        if (empty($teams)) {
-            return;
-        }
-
         // make sure that admins see all timesheet records
         if (null !== $user && ($user->isSuperAdmin() || $user->isAdmin())) {
             return;
+        }
+
+        if (null !== $user) {
+            $teams = array_merge($teams, $user->getTeams()->toArray());
         }
 
         $qb
             ->leftJoin('p.teams', 'teams')
             ->leftJoin('c.teams', 'c_teams');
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
ping @AlexanderOsswald can you test this change in your installation and provide feedback ?

kevinpapst:
What exactly did you do for testing? Upgrade to 1.8 and then replace the one file + cache rebuild?

kevinpapst:
I was able to reproduce your problem and this PR fixed the issue on my end. 
If it doesn't for you, then you have a different problem 😢 
Anyways, this fix is necessary ...

PR #1568, opened 2020-03-17
switch to kimai specific user-bundle fork

kevinpapst:
## Description

This PR replaces the FOSUserBundle with a Kimai fork, see https://github.com/kimai/user-bundle

This step was necessary as FOSUserBundle is neither recommended nor can be claimed a reliable and active project (even though some commits happen here and there). It wasn't tagged in more than 1,5 years and the Kimai composer package was pointing to master.

It ships a lot of unnecessary code (for Kimai), which can now be eliminated and one-by-one integrated into Kimai directly with a lot of new possibilities.

Finally the log gets cleaned up, as several outstanding deprecations can be fixed one after another (some already are).

## Types of changes

## Checklist


jkabat:
@kevinpapst you may want to check out this [repo](https://github.com/msgphp/user-bundle).

I've got rid of FOS long time ago and used above replacement on several projects.

kevinpapst:
Thanks for the suggestion @jkabat 
Thats the only repository I really considered, while I was searching for replacements. 
Good to know that you already used it successful. 

For now I wanted to have a solution that works fast with the least possible work 😁 but when I will do a real refactoring on that part I will try it.

Issue #1560, opened 2020-03-16
PWA Support to allow Desktop installation


leopf:
Add a service worker and manifest.json to the project, allowing desktop installation and asset caching.

I suggest to add the "workbox-webpack-plugin" to the webpack.config. It automatically generates all caching rules required for the application.

For the manifest.json generation (including icon scaling etc.) I suggest the webpack plugin "webpack-pwa-manifest". 

I have workes with both plugins before. They are very easy to use and don't obstruct the deployment process. I would be happy to contribute, solving this request!

kevinpapst:
That would be awesome! I would love to see that. Can you share some infos what we can do with it? I have only theoretical knowledge about PWAs from a few years ago.

If you start working on it, I can contribute my knowledge about Kimai and its "frontend stack" ... but I guess it is easy to understand for real frontend/JS developers. Mainly basic compilation tasks and a self-baked plugin system.

leopf:
Great to hear!

There are a lot of things you could do!

1. Web-Notifications without having the Website opened.
2. Use the share API of mobile phones. That way you could start Time-Tracking by sharing a text to Kimai (there are probably better use cases).
3. Theoratically allow complete offline use. That would require the frontend to be decoupled from the backend. With local storage or IndexedDB you could f.a. store not yet saved time entries and commit them once the client is online again.
4. Obv. desktop/homescreen installation.

There are a lot more things beeing added and things that I don't know about, but it seems like a great fit for Kimai. 

For now I would start with caching css, woff2, js and ico files. The service worker can then be extended to allow for notifications, complete offline use etc.

kevinpapst:
Might be a lot of necessary changes, but they are worth the effort. 
I wanted to decouple front-end / backend for a longer time, so this could be a very good reason to start it.

adrianrudnik:
PWA support is nice, but it has hard limits and a very high technical debt once you come to topics like offline or disconnected usage scenarios. It mostly works on Google products and only makes sense if you want to offer a shortcut to install a website as a website-app. I dont think that Safari and Firefox are feature complete as Chrome is in this regard.

That limits the feature set that makes sense for PWA to propose an app install. As the app itself is always-online, it is also always-cached if the headers are set correctly and there is no scenario where it should be unusable on desktop or mobile right now.

[...]

PR #1571, opened 2020-03-19
Installable PWA

adrianrudnik:
## Description

This is an example on how to get Kimai on the way as an installable PWA app for desktop and mobile devices as discussed in #1560 

## Types of changes
- [x] Added header directory for sw.js to apache .htaccess. What is the correct Dockerfile right now to execute an `a2enmod headers`?
- [x] Added controller to handle localized manifest.json generations.
- [x] Added manifest.json exclusion to firewall.
- [x] Added registration of service worker to app.js

## Checklist

## Remarks

I attached a [pwa report](https://github.com/kevinpapst/kimai2/files/4355367/pwa.pdf) that indicates that these changes should satisfy any requirements to be installable. There are other points that could be improved but are not in scope of this.


webpack.config.js
```diff
@@ -1,11 +1,12 @@
 // Hint: if something doesn't work as expected: yarn upgrade --latest
 
 var Encore = require('@symfony/webpack-encore');
+var WorkboxPlugin = require('workbox-webpack-plugin');
 var webpack = require('webpack');
 
 Encore
     .setOutputPath('public/build/')
-    .setPublicPath('build/')
+    .setPublicPath('/build')
```
kevinpapst: Besides all other changes, which I will have to look at when there is more time - this one is wrong.
The setting is used like this to allow usage in subdirectory scenarios.
The webpack/encore warning is highly annoying and misleading in this case.

kevinpapst:
Thanks for this quick prototype @adrianrudnik - really appreciated!

I never worked with either PWA or service workers. 
As you said in #1560 adding this is a technical debt and for now it doesn't add a "real benefit", does it? Besides being a great example and starting point.

I just want to start a discussion about pros and cons, which I cannot overlook right now.

webpack.config.js
adrianrudnik: ah, wondered why this was changed. Just a draft PR as an idea how to get there :)

muety:
I'd love to have Kimai as a PWA!

samuelT2:
Having Kimai as a PWA would be awesome. I've already added the app to my homescreen on my iPhone.

> ... it doesn't add a "real benefit", does it? ...

I think it would, eg. I could run the app fullscreen (manifest.json: "display":"standalone") and cached. It would more or less behave like a native app. I just made two projects of mine PWA ready, it took some time to get into service workers, but the result is, in my opinion, really cool.

kevinpapst:
I need support with this topic, as I do not have any experience. If someone wants to rebase this on #2902 and send a new PR which I can test, that would be awesome 🚀

PR #1598, opened 2020-03-29
add event for updated timesheet entites

lduer:
## Description

I've added the new `TImesheetUpdateEvent` which is dispatched each time a modification is done to a `Timesheet` entity. I use this for some plugins and think it is useful for all.

## Types of changes

## Checklist


kevinpapst:
I like the idea, but I am thinking if that could be done differently. Currently this needs to be attached to every new place where a timesheet is somehow manipulated. 

Doctrine for example already triggers events that could be used, like this one: 
https://github.com/kevinpapst/kimai2/blob/master/src/Doctrine/TimesheetSubscriber.php#L49

Do you use all event types already?

src/Event/TimesheetUpdateEvent.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace App\Event;
+
+use App\Entity\Timesheet;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * This event can be used to call functions after timesheet creations/updates
+ */
+final class TimesheetUpdateEvent extends Event
```
kevinpapst, line 18: Instead of one event I would like to have one class for every use-case (TimesheetUpdateEvent, TimesheetDeleteEvent, TimesheetMultiUpdateEvent, TimesheetMultiDeletEvent...), with two base classes AbstractTimesheetEvent and AbstractTimesheetMultiEvent.
Then the method signatures could be made explicit: either for one entity or for a collection.

lduer:
> Do you use all event types already?

Currently, I'm listening to all, except the duplicate event type.

> Doctrine for example already triggers events [...]

I know the doctrine events and also tried to use this events, but I for me it was simpler with new listeners. But after retrying with them, I think I could get it to work there.

Sometimes, I need different views and hints from other people, to get my work done properly ;)

After all, I think that much code is added to the repository, which is not used oftern and the same could be achieved with the doctrine listeners (currently, at least for me). Additionally, IMO the updates of other entities (Customer, Project, Activity, ....) need this events too, if we want to be consistent through the project.

Do you want to follow the ideas of this PR [`Instead of one event I would like to have one class for every use-case`]? I'd like to contribute if necessary. Or can we close it?

kevinpapst:
Yes I definitely want to continue with this feature.
Sorry about taking over your PR with my changes 😁 

So I pushed some changes regarding the explicit event classes vs event names and added tests.
Moved the event dispatching to TimesheetService, so we have one place where everything is handled. 
And there is now always an event dispatched before and after the action, so plugins could for example change values before the timesheet is updated or log something before its deleted.

Two events were removed and two stills need to be changed (the multi update/delete).
What do you think?

lduer:
No problem, I like your changes :) My existing code can be updated easily with this changes. 

My Idea of the multiupdate is something like this: (e.g. here: [src/Controller/TimesheetAbstractController.php#L346](https://github.com/kevinpapst/kimai2/blob/master/src/Controller/TimesheetAbstractController.php#L346)) 
```
$this->dispatcher->dispatch(new TimesheetMulitUpdatePreEvent($dto->getEntities()));
$this->repository->saveMultiple($dto->getEntities());
$this->dispatcher->dispatch(new TimesheetMulitUpdatePostEvent($dto->getEntities()));           
```
The `TimesheetMulitUpdatePreEvent` extends the `TimesheetUpdatePreEvent` and saves the entities as array instead.

kevinpapst:
I would like to do it like that:
- create two proxy methods in timesheet service for the multiX calls
- add the events around the repository calls in that service  (then the only location where these events are triggered is the service)
- add two new event classes + an abstract base class

The multi events should extend a new AbstractTimesheetMultiEvent. Otherwise the construct would be incompatible and `getTimesheet()` would be leaked to these events, which sounds wrong as they can't return a proper value.

kevinpapst:
First I was thinking if we could call the update/delete for each item separatly and get rid of the "Multi" methods in the repository. 
If would be easier, but we would loose the atomic operation/transaction, which in a multi update might result in undefined state...

kevinpapst:
@lduer I want to release 1.9 soon. Do you want to add the missing methods/events or do you expect me to do it?

lduer:
Thanks for adding and merging and releasing in 1.9. 
Sry, I was not able to contribute because of lack of time ;)

kevinpapst:
I do know that time problem 😁

PR #1607, opened 2020-04-02
Esperanto translation

KuboF:
## Description
Esperanto translation of Kimai. Translated by Sofia Zaretskaia, reviewed by Michal Matúšov (@KuboF) for E@I.

## Types of changes

## Checklist


kevinpapst:
@KuboF I have pushed some changes, to support Esperanto (where possible) in the frontend tools. The `fullcalendar` doesn't know the `eo` locale, so it falls back to `en`. Same goes for the selectpicker `select2`.

Select2 is very simple to translate, just a few phrases (as example the [english file](https://github.com/select2/select2/blob/develop/src/js/select2/i18n/en.js)).
Just if you feel inspired to translate it as well...

Anyway: Can you please do a test-run locally with the changes?

KuboF:
Thanks for rebuild, @kevinpapst!

My local test works well!

I have found some translation problems - now fixed with the 2 new commits.

Esperanto translations / locales for dependencies:

- 'select2': https://github.com/select2/select2/pull/5823 (open)
- 'fullcalendar': translation prepared, I will PR after merge of my previous PR (I missed my branches a bit)

kevinpapst:
Kimai is still using Fullcalendar v3 and there seems to be a version v5 coming soon. 
Not sure which version you translated. But its great to see that you are actively working on raising awareness for Esperanto 👍 

Anyways - let me know when you are done with the PR, then I will merge it. 
Changes can be done any time later, I want you to take the time needed for final checks.

translations/export.eo.xlf
```diff
@@ -0,0 +1,16 @@
+
```
kevinpapst, line 1: please remove the new line, it is invalid XML otherwise and the build fails

KuboF: Fixed. Thanks for info!

PR #1631, opened 2020-04-13
Add color to tags

jojoemade:
## Description

## Types of changes

## Checklist
- [] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))


src/Form/TagEditForm.php
```diff
@@ -10,13 +10,16 @@
...
 use App\Entity\Tag;
+use App\Form\Type\ColorPickerType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 class TagEditForm extends AbstractType
 {
+    use EntityFormTrait;
```
kevinpapst, line 21: I think this trait is not required, can be removed!?

src/Migrations/Version20200413133226.php
```diff
@@ -0,0 +1,38 @@
...
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
```
kevinpapst, line 15: please use [this base class](https://github.com/kevinpapst/kimai2/blob/master/src/Doctrine/AbstractMigration.php):

```php
use App\Doctrine\AbstractMigration; 
```

templates/macros/widgets.html.twig
```diff
@@ -369,7 +369,7 @@
 {% macro tag_list(taglist) %}
     {% import _self as macro %}
     {% for tag in taglist %}
-        {{ macro.badge(tag.name , 'green') }}
+        {{ macro.badge(tag.name , tag.color ?? 'red') }}
```
kevinpapst, line 372: 'green' is not 'red' 😁 

And we should not rely on the pre-defined browser colors, but use the color codes from the css. 
bg-green: `#00a65a`
bg-red: `#dd4b39`

So: `tag.color ?? '#00a65a'`

templates/macros/widgets.html.twig
```diff
@@ -160,7 +160,7 @@
 
 {% macro badge(title, color) %}
     {# black, green, blue, yellow #}
-    <span class="badge bg-{{ color|default('red') }}">{{ title|trans }}</span>
+    <span class="badge" style="background-color:{{ color }}">{{ title|trans }}</span>
```
kevinpapst, line 163: `{{ color|default('#dd4b39') }}`

templates/macros/widgets.html.twig
jojoemade: Implemented the suggestion below:
`{{ color|default('##00a65a') }}`
So this isn't needed anymore, right?

kevinpapst: Please push your changes, so I can see them

jojoemade: Done ;)

kevinpapst: With the latest push you still change the default color from red to green. We could discuss whether red is the correct default color for a badge ... but then we have to check all usages of the macro (not only in core but also in plugins).

Badges are not only used in combination with tags. Please change the default color for tags to `#dd4b39` and the default for tags as you had it before to `tags.color|default('#00a65a')`

jojoemade: Now I got it, sorry. ;)

tests/Entity/TagTest.php
```diff
@@ -32,6 +33,9 @@ public function testSetterAndGetter()
         $this->assertInstanceOf(Tag::class, $sut->setName('foo'));
         $this->assertEquals('foo', $sut->getName());
         $this->assertEquals('foo', (string) $sut);
+
+        $this->assertInstanceOf(Tag::class, $sut->setColor('#fffccc'));
+        $this->assertEquals('#fffccc', $sut->getColor());
```
jojoemade, line 38: Thanks for the addition!

Wouldn't it make more sense to test the trait just once instead of testing every place where it's used?

See https://phpunit.readthedocs.io/en/9.1/test-doubles.html#mocking-traits-and-abstract-classes

kevinpapst: I am not really testing that the method works as expected, but I rather want to make sure that the methods exists on the object.

kevinpapst:
Thanks @jojoemade really appreciate your contribution 👍

PR #1639, opened 2020-04-15
fix duration calculation in admin timesheet view

### LINKED ISSUE #1638: "To" is not correctly calculated from timesheets view
**Describe the bug**
When adding a new entry from the timesheets view and just filling out the duration, "To" is not calculated but left empty. The entry is then generated as an active timesheet with the duration being ignored.
Creation from "My times" view works fine: Filling out a duration leads to "To" being calculated.

**To Reproduce**
1. Go to 'Timesheets'
2. Click on 'Add'
3. Fill out 'From', 'Duration' and all mandatory fields.
4. Click on 'Save'

**Screenshots**
image
image

kevinpapst:
## Description

Fixes #1638 

ping @jojoemade can you please test?

## Types of changes

## Checklist

### DIFF
--- a/templates/timesheet/layout-edit.html.twig
+++ b/templates/timesheet/layout-edit.html.twig
@@ -71,16 +71,17 @@
         {% endblock %}
         {% block form_after %}
             {% if form.begin is defined and form.end is defined and form.duration is defined %}
+                {% set blockPrefix = form.vars.id %}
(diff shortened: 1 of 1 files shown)


kevinpapst:
Testing in normal user and admin view: everything works as expected now

jojoemade:
Works for me as well. 😁

Issue #1605, opened 2020-04-01
Time based permissions


SpoorthiPalakshaiah:
Can Kimai provide time based permissions for users.
Like we can edit timesheets only for one day from the day timesheet was created.
or something like add timesheets only for the dates in the same week and not previous weeks.

kevinpapst:
This would be a check like the "allow future times" check that is already available. 
Makes totally sense, yes.
No its not existing yet, but a good idea or the future

adrianrudnik:
If this is every considered for implementation, please ensure this works hand in hand with already exported and invoiced items and offers hard rejections for both scenarios, not just a warning. Maybe even an "allow edits until exports/invoicings" may be an option.

kevinpapst:
"allow edits until exports" already exist, it is the role permission `edit_exported_timesheet`.

Issue #245, opened 2018-07-28
Feature request - disable time recording for past months


eporsche:
**Is your feature request related to a problem? Please describe.**
For accounting reasons it should be possible to "close" a past month so that no further changes are possible. A past month can only be opened again by an administrator.  
This gives assurance to an employer that times in the time tracking tool are consistent to what has been invoiced to customers or internally to cost centers  

**Describe the solution you'd like**
* past month time booking can be automatically closed with a configurable amount of days, e.g. 5 days
* past month can be manually closed o reopened by users with the correct rights

kevinpapst:
Collecting ideas/todos for this request:

- There should be a configuration entry for a "don't allow to edit before date". 
  - Use \DateTime so we can use not only a date, but also [relative formats](https://secure.php.net/manual/en/datetime.formats.relative.php) (like "first day of this month")
  - If set, it is used to prevent adding entries before that date in the user timesheet. 
  - If set, it is used to prevent changing entries before that date in the user timesheet. 
- The admin view for timesheet will only display a warning, but still allow to change existing or add new entries in that "past period".

Regarding invoices: that will be handled in a different request, as I plan to assign timesheet entries to invoices after generation. These entries will not be editable afterwards any longer.

HeinzWuert:
I don't know if a fixed date is the best way, may be one of two!
The second way would be the user finishes his invoice for the month. After he has finished (with warning windows, that changes are not further possible) all bookings can be made by the company administration.
An automatic completion should be possible too.
BUT: For the Kimai administration it must be possible to make changes!

cbellersen:
We would greatly appreciate this feature as well. I would like contribute a suggestion, based on how this is implemented in our current time tracking software: 

- After a user is done entering his time reports, he can pick a date, until he wishes to “release” his reports. He can no longer add or edit reports before this date.
<img width="764" alt="release" src="https://user-images.githubusercontent.com/3328244/54044952-8d3b7d80-41d0-11e9-8bae-7b7ec12f6c2d.png">


- An admin user / team lead can see all the dates, until other users have released their time reports. He can also re-open the reports, if regular users wishes to change something or add a report prior to the release date.  
<img width="673" alt="admin" src="https://user-images.githubusercontent.com/3328244/54044937-81e85200-41d0-11e9-988b-205af7dc1301.png">

Our currenct software also offers a “finalize” feature, which the admin user can use on all released reports, but I don’t think that is essential. 



[...]

PR #1644, opened 2020-04-16
Timesheet lockdown with grace period

HonzaKopecky:
## Description
Introducing feature that enables admin to setup a lockdown period after which timesheet entries are no longer editable/creatable. Using PHP relative DateTime syntax a lockdown period is defined using `lockdown_period_start` (e.g. `first day of last month 00:00:00`) and `lockdown_period_end` (e.g. `last day of last month 23:59:59`) settings.  All entries before the start are not editable. All entries inside of the period are only editable until the 'now' date is inside of the period.

Feature also introduces setting `lockdown_grace_period`. This is also PHP relative DateTime setting. It allows users to modify/create timesheet entries even after the lockdown period has ended. It doesn't allow users to edit/create entries before the lockdown period start.

Feature also introduces two new permissions. `lockdown_complete_override_timesheet` is by default assigned to admin users and it allows them to edit/create any entries no matter the lockdown/grace period. `lockdown_grace_override_timesheet` is by default assigned to admin users and allows users to edit/create entries of the last lockdown period (between start and end) even though the lockdown and grace period passed.

## Types of changes

## Checklist

## Possible todos/future work
- The fact that all settings fields are required forced me to do this hacky way with space. When the setting contains just space, the getters will return null.
- The settings should probably be validated at the time of input. Can that be achieved by creating another Type (e.g. RelativeDateTimeType)? 
- Also, `lockdown_grace_period` is dependent on `lockdown_period_end`. How would one validate this at the time of input?
- Timezones. It seems to work with timesheet entries that have other timezones but it's not really clear what timezone are the settings (start, end, grace) in thus it's a bit confusing.


kevinpapst:
Hi! Thanks for the PR 👍  Sounds awesome.
I have a broken internet connection @ home and cannot test the next days.
Will get back ASAP

kevinpapst:
The command `composer kimai:pre-commit` is meant to run as pre commit hook and only checks the stuff that works rather fast. 
Integration tests take too long for a pre commit hook.

So you need to run `composer kimai:code-check` to check everything locally.

HeinzWuert:
Reads interesting.
Will it be possible to
- create a permanent rule which
- on a specific day of the current month
- complete the entire previous month (marked as exported)?
We always have the final invoice for the previous month on the 10th of the current month. From then on, changes to the bookings are no longer permitted.
That would be great as an automatism.

kevinpapst:
fixes #245
fixes #1605

hmr-it-jr:
@HonzaKopecky will you finish the development on this PR?

kevinpapst:
@HonzaKopecky that took quite a while, but I am happy with the PR now. I changed a couple of things:
- extracted timesheet validators into single classes (not directly related to this change, but it was a good moment to do it)
- added tests for validators
- renamed the new permissions
- added a constraint to validate if the configuration are valid date time strings 
- validate that end is not before start
- added documentation (see https://github.com/kimai/www.kimai.org/pull/67)

Do you want to give it a try?

src/Validator/Constraints/TimesheetLockdownValidator.php
```diff
@@ -0,0 +1,130 @@
...
+
+        // lockdown never takes effect for users with special permission
+        if ($this->auth->isGranted('lockdown_override_timesheet')) {
+            return;
+        }
+
+        if (!empty($constraint->now)) {
+            if ($constraint->now instanceof \DateTime) {
+                $now = $constraint->now;
+            } else {
```
kevinpapst, line 98: elseif (is_string())

src/Validator/Constraints/TimesheetOverlapping.php
```diff
@@ -0,0 +1,26 @@
...
+
+final class TimesheetOverlapping extends TimesheetConstraint
+{
+    public const RECORD_OVERLAPPING = 'kimai-timesheet-overlapping-01';
+
+    protected static $errorNames = [
+        self::RECORD_OVERLAPPING => 'You already have an entry for this time.',
+    ];
+
+    public $message = 'Overlapping entries found.';
```
kevinpapst, line 20: unify message

kevinpapst:
thanks @HonzaKopecky for the idea and input 👍

PR #1724, opened 2020-05-20
added hebrew translations

### LINKED ISSUE #1723: Translation to Hebrew
I have translated kimai to Hebrew (if you're interested I can send you the translation files).
Everything works fine except for the name of days in calendars which looks like Chinese...
I've added the following lines to assets\app.js:
    require('select2/dist/js/i18n/he');
    require('moment/locale/he');
And this line to assets\calendar.js:
    require('fullcalendar/dist/locale/he');
However I don't see any change.

Also, how can I translate the login page?
I've created the translation file under: 
kimai2\vendor\kevinpapst\adminlte-bundle\Resources\translations\AdminLTEBundle.he.xliff
But the page is still in English.

kevinpapst:
## Description

thanks @eransurg

Can be tested at https://demo-branch.kimai.org/he/login with:
- Username: `susan_super`
- Password: `kitten`

Fixes #1723 

## Types of changes

## Checklist


translations/about.he.xlf
```diff
@@ -0,0 +1,46 @@
...
+        <source>tab.thanks</source>
+        <target>תודות</target>
+      </trans-unit>
+      <trans-unit id="mz4ieCN" resname="donate">
+        <source>donate</source>
+        <target>תרומה</target>
+      </trans-unit>
+      <trans-unit id="DwbowBh" resname="published_under">
+        <source>published_under</source>
+        <target>%kimai% is published under</target>
```
kevinpapst, line 34: still english

eransurg, line 34:
```suggestion
        <target>%kimai% מופץ תחת</target>
```

eransurg: Fixed

kevinpapst: Thanks, there are two more in the same file

translations/about.he.xlf
```diff
@@ -0,0 +1,46 @@
...
+      </trans-unit>
+      <trans-unit id="A0nWoXZ" resname="special_thanks">
+        <source>special_thanks</source>
+        <target>Special thanks go to …</target>
```
eransurg, line 38:
```suggestion
        <target>תודה מיוחדת הולכת אל …</target>
```

translations/about.he.xlf
```diff
@@ -0,0 +1,46 @@
...
+      </trans-unit>
+      <trans-unit id="L7cff3Q" resname="library_authors">
+        <source>library_authors</source>
+        <target>… the authors of the following software-libraries, Kimai wouldn't be possible without you 👍</target>
```
eransurg, line 42:
```suggestion
        <target>… המפתחים של רכיבי התכנה האלה, Kimai לא הייתה אפשרית בלעדיכם 👍</target>
```

translations/about.he.xlf
eransurg: Done

kevinpapst:
ping @eransurg 

Sorry that it took longer with the demo, but it was still needed for another test. No your hebrew translations can be tested at https://demo-branch.kimai.org/ro/login with:
- Username: `susan_super`
- Password: `kitten`

eransurg:
1. When trying to register a new user from login page - the value for "Email" has \. I think it's because the value has two single quotes and not one double quotes.
2. The names of week days in calendar are ok but the week should start with Sunday (Friday+Saturday are weekend)

kevinpapst:
1. Do you have any idea how to fix that? Could that be written in another way? Using html `&quot;` instead of `"` leads to the same result.

2. Isn't that the case already? I don't have much experience with other locales, but that looks like right to left with today (24 being a Sunday) the first day of the week? Or am I reading that wrong?
Bildschirmfoto 2020-05-24 um 11 33 57

kevinpapst:
Ah, I guess you mean the datepicker, right?
I have to check if/how that is set. I expected it to be fetched from the momentjs library, but seems it isn't...

eransurg:
Exactly, what I meant was the datepicker...and now I see that it's fixed.
About the translation value, you can put there:
דואר אלקטרוני
That's simply the complete word and not the shortcut for email.

kevinpapst:
Okay, found an easy solution to the datepicker problem.
It was always initialized with "Monday" being the first day of the week (my bad).

As I expect a lot of new issues coming up for the next release, telling me that Sunday is not the first day of the week (from users that use english, because their native language is not available) it might be necessary to make it configurable - see #1732

kevinpapst:
Could you please propose the change here: https://github.com/kimai/user-bundle/blob/master/Resources/translations/FOSUserBundle.he.yml#L21

I tried to fix it, but failed due to the usage of RTL that I never learned. When I try to type, my editor is getting "nervous" 😁

eransurg:
Yeah....Hebrew is a pain in the a** 😂
I've proposed the change

kevinpapst:
Thanks, demo is updated

eransurg:
Everything looks good now
Thanks for the great work!
How can I apply the datepicker fix to my environment?
Or should I wait for the next version to upgrade?

kevinpapst:
Fantastic contribution, very much appreciated @eransurg 😄 

You could update to the master branch if you know how that works.
Otherwise you have to wait for the next release to get your changes.

Issue #1010, opened 2019-08-05
Project view


nicolas-75:
**Is your feature request related to a problem? Please describe.**
The "my times" screen displays all the records. Is it possible to display only projects? Thus, to have the detail of the time recorded by the project, it is possible to click on the name of the project or to filter.

Indeed, I rarely need to have the details of the time spent and only the full time or the full time per week interests me.

I can not get this information in the demo. There are statistical reports that can be created?

**Describe the solution you'd like**
Display one line per project and four columns: time spent today; time spent in the week; time spent from the beginning; expected time
To see the details of the time spent, click on the name of the project.

**Additional context**
Freelance working on several projects in parallel and can last several months with long interruptions

kevinpapst:
You may get some of these informations through various way, eg. 
- create an invoice for the project and select the time range
- check the budget view of the project to see all time recorded.
- create an export and do the calculation in Excel

But there is no way to get these informations in Kimai via a mouse-click right now.

nicolas-75:
I made a model of the view that interests me so if someone wants to contribute I can make a transfer of 200 €. There is a bounty pot for changes requests?

![project-view](https://user-images.githubusercontent.com/3610660/62548933-e420ae00-b867-11e9-98e3-9abfaac44f08.png)

kevinpapst:
I have setup a Bountysource team for Kimai 2:
https://www.bountysource.com/teams/kimai2

And here is your issue:
https://www.bountysource.com/issues/78268231-project-view

nicolas-75:
Great ! I thank you for this step. I published a bonus on this platform. Thank you

kevinpapst:
Awesome, thanks for posting the first bounty!

Can you explain a bit, so we can gather a list of all requirements:
- the expected duration could be the time budget
- the expected delivery has no matching field yet, so a new "end date" would be required (which makes totally sense for projects)
- If there is an end date, a start date would be good to have as well, so it would be possible to show a progressbar?!
- What is up with the eye icon?
- Calculating all these values will be quite "expensive" for the database, so it might become rather slow with a growing database of timesheet records. How did you imagine is the list of projects populated? List all projects the current user worked at in the selected time-range?

nicolas-75:
The eye icon is a way to hide projects that are no longer to display. But that implies being able to add them again by another means so it is not the most urgent.

Indeed, some projects can be suspended for several months so it is not necessary that they are present in the list.

[...]

PR #1738, opened 2020-05-25
Project view

willianveiga:
## Description
I've created this PR to start fixing the bug #1010.

## Types of changes

## Checklist


src/EventSubscriber/MenuSubscriber.php
```diff
@@ -81,6 +81,10 @@ private function configureMainMenu(SidebarMenuEvent $menu)
                 new MenuItemModel('export', 'menu.export', 'export', [], $this->getIcon('export'))
             );
         }
+
+        $menu->addItem(
```
willianveiga, line 85: I'm not sure which permissions to require here. Do I need to define a new or use an existing one?

src/Controller/ProjectViewController.php
```diff
@@ -0,0 +1,68 @@
...
+use App\Repository\Query\ProjectQuery;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route(path="/project_view")
+ */
+final class ProjectViewController extends AbstractController
```
willianveiga, line 24: I'm not sure which permissions to require here. Do I need to define a new or use an existing one?

kevinpapst: I'd say project_budget or project_details can be reused

src/EventSubscriber/MenuSubscriber.php
kevinpapst: why adding a new menu here? I would move it to the toolbar (action menu) of the project listing instead

src/Repository/ProjectRepository.php
```diff
@@ -460,6 +460,62 @@ public function deleteProject(Project $delete, ?Project $replace = null)
...
+                FROM App\Entity\Timesheet AS t1
+                WHERE t1.project = t.project
+                    AND DATE(t1.begin) = '$today'
+            ) AS today
+        ";
+    }
+
+    private function weekProjectViewSubQuery(): string
+    {
+        $today = \DateTime::createFromFormat('U', time());
```
kevinpapst, line 499: please write dql queries with objects and parameters

templates/project_view/index.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% block page_title %}{{ 'project_view.title'|trans }}{% endblock %}
+
+{% block main_before %}
+    {{ tables.data_table_column_modal(tableName, columns) }}
+{% endblock %}
+
+{% block main %}
+
+    {% if entries is null %}
+        {{ widgets.nothing_found('kimai.activityUpdate') }}
```
kevinpapst, line 37: the event is wrong. if you only have read logic, then this can be completely removed (same below in the table header)

src/Repository/ProjectRepository.php
```diff
@@ -460,6 +460,62 @@ public function deleteProject(Project $delete, ?Project $replace = null)
         }
     }
 
+    public function getProjectView(): array
```
kevinpapst, line 463: pass in the start date, so the query can be re-used

src/EventSubscriber/MenuSubscriber.php
willianveiga: image
Considering the image above, should I add it to the menu number 1 or 2?

kevinpapst: 1 - see [here](https://github.com/kevinpapst/kimai2/blob/master/templates/project/actions.html.twig#L9)

kevinpapst:
I am sorry for not coming back to this topic earlier, but I had to add some features first before we can go on with his topic. I guess you want to get the bounty, which is totally fine for me. But it should be in a way that fits into my overall vision.

For the next release 1.10 I added a reporting screen, which can be extended by further views.
Basically this is list of links :
https://github.com/kevinpapst/kimai2/blob/master/templates/reporting/actions.html.twig#L5

The list of links can be extended by plugins and you can add default reports in this twig extension:
https://github.com/kevinpapst/kimai2/blob/master/src/Twig/ReportingExtension.php#L52

From there on its up to you what to display, all you have to do is to extend the reporting layout:
https://github.com/kevinpapst/kimai2/blob/master/templates/reporting/layout.html.twig

See for example this report:
https://github.com/kevinpapst/kimai2/blob/master/templates/reporting/monthly_user_list.html.twig

You can add your screen to the default reporting controller:
https://github.com/kevinpapst/kimai2/blob/master/src/Controller/ReportingController.php

Or if you need more dependencies add a new directory src/Controller/Reporting/ and more your controller there.

Last point I'd like to address: these new report queries should go into a new repository. I will refactor the existing respotiroes in the future and move all statistic/report code there as well. Maybe you can create that new class while doing the last changes?

kevinpapst:
Thanks for working on it!

Now 456 changed files are shown, which is not possible to review (likely all changes on master since your initial commit in this branch). So I need you to either rebase your changes or squash them.

kevinpapst (approved):
That was the longest outstanding PR I have ever worked on. Thank you so much for the initial idea and implementation @willianveiga 🎉

PR #1766, opened 2020-06-04
#1765 force 2 digits after the decimal point

fredyb:
## Description
A clear and concise description of what this pull request adds or changes.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I verified that my code applies to the guidelines (`composer kimai:code-check`)
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [ ] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

Issue #1765, opened 2020-06-04
Invoice: Decimal fraction is different depends on the value


fredyb:
When I generate an invoice (docx), the decimal fractioin are different, depending on the fraction value (e.g. 10 / 5.**5** / 35.**95** / ..)

Proposed solution (and tested):
add in: src/Twig/Extensions.php
 
    private function initLocale()
    {
        $locale = $this->localeSettings->getLocale();

        if ($this->locale === $locale) {
            return;
        }

        $this->locale = $locale;
        $this->numberFormatter = new \NumberFormatter($locale, NumberFormatter::DECIMAL);
        $this->moneyFormatter = new \NumberFormatter($locale, NumberFormatter::CURRENCY);

        _**// add formatter FRACTION_DIGITS #.## 
        $this->numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
        $this->moneyFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);**_
    }

For me this change works fin.

Thanks for the great software!!

kevinpapst:
You mean the length is not forced, but can differ between zero, one and two digits?
Can you submit a PR?

fredyb:
yes exactly

fredyb:
What is a PR?

kevinpapst:
A PR is a "pull request" via GitHub. Basically you do the change in a branch of your fork and then sent this change as PR to this repo. Then I can check and accept it.

If all of that is new to you and/or you are not a developer: don't bother.

PR #1788, opened 2020-06-22
#1765 force 2 digits after the decimal point (replace PR #1766)

fredyb:
## Description

## Types of changes

## Checklist

### DIFF
--- a/src/Export/Base/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Base/AbstractSpreadsheetRenderer.php
@@ -656,7 +656,7 @@ abstract class AbstractSpreadsheetRenderer
     abstract protected function saveSpreadsheet(Spreadsheet $spreadsheet): string;
 
     /**
-     * Get the value of durationDecFormat
+     * Get the value of durationFormat
      *
      * @return bool
      */
@@ -666,9 +666,9 @@ abstract class AbstractSpreadsheetRenderer
     }
 
     /**
-     * Set the value of durationDecFormat
+     * Set the value of durationFormat
      *
-     * @param bool $durationDecFormat
+     * @param bool $durationFormat
      *
      * @return self
      */
--- a/src/Export/Base/AbstractSpreadsheetRenderer.php
+++ b/src/Export/Base/AbstractSpreadsheetRenderer.php
@@ -41,6 +41,7 @@ abstract class AbstractSpreadsheetRenderer
     public const DATETIME_FORMAT = 'yyyy-mm-dd hh:mm';
     public const TIME_FORMAT = 'hh:mm';
     public const DURATION_FORMAT = '[hh]:mm';
+    public const DURATION_FORMAT_DEC = '#0.00';
     public const RATE_FORMAT_DEFAULT = '#.##0,00 [$%1$s];-#.##0,00 [$%1$s]';
     public const RATE_FORMAT_LEFT = '_("%1$s"* #,##0.00_);_("%1$s"* \(#,##0.00\);_("%1$s"* "-"??_);_(@_)';
     public const RATE_FORMAT = self::RATE_FORMAT_LEFT;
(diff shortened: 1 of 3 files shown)


src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -61,6 +62,10 @@ abstract class AbstractSpreadsheetRenderer
      * @var AuthorizationCheckerInterface
      */
     protected $voter;
+    /**
+     * @var bool
+     */
+    protected $durationFormat;
```
kevinpapst, line 68: please initialize this with false (for BC reasons)

src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -632,4 +654,28 @@ protected function getFileResponse($file, $filename): Response
      * @throws \Exception
      */
     abstract protected function saveSpreadsheet(Spreadsheet $spreadsheet): string;
+
+    /**
```
kevinpapst, line 658: please remove phpdoc, doesn't add any value

src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -632,4 +654,28 @@ protected function getFileResponse($file, $filename): Response
...
+     * Get the value of durationFormat
+     *
+     * @return bool
+     */
+    public function getDurationFormat(): bool
+    {
+        return $this->durationFormat;
+    }
+
+    /**
```
kevinpapst, line 668: please remove phpdoc, doesn't add any value

kevinpapst:
I am not sure if that PR does the right thing.
The problem is that you force two digits instead of relying on the locale specific behavior.
The test that you changed for example regarding `14 ¥` vs `13,75 ¥` is a great example: Yen doesn't know digits, this change would introduce a bug. 
All I can currently think of is adding some kind of configuration to overwrite the amount of digits (in invoices only?).

src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -632,4 +654,28 @@ protected function getFileResponse($file, $filename): Response
      * @throws \Exception
      */
     abstract protected function saveSpreadsheet(Spreadsheet $spreadsheet): string;
+
+    /**
+     * Get the value of durationFormat
+     *
+     * @return bool
+     */
+    public function getDurationFormat(): bool
```
kevinpapst, line 663: bools should use "isser" => `isDecimalDuration()`

src/Export/Base/AbstractSpreadsheetRenderer.php
```diff
@@ -632,4 +654,28 @@ protected function getFileResponse($file, $filename): Response
...
+    }
+
+    /**
+     * Set the value of durationFormat
+     *
+     * @param bool $durationFormat
+     *
+     * @return self
+     */
+    public function setDurationFormat(bool $durationFormat)
```
kevinpapst, line 675: please rename `setIsDecialDuration`

src/Twig/Extensions.php
```diff
@@ -249,6 +249,10 @@ private function initLocale()
         $this->locale = $locale;
         $this->numberFormatter = new NumberFormatter($locale, NumberFormatter::DECIMAL);
         $this->moneyFormatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
+
+        // returns always two decimal digits
+        $this->numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
```
kevinpapst, line 254: These lines are the most critical ones and I doubt if that is correct for every locale. 
You are skipping the configured ICU formatting rules and replacing them by a custom implementation. 
Even if the current system has flaws (which I am sue there are) the change to a hard coded value seems to be a breaking change that is likely wrong when you have customers is different countries with different currencies.

kevinpapst:
Thanks for your contribution @fredyb 👍 

Closing as this is unmergable and no feedback was given in one year...

PR #1790, opened 2020-06-24
Fixed SV locale date format

Oscariremma:
The Swedish locale didn't have the correct date format. I have fixed this and verified that it works.

## Description
Fixed the Swedish locale date format

## Types of changes
- [x] Added Swedish locale date format

## Checklist

### DIFF
--- a/config/packages/kimai.yaml
+++ b/config/packages/kimai.yaml
@@ -292,9 +292,6 @@ kimai:
             date_time: 'd. m.  H:i'
         sv:
             duration: '%%h:%%m tim'
-            date_time_type: 'yyyy-MM-dd HH:mm'
-            date_type: 'yyyy-MM-dd'
-            date: 'd-m-Y'
             date_time: 'd/m H:i'
         pl:
             date_time_type: 'dd. MM. yyyy HH:mm'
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks @Oscariremma for your contribution!

So you want to use 3 different kind of formats for the date part. 
From my POV that looks disturbing and I don't believe that it improves the UX.

Can you share a link to some documentation about Swedish date formats?

Oscariremma:
Hi @kevinpapst 

I have double checked everything and in general we follow the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) so date_time_type, date_type and date where the same as in EN and didn't need changing. I am sorry for not checking this beforehand. 

But the thing that got me to look in to this in the first place was that when I exported my time table to PDF, the format that was used on the individual entries (date_time) used m-d for the date. In Sweden we never write month before day and when you only write the day and the month we often use a / instead of -. If you want some kind of documentation on this the swedish wikipedia page on [date ](https://sv.wikipedia.org/wiki/Datum) explains how we generally write dates.

So the only change that i feel is needed is to swap the day and month in the date_time format and replace the - with a /.

kevinpapst:
Thanks for double checking this! Looks good to me (if I understand the wiki page correctly).

kevinpapst:
Thank you @Oscariremma 👍

PR #1836, opened 2020-07-21
invoice: allow to increase number, customer specific counter

### LINKED ISSUE #1554: invoice counter for customer
We use an invoice number system starting with the year, followed by the number of the invoice of the year, followed by the number of the invoice of the customer in this year. 

Let's say we already issued 20 invoices in 2020 and we want to issue the second invoice to Customer A. The invoice number would look like 2020-021-02

Thus, I would appriciate if a year counter for the customer like ``{cyc}`` could be added.

### LINKED ISSUE #1673: manipulate invoice counter
Is there a way to manipulate the invoice counter in a way that it does not start from 1?

Practical application: New user installs Kimai in June; until then, the user has already issued 30 invoices and therefore wants to have the counter start at 31 (or the feature is introduced during the year and in order to use it in the current calendar year, the counter would need to be modified so that it starts with a higher number than 1)

kevinpapst:
## Description

Added customer specific counter variables: 
- `{cc}` - customer counter all time
- `{ccy}` - customer counter for this year
- `{ccm}` - customer counter for this month
- `{ccd}` - customer counter for this day

Fixes #1554

Allow to add a increment to the calculated invoice number.
Works for `c` and `cy` and `cm` and `cd` and `cc` and `ccy` and `ccm` and `ccd`.
You can add a `+x` eg `+73` to the format: `{cy+73}` or formatted `{cy+73,3}`.

Fixes #1673 

## Types of changes

## Checklist

### DIFF
--- a/src/Invoice/NumberGenerator/ConfigurableNumberGenerator.php
+++ b/src/Invoice/NumberGenerator/ConfigurableNumberGenerator.php
@@ -62,19 +62,46 @@ final class ConfigurableNumberGenerator implements NumberGeneratorInterface
 
         preg_match_all('/{[^}]*?}/', $format, $matches);
         foreach ($matches[0] as $part) {
-            $formatter = null;
+            $formatterLength = null;
+            $increaseBy = 1;
+
             $tmp = str_replace(['{', '}'], '', $part);
 
-            // number format
-            if (substr_count($tmp, ',') !== 0) {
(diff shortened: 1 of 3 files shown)


groovyrb:
Hi @kevinpapst How to change the file name for invoice which saved because if we regenerate it with different template it tels somesheetname.doc already exist, so how can i change the file name which saved to the system every time.

PR #1840, opened 2020-07-25
Icons

kevinpapst:
## Description

Not entirely sure about this PR, as it adds some overhead to each request/memory without knowing if it is really used.
Previously icons were registered with a dedicated name, now there are more than 1500 icons and not even 20% of them are used.

- make all fontawesome icons available via alias
- added profiler panel for icons (in dev environment only)

## Types of changes

## Checklist


kevinpapst:
Gave it a second thought and decided to ditch this one. Closing

PR #1865, opened 2020-08-03
added js plugin to handle native notifications

kevinpapst:
## Description

Just a technical proof of concept and a good point to start some discussions.

The included plugin is just a simple API to trigger native/desktop notification in browsers that support it.
So this feature should be seen as a "goodie" and no critical features should depend on it, at the end every user could block those notification as well.

I envision a new preference page with a collection of features that can be de-/activted:
- warn after X hours for a break
- warn about long running entries
...
- a button to de-/activate these notifications (FF only allows to ask for permissions after clicking an element)

I am sure there are a couple of issues already existing, that talk about such features.

**TODO**
- [ ] add a section on the profile page (?) to show current status and allow to re-ask for permission, in case it was declined at first
- [ ] use new "max length for entry" for a first proof of concept notification

## Types of changes

## Checklist

### DIFF
--- a/assets/js/plugins/KimaiNotification.js
+++ b/assets/js/plugins/KimaiNotification.js
@@ -1,0 +1,83 @@
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/*!
+ * [KIMAI] KimaiAlert: notifications for Kimai
+ */
+
+import KimaiPlugin from "../KimaiPlugin";
+
+/**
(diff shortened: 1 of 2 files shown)


assets/js/plugins/KimaiNotification.js
```diff
@@ -0,0 +1,83 @@
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/*!
+ * [KIMAI] KimaiAlert: notifications for Kimai
```
kevinpapst, line 9: copy & paste

kevinpapst:
merged manually into #2902

PR #1869, opened 2020-08-05
Timesheet Summary

jankowski-michal:
## Description
Preference to show summary for whole searched timesheet records. 
Adds another row to results, similar to the daily summary, but sums all queried records.

## Types of changes

## Checklist

### DIFF
--- a/src/Controller/TimesheetAbstractController.php
+++ b/src/Controller/TimesheetAbstractController.php
@@ -122,7 +122,7 @@ abstract class TimesheetAbstractController extends AbstractController
         $durationTotal = $this->repository->getDurationTotal($query);
         $summaryStartDate = $query->getBegin();
         $summaryEndDate = $query->getEnd();
-        $summaryDuration = sprintf('%02d:%02d h ', ($durationTotal/3600),($durationTotal/60%60));
+        $summaryDuration = sprintf('%02d:%02d h ', ($durationTotal / 3600), ($durationTotal / 60 % 60));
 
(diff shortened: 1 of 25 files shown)


kevinpapst:
Hi @jankowski-michal and thanks for the contribution!
Does that sum up the entries on the first page or of all records / all pages?

jankowski-michal:
The sum is for all records on all pages, there is also information of dates from-to as on screenshot

Selection_027

kevinpapst:
In that case let me get straight to the point: we can't use it like that.  It might work for your use case, but what if someone has tenth of thousand of entries (yes, bigger customers reach these numbers easily)? 

Do you really want to fetch ALL records with all joined elements only to calculate one value? 
Or would you rather build an explicit database query targeted to that specific feature?

If you want to displayed the money as well, then you need to take care of the different customer currencies as well.

And why did you change so many (unrelated) translations? Do you speak all these languages?
Please let's talk about each feature briefly, before implementing it.

jankowski-michal:
Ad. Or would you rather build an explicit database query targeted to that specific feature?
You're right, I'll update it.

Ad. If you want to displayed the money as well, then you need to take care of the different customer currencies as well.
I'll remove currencies.

Ad. And why did you change so many (unrelated) translations? Do you speak all these languages?
Sorry, it was rebase issue. Will update.

jankowski-michal:
I'll close this review and create another one because intellij idea still reformats translation files.

Issue #1858, opened 2020-07-31
Calendar view displays wrong start and end hour


jpaczes:
Created activity is wrongly displayed in calendar view, calendar shows start and end date 2 hours earlier than actual activity times.

**To Reproduce**
I managed to reproduce this problem on demo version (1.10 stable ) available on https://demo.kimai.org when i switched to super admin and changed setting to match one that we use in our company.

![image](https://user-images.githubusercontent.com/26572441/89041419-a274f700-d345-11ea-88ba-5058a7d0b281.png)
![image](https://user-images.githubusercontent.com/26572441/89041483-bc163e80-d345-11ea-9cd1-f29b900bf678.png)
![image](https://user-images.githubusercontent.com/26572441/89041531-d2bc9580-d345-11ea-98f6-f7861f70ab03.png)
![image](https://user-images.githubusercontent.com/26572441/89041574-e5cf6580-d345-11ea-9ea1-a10af0da5910.png)

After that i created new activity in calendar view

![image](https://user-images.githubusercontent.com/26572441/89041759-31820f00-d346-11ea-8985-8f9c5cd85f64.png)

and it's displayed like this

![image](https://user-images.githubusercontent.com/26572441/89042216-df8db900-d346-11ea-8fbd-f32a2d9b2319.png)

Other places like 'My times', 'Timesheets' or Record details display correct start and end time for this activity.

kevinpapst:
Verified! Thanks for the report.
That has nothing to do with all these settings, there is something off with the timezone in the calendar.

kevinpapst:
@jpaczes are you using master in your environment?

I have the feeling that this bug is ONLY related to the current dev version and has nothing to do with the calendar, as it was working in the past.

kevinpapst:
Fixed by #1888

PR #1879, opened 2020-08-09
fix calender view #1858

fredyb:
## Description
fix issue #1858 Calendar view displays wrong start and end hour 
## Types of changes

## Checklist

### DIFF
--- a/templates/calendar/user.html.twig
+++ b/templates/calendar/user.html.twig
@@ -78,12 +78,14 @@
                                             color = record.project.customer.color;
                                         }
                                     }
+                                    var begin = new Date(record.begin);
+                                    var end = new Date(record.end);
                                     apiEvents.push({
                                         id: record.id,
                                         title: record.activity.name,
                                         description: record.description,
-                                        start: record.begin,
-                                        end: record.end,
+                                        start: begin,
+                                        end: end,
                                         activity: record.activity.name,
                                         project: record.project.name,
                                         customer: record.project.customer.name,


kevinpapst:
@fredyb I gave it a try and it looks good. 

And even though it seems to work, I am not sure about the end date without a protective null check, because:
Bildschirmfoto 2020-08-12 um 14 36 08
This could lead to problems with running records (with `null` end date).

I added two console logs, one in the `events` function and one in `eventRender` and it seems that the calendar somehow magically filters/removes such end values:
Bildschirmfoto 2020-08-12 um 14 40 10

But as I plan to update the currently used version of the calendar soon, I'd like to prevent relying on such a behavior.

Additionally: this change does not fix the original problem - it only works for datetimes which were stored in the currently used timezone. You can try that by switching your users timezone to something far away from your (try `Pacific/Tongatapu` for example). 
Then create a new record and switch the timezone back afterwards.
Now check the calendar: still wrong date.

The original problem lies in the API, see the fix in #1888 - the datetimes were simply returned in UTC always.

I really do appreciate your contribution and it definitely helped me to find the solution for the bug.
But I think we should close this without merging and wait for the update to FullCalendar v5 before changing anything within the calendar logic. Are you okay with that?

fredyb:
@kevin: that's fin for me. May solution helps gaping until you have fix it #1888.

kevinpapst:
Thanks @fredyb !

Original issue was fixed, so closing here.

PR #1797, opened 2020-06-28
xlsx export render don't export decimal duration of timesheet if user…

fredyb:
… preferences 'Use decimal duration in export' is checked #1795 

## Description
A clear and concise description of what this pull request adds or changes.

# Types of changes
- [x] Bug fix (non-breaking change which fixes an issue) #1795
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I verified that my code applies to the guidelines (`composer kimai:code-check`)
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [x] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

Due to a lack of knowledge, I was unable to adapt the phpunit test.

PR #1880, opened 2020-08-09
Xlsx export export decimal fix (#1797)

fredyb:
## Description
fix xlsx export render don't export decimal duration of timesheet if user… (#1797)

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist


kevinpapst:
Thank you @fredyb and sorry for not responding earlier!
Your code was very good example and I used it to fix this bug via #2660 - will be released with 1.15.

PR #1930, opened 2020-08-27
WIP: OAuth Login Support

### LINKED ISSUE #1317: OAuth Login Support - Google Social Provider
Hi, is there any plan to support OAuth Login? We would like to sign up / sign in our users via Google's GSuite. 
Thank you

robjuz:
## Description
Resolve #1317 
This PR implements oauth login with google and gitlab. It can be simple extended with other providers.

## Types of changes

## Checklist


kevinpapst:
First of all: Thanks for having a look into this topic!

A couple of remarks before going into a full review: 
- I am more a friend of HWIOAuthBundle - more stars, more users ... no packages required from other repositories needed ... Why did you choose this one and would you change it?
- Routes should be located below `/auth/` - this is the prefix for all authentication routes and already prepared for anonymous access
- Everything needs to be fully optional and the configuration needs to be done via `.env` and `config/packages/local.yaml` only

For the last point you can find extensive examples in the LDAP and SAML authenticators (both code and documentation).

robjuz:
@kevinpapst I will take a look into the other package. Thanks for all the hints.

robjuz:
@kevinpapst 
I just check the HWIOAuthBundle package. Now I know why I choose the other one. You can't use self hosted gitlab with HWIOAuthBundle.

kevinpapst:
Says who? Why shouldn't that work if there is already Gitlab support and a generic provider. 
See https://gist.github.com/lenybernard/2b5d5517104b595f14dbcd4411f6a901 for example

PR #1951, opened 2020-09-04
Fix cwd in kimai reload

tobybatch:
## Description
If the command ```kimai:reload ``` is run from a folder other than the project root (e.g. [startup.sh](https://github.com/tobybatch/kimai2/blob/main/startup.sh#L41) the the config folder can't be found and the error ```[ERROR] File or directory "config" is not readable.``` is reported.

## Types of changes

## Checklist
- N/A I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))


kevinpapst:
LOL, I worked today on the similar problem (just for bundle installer commands).
I'll have a look at the other commands and merge it later, thanks Toby!

kevinpapst:
Thanks @tobybatch !

PR #1952, opened 2020-09-05
User: configure first day of week

### LINKED ISSUE #1572: Calendar date picker not using i18n setting
**Describe the bug**
When user preference is set to English, so first day of week is Sunday, the date-picker still uses Monday as first day of week, although the main calendar view displays the correct order of days. See screenshot.

**To Reproduce**
1. Set users's preference to Enlish so calendar starts on Sunday
2. Add entry in timesheet.
3. Click on *From* field to bring up calendar
4. Observer calendar has first day of week as Monday.

**Screenshots**

Screen Shot 2020-03-19 at 11 31 12

**Desktop/Smartphone**
Please complete the following information if this seems to be a frontend issue:
 - Firefox 74.0

**Additional context**
 - Kimai version: 1.8
 - PHP version: php-7.2

### LINKED ISSUE #1732: Configurable first day of the week

Problem is the first day of the week being bound to the users locale.

While that fits in an environment where the user can choose its native language it doesn't work for the userbase with English as UI language, because English might only be a fallback if their native language is not available. Maybe there are other languages which are used in multiple timezones (eg. French).

Until 1.9 the datepicker was hardcoded to Monday as being the first day of the week.
With the translation to hebrew in #1724 that behavior was changed and adapted to the timesheet calendar, which already used the `dow` from momentjs.

The `first day of the week` should be configurable. But the question is: where? Is that a user specific setting or a global setting? Being bound to the locale definition of Kimai or available through the system settings screen (I prefer latter one).

Stick to the definition of moment and create empty locales without translations for every language. Then the user needs to choose its own language but will see English as fallback.

kevinpapst:
## Description

Allows to select between `monday` and `sunday` as first day of the week.
This setting is user specific and ONLY used in:
- the calendar page (if weekends aren't shown, the setting `sunday` has no effect and the week starts with monday)
- all date and date-time pickers

Fixes #1572 
Fixes #1732 

## Types of changes

## Checklist


Recoil1980:
@kevinpapst Thank you for this update!  I followed the [upgrade](https://www.kimai.org/documentation/updates.html) guide to get version 1.10.2, but do not see the option for selecting the starting day in any of the user settings.  Can you point me in the right direction?

kevinpapst:
The milestone that issue is attached to says 1.11. So its coming with the next release

PR #1982, opened 2020-09-18
Execute migrations in kimai:reset-dev command

kx1000:
## Description
Following [this](https://www.kimai.org/documentation/developers.html) guide I had an error: `SQLSTATE[42S02]: Base table or view not found: 1146 Table 'kimai2.kimai2_sessions' doesn't exist`.

`kimai:reset-dev` command executes `doctrine:schema:create` wchich generates schema from entities, but schema required for SessionHandler is stored in migration. Even after this command it's imposibble to execute migrations, because `kimai:reset-dev` executes also `doctrine:migrations:version --add --all`, so we have confusing `migration_versions` table.

## Types of changes

## Checklist

### DIFF
--- a/src/Command/ResetCommand.php
+++ b/src/Command/ResetCommand.php
@@ -92,7 +92,7 @@ EOT
         if ($this->askConfirmation($input, $output, 'Do you want to drop and re-create the schema y/N ?')) {
             try {
                 $command = $this->getApplication()->find('doctrine:schema:drop');
-                $command->run(new ArrayInput(['--force' => true]), $output);
+                $command->run(new ArrayInput(['--force' => true, '--full-database' => true]), $output);
             } catch (Exception $ex) {
                 $io->error('Failed to drop database schema: ' . $ex->getMessage());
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks, I didn't think about this command, haven't used for too long.
I never used the `full-database` flag. Is that required, because there is no entity for the session and the table is therefor not deleted?

kx1000:
Yes, this flag is required to delete tables out of the entities too. :)

kevinpapst:
Hm, but is that an expected behavior that a command for Kimai deletes unrelated tables?
We cannot guarantee that the user doesn't have any other tables in the DB that do not belong to Kimai (unlikely for a dev, but users do all kinds of strange things).

What do you think about instead calling `bin/console doctrine:query:sql "DROP kimai2_session"` from the reset command?

kx1000:
It's good idea, I pushed changes. I added `IF EXISTS` statement, because of possibility running command with empty databse.

kevinpapst (approved):
Thanks 👍

PR #2028, opened 2020-10-08
hide create copy button for users if punch mode is active

hmr-it-jr:
## Description
If punch mode is active users are only allowed to start and stop times. They were allowed to create a duplicate of a timesheet entry. The pull requests prevents this. The "create copy" button is hidden if punch mode is active.

## Types of changes

## Checklist

### DIFF
--- a/src/Voter/TimesheetVoter.php
+++ b/src/Voter/TimesheetVoter.php
@@ -11,6 +11,9 @@ namespace App\Voter;
 
 use App\Entity\Timesheet;
 use App\Entity\User;
+use App\Security\AclDecisionManager;
+use App\Security\RolePermissionManager;
+use App\Timesheet\TrackingModeService;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 
 /**
@@ -27,6 +30,7 @@ class TimesheetVoter extends AbstractVoter
     public const VIEW_RATE = 'view_rate';
     public const EDIT_RATE = 'edit_rate';
     public const EDIT_EXPORT = 'edit_export';
+    public const DUPLICATE = 'duplicate';
 
     /**
      * support rules based on the given $subject (here: Timesheet)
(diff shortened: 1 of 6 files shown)


kevinpapst:
Thanks!

Have you tried adding it to the permission voter instead?
https://github.com/kevinpapst/kimai2/blob/master/src/Voter/TimesheetVoter.php#L103

hmr-it-jr:
@kevinpapst like this?

src/Voter/TimesheetVoter.php
```diff
@@ -173,4 +191,18 @@ protected function canDelete(User $user, Timesheet $timesheet): bool
 
         return true;
     }
+
+    protected function canDuplicate(User $user, Timesheet $timesheet): bool
+    {
+        if ($timesheet->isExported() && !$this->hasRolePermission($user, 'edit_exported_timesheet')) {
```
kevinpapst, line 197: Lets re-use the code:
```suggestion
        if (!$this->canEdit($user, $timesheet)) {
```
but actually ... hm ... I don't think that the check is even required.
Duplicating a timesheet shouldn't be a problem, even if the original one is exported. Or do you see a good reason for preventing it?

kevinpapst (changes requested):
Looks much better then the last approach 😁 
But I guess you have to adapt at least the App\Tests\Voter\TimesheetVoterTest for the new constructor.

src/Voter/TimesheetVoter.php
hmr-it-jr: In my opinion duplicating is a problem. Because the time get's doubled. Let's assume punch mode is active. Users can only punch in and out.

If they punch in and out they are only able to creat times when they punch in -> out.

Looks on a normal day like this:
grafik

if they duplicate an entry - the time get's doubled. So if you check at the end of the month you just can't check the total duration if it matches daily working time * working days of current month -> you have to check for overlapping times, too.

-----

Just found out copying is not possible if "allow overlapping time entries" is disabled (in latest master) but possible in latest stable. Guess this is related to the same thing we already discussed last week or so.

How to proceed? Close pull request and delete it?

kevinpapst: I wasn't questioning this PR.
What I meant is: do we need the check about the export state in the voter. 

I think it is valid to hide the button in punch-in-out mode.

kevinpapst:
Thanks @hmr-it-jr 👍 
But ... this is not required any longer, because 1.15 will open an edit-dialog with pre-filled fields instead of duplicating the entire record.

PR #2031, opened 2020-10-08
WIP - improve export screen for unexpected permission setups

### LINKED ISSUE #1473: edit_export_other_timesheet
Possibly not a bug: 
According to the documentation, with the edit_export_other_timesheet permission one can 'Set the export state of for other users timesheet records'. So, I assume that if I remove the permission for a specific role that the user with that role may not set the export state of other users records.

I have removed the edit_export_other_timesheet permission for the teamleader role and logged in as the assigned teamleader. I was still able to set the 'Mark as exported' flag in the export form and after export the status was 'cleared'.

Or is there another way to prevent a teamleader to mark other records as exported?

Thank you.

kevinpapst:
# Description

Fixes #1473 

prevent to select other user if create_export is allowed, but not view_other_timesheet

## Types of changes

## Checklist

### DIFF
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -14,7 +14,6 @@ use App\Export\ServiceExport;
 use App\Form\Toolbar\ExportToolbarForm;
 use App\Repository\Query\ExportQuery;
 use App\Repository\TimesheetRepository;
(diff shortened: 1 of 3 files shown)


kevinpapst:
closing in favor of #2182

Issue #2057, opened 2020-10-24
Dockerfile build fails


kulturbande:
**Describe the bug**
If I run docker build from the [documenation](https://www.kimai.org/documentation/docker.html) it results in an error at the end. The container is usable, but results in some weird errors, if I want to use other invoice templates for example.

**To Reproduce**
1. run `docker build -t kimai/kimai2:dev .`
2. an error occurs in the fouth step

**Logfile**
```
Step 4/8 : RUN git clone https://github.com/kevinpapst/kimai2.git /opt/kimai &&     sed "s/prod/dev/g" /opt/kimai/.env.dist > /opt/kimai/.env &&     composer install --working-dir=/opt/kimai --dev --optimize-autoloader &&     /opt/kimai/bin/console doctrine:database:create &&     /opt/kimai/bin/console doctrine:schema:create &&     /opt/kimai/bin/console doctrine:migrations:version --add --all &&     /opt/kimai/bin/console cache:warmup &&     chown -R www-data:www-data /opt/kimai/var &&     chown www-data:www-data /opt/kimai/vendor/mpdf/mpdf/tmp
 ---> Running in 778e7648cc31
Cloning into '/opt/kimai'...
You are using the deprecated option "dev". Dev packages are installed by default now.
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested PHP extension ext-xsl * is missing from your system. Install or enable PHP's xsl extension.
  Problem 2
    - Installation request for lorenzo/pinky 1.0.5 -> satisfiable by lorenzo/pinky[1.0.5].
    - lorenzo/pinky 1.0.5 requires ext-xsl * -> the requested PHP extension xsl is missing from your system.
  Problem 3
    - lorenzo/pinky 1.0.5 requires ext-xsl * -> the requested PHP extension xsl is missing from your system.
    - twig/inky-extra v3.0.3 requires lorenzo/pinky ^1.0.5 -> satisfiable by lorenzo/pinky[1.0.5].
    - Installation request for twig/inky-extra v3.0.3 -> satisfiable by twig/inky-extra[v3.0.3].

  To enable extensions, verify that they are enabled in your .ini files:
    - /usr/local/etc/php/php.ini
    - /usr/local/etc/php/conf.d/docker-php-ext-gd.ini
    - /usr/local/etc/php/conf.d/docker-php-ext-intl.ini
[...]

PR #2088, opened 2020-11-03
WIP: Updated develop docker image

adrianrudnik:
## Description

Current development docker image does no longer work due to missing php-extensions, missing and deprecated webserver-bundle approach and changes to default security requirements. Related issue #2057.

Things I had to change:

- `/opt/kimai` to `/var/www/html`, as this is the default location for many images like the apache ones and the symfony-cli commands.
- Change the default password for the demo `admin` account, the one mentioned in the docs fails as it's not 8 chars or more long.

Things I updated:

- Upgrade to PHP 7.4.12
- Added XSL/XML, opcache and freetype support
- Added symfony cli, as webserver-bundle is no longer included and this is the 5.x way of spooling up dev servers easily.
- Used fixed versions for composer, symfony-cli and PHP to avoid build cache conflicts.
- Added a default user to the basic SQLite setup with `admin` and `adminpassword`.
- Removed the composer memory limit
- Forced the port to stay at 8001, through symfony-cli uses 8000 as default.
- Disabled optional CA/tls support offered by symfony-cli, not important for development.
- Added `symfony check-requirements` and `symfony security:check` on build to give a hint if anything is missing.
- Added some optimizations that are also useful to dev from https://symfony.com/doc/current/performance.html

Things I'm unsure about:

- The current approach has a minor drawback: Once you mount `/var/www/html`, the default sqlite database setup will vanish, as it will try to search a non-existant on the docker host. Maybe put the sqlite default-empty-database somewhere else, so you can mount the source and still be ready to develop?

Things to mention:

`d:schema:create`, then `d:migrations:version --add` no longer works. It's missing kimai2_sessions. Flipping it
to `d:migrations:migrate` first, then `d:schema:update` leads to alot(!) of changes that are not captured by the migrations.
Sadly this is the price paid with doctrine and their "we support many drivers, but hopefully you're on mysql" behaviour,
same situation on PostgreSQL where you can end up in a schema state that will never migrate fully. The database
will never be in sync with SQLite due to how doctrine inspection always leads to a diff.

How to spool up a demo instance:

```bash
docker build -t kimai/kimai2:dev .
docker run --rm -it -p 8001:8001 kimai/kimai2:dev
```

Related todos:

- [ ] Need to update the [documentation](https://www.kimai.org/documentation/docker.html)
- [ ] Documentation should mention the image is for dev purposes only and will expose sensitive system information to the public, including database credentials and other related system information.
- [ ] Fix [CVE-2020-15094](https://symfony.com/cve-2020-15094)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+# The given Dockerfile is for development only and will always
+# clone a clean version from github, so we ignore everything on build.
+**
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ RUN set -ex \
         opcache
 
 # Install Xdebug
-ARG XDEBUG=1
+ARG XDEBUG=0
 
 RUN set -ex \
     && if [ "$XDEBUG" = "1" ]; then \
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,29 @@ RUN set -ex \
         zip \
         opcache
 
+# Install Xdebug
+ARG XDEBUG=1
+
+RUN set -ex \
+    && if [ "$XDEBUG" = "1" ]; then \
+      mkdir /tmp/build \
+      && curl -sSL https://github.com/xdebug/xdebug/archive/2.9.8.tar.gz | tar -xzC /tmp/build \
+      && cd /tmp/build/xdebug-* \
+      && phpize \
+      && ./configure --enable-xdebug \
+      && make \
+      && make install \
+      && rm -rf /tmp/build \
+      && echo "zend_extension=xdebug.so" > /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.idekey=PHPSTORM" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_enable=1" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_connect_back=1" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_host=host.docker.internal" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_port=9000" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_autostart=1" >> /usr/local/etc/php/conf.d/xdebug.ini \
+      && echo "xdebug.remote_handler=dbgp" >> /usr/local/etc/php/conf.d/xdebug.ini; \
+    fi
+
 WORKDIR /var/www/html
 
 # Install composer
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,6 +92,10 @@ RUN set -ex \
     && git clone --depth 1 https://github.com/kevinpapst/kimai2.git . \
     && composer install --optimize-autoloader
 
+# Setup demo specific ENV
+ENV APP_ENV=dev \
+    DATABASE_URL=sqlite:////var/data/kimaidemo.sqlite
+
 # Create default SQLite database with a default user and compile it into the
 # image to offer some default setup to spool up incase of a demo.
 RUN set -ex \
@@ -102,10 +106,6 @@ RUN set -ex \
 RUN set -ex \
     && symfony check:requirements \
     && symfony security:check --disable-exit-code
-
-# Setup demo specific ENV
-ENV APP_ENV=dev \
-    DATABASE_URL=sqlite:///var/data/kimaidemo.sqlite
 
 EXPOSE 8001
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,6 @@
 # file that was distributed with this source code.
 FROM php:7.4.12-cli
 
-ENV APP_ENV dev
-
 # Install tools and dependencies
 RUN set -ex \
     && apt-get update \
@@ -84,11 +82,10 @@ RUN set -ex \
     && sed -i 's/;opcache.memory_consumption=128/opcache.memory_consumption=256/g' /usr/local/etc/php/php.ini \
     && sed -i 's/;opcache.max_accelerated_files=10000/opcache.max_accelerated_files=20000/g' /usr/local/etc/php/php.ini \
     && sed -i 's/log_errors_max_len = 1024/log_errors_max_len = 8192/g' /usr/local/etc/php/php.ini \
-    && chown -R www-data:www-data /var/www
+    && mkdir -p /var/data \
+    && chown -R www-data:www-data /var/www /var/data
 
 USER www-data:www-data
-
-WORKDIR /var/www/html
 
 # Create shallow clone of kimai and configure it as dev role
 RUN set -ex \
@@ -105,6 +102,10 @@ RUN set -ex \
 RUN set -ex \
     && symfony check:requirements \
     && symfony security:check --disable-exit-code
+
+# Setup demo specific ENV
+ENV APP_ENV=dev \
+    DATABASE_URL=sqlite:///var/data/kimaidemo.sqlite
 
 EXPOSE 8001
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,54 +2,88 @@
 #
 # For the full copyright and license information, please view the LICENSE
 # file that was distributed with this source code.
-FROM php:7.2.9-apache-stretch AS tmp_kimai2_base
+FROM php:7.4.12-cli
 
-RUN apt update && \
-    apt install -y --allow-unauthenticated \
-        git \
-        haveged \
-        libicu-dev \
-        libjpeg-dev \
-        libldap2-dev \
-        libldb-dev \
-        libpng-dev \
-        mysql-client \
-        unzip \
-        wget \
-        zip \
-        && \
-    EXPECTED_SIGNATURE="$(wget -q -O - https://composer.github.io/installer.sig)" && \
-    php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
-    ACTUAL_SIGNATURE="$(php -r "echo hash_file('sha384', 'composer-setup.php');")" && \
-    if [ "$EXPECTED_SIGNATURE" != "$ACTUAL_SIGNATURE" ]; then >&2 echo 'ERROR: Invalid installer signature'; rm composer-setup.php; exit 1; fi && \
-    php composer-setup.php --quiet && \
-    rm composer-setup.php && \
-    mv /var/www/html/composer.phar /usr/bin/composer && \
-    docker-php-ext-install \
+ENV APP_ENV dev
+
+# Install tools and dependencies
+RUN set -ex \
+    && apt-get update \
+    && apt-get -y install \
+        git haveged unzip zip wget \
+        libicu-dev libcurl4-openssl-dev libjpeg-dev libldap2-dev \
+        libldb-dev libpng-dev libxslt-dev libxml2-dev libssl-dev \
+        libzip-dev libfreetype6-dev libjpeg62-turbo-dev libmcrypt-dev \
+        libwebp-dev libmagickwand-dev \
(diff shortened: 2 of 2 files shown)


kevinpapst:
Thanks a lot 👍 I am really under pressure right now work-wise, so doing a proper review takes a bit.

Just a quick first feedback: 
Can we use non-fixed version of the tools? I know that the docker file needs to be maintained, but having the least possible work would be fantastic 😁 

> Doctrine migrations

Kimai MUST run correctly with the single `bin/console kimai:install` call (otherwise I need to check it), so the 3 doctrine commands can be replaced.

> Sadly this is the price paid with doctrine and their "we support many drivers, but hopefully you're on mysql" behaviour,

Hahaha, that caused me a lot of headaches in the past, even when switching between MySQL and MariaDB. 
The best approach for me was to set the database version manually. I have [added an env var](https://github.com/kevinpapst/kimai2/blob/master/config/packages/doctrine.yaml#L18) to do that, but I never tried this approach with SQLite

adrianrudnik:
> Thanks a lot +1 I am really under pressure right now work-wise, so doing a proper review takes a bit.

Sure, no worries, same here.

> Can we use non-fixed version of the tools? I know that the docker file needs to be maintained, but having the least possible work would be fantastic grin

Sure, I just try to avoid it, i.e. are you ready for composer 2.x which would install on a default composer install now? There were some deprecations and warnings about it if I remember correctly. Same goes for PHP upgrades. The "bad" thing about docker is that it will always prefer the local available image: 7.4.3 was pulled to your machine 2 months ago? You build an image with `php:7.4` on your machine, so it will much likely build it with PHP 7.4.3 instead of 7.4.12 and you must ensure you use `--pull` on every docker command related to it. Fixing the version into the tag or module will avoid any upgrade pains and be specific about the correct or min/max versions to use.

> Kimai MUST run correctly with the single `bin/console kimai:install` call (otherwise I need to check it), so the 3 doctrine commands can be replaced.

Ah good, I'll try to change that and replace the other stuff related to the setup that becomes obsolete. Same goes for the db version config, maybe it helps on SQLite :+1:

kevinpapst:
> Sure, I just try to avoid it, i.e. are you ready for composer 2.x which would install on a default composer install now?

Yes, I added compatibility for composer 2 with the 1.11 release. 

Regarding the docker version: my docker knowledge is so small, I won't/can't be of any help with that decision 😁 do what you think is best. In the end it is a development container, so its not that big of a problem if some version gets outdated.

You can update the docs via this repo: https://github.com/kimai/www.kimai.org/blob/master/_documentation/docker.md

tobybatch:
@adrianrudnik If you have docker question you may want ask us over at https://github.com/tobybatch/kimai2/issues

We've been supporting Kevin's work on Kimai by providing containers for a couple of years now.  Kevin will be first to answer "I don't know much about dockers but...." and the go on to try and help you anyway. :smile: 

@kevinpapst Do you want to swap out the docker page with a list of links to the various docker solutions and remove the docker file from core repo.  It's out of date and we've been woeful at keeping it up to date.  If you do I can provide a PR for the the page that just lists and a quick note explaining what these are:

 * https://github.com/tobybatch/kimai2/issues
 * https://hub.docker.com/r/dysonsphere/kimai2/
 * https://github.com/felixhummel/kimai-in-docker/tree/kimai2
 * https://github.com/Haidy777/kimai2/blob/docker/Dockerfile

And Adrian's if he makes it public too.

Dockerfile
```diff
@@ -2,54 +2,88 @@
...
+    && composer install --optimize-autoloader
+
+# Create default SQLite database with a default user and compile it into the
+# image to offer some default setup to spool up incase of a demo.
+RUN set -ex \
+    && bin/console doctrine:database:create -n \
+    && bin/console doctrine:migrations:migrate -n  \
+    && bin/console doctrine:schema:update -n --force \
+    && bin/console cache:warmup -n \
+    && bin/console kimai:create-user admin admin@kimai.local ROLE_SUPER_ADMIN adminpassword
```
kevinpapst, line 80:
```suggestion
    && bin/console kimai:create-user admin admin@example.com ROLE_SUPER_ADMIN password
```

kevinpapst:
I don't want to replace your docker solution @tobybatch - its just about this shot-and-go container for development that is included in this repo (you provided that a long time ago) which is completely outdated and doesn't work anymore. 
So many things changed in between and it is nice to bring it back to life, purely a simple development container for first testing purposes.

And we should have a look at the documentation page and clearly state which container is meant for what purpose and explain shortly how to use them. The other links can be removed, they are completely outdated and the container haven't been updated in a long while.
I started a doc PR here: https://github.com/kimai/www.kimai.org/pull/83 - please add your notes there!

adrianrudnik:
@kevinpapst do you use xdebug or should we enable it for the dev image to allow for runtime debugging?

kevinpapst:
I only enable it if required. XDebug leads (at least on my end) to terrible performance issues in dev mode.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+# The given Dockerfile is for development only and will always
+# clone a clean version from github, so we ignore everything on build.
+**
```
kevinpapst, line 3: Can that potentially interfere with the other docker solution(s)?

Dockerfile
```diff
@@ -2,54 +2,111 @@
...
+        zip \
+        opcache
+
+# Install Xdebug
+ARG XDEBUG=0
+
+RUN set -ex \
+    && if [ "$XDEBUG" = "1" ]; then \
+      mkdir /tmp/build \
+      && curl -sSL https://github.com/xdebug/xdebug/archive/2.9.8.tar.gz | tar -xzC /tmp/build \
```
kevinpapst, line 47: is debug not available via pecl or apt itself?

Dockerfile
```diff
@@ -2,54 +2,111 @@
...
+    && curl -sS https://get.symfony.com/cli/installer | bash \
+    && mv /root/.symfony/bin/symfony /usr/local/bin/symfony
+
+# Install production php.ini and make sore there is a folder to hold the demo database
+RUN set -ex \
+    && cp /usr/local/etc/php/php.ini-development /usr/local/etc/php/php.ini \
+    && sed -i 's/memory_limit = 128M/memory_limit = 512M/g' /usr/local/etc/php/php.ini \
+    && sed -i 's/;opcache.memory_consumption=128/opcache.memory_consumption=256/g' /usr/local/etc/php/php.ini \
+    && sed -i 's/;opcache.max_accelerated_files=10000/opcache.max_accelerated_files=20000/g' /usr/local/etc/php/php.ini \
+    && sed -i 's/log_errors_max_len = 1024/log_errors_max_len = 8192/g' /usr/local/etc/php/php.ini \
```
kevinpapst, line 84: Can we add a longer session timeout as well? The default of 24 minutes is ridiculously short:
```
&& sed -i 's/session.gc_maxlifetime = 1440/session.gc_maxlifetime = 43200/g' /usr/local/etc/php/php.ini \
```

.dockerignore
adrianrudnik: Will take a look again, but as far as I can see the other solutions also depend on a fresh git clone. Docker context is mainly used in a CI, when the commit automatically triggers the CI, the CI clones the repository at that commit and you can just use that instead of doing a manual clone.

Dockerfile
adrianrudnik: Will take a look!

Dockerfile
adrianrudnik: Sure, don't see a reason why not!

Dockerfile
```diff
@@ -2,54 +2,111 @@
...
+    && mkdir -p /var/data \
+    && chown -R www-data:www-data /var/www /var/data
+
+USER www-data:www-data
+
+# Create shallow clone of kimai and configure it as dev role
+RUN set -ex \
+    && git clone --depth 1 https://github.com/kevinpapst/kimai2.git . \
```
xeroc, line 92: It's actually quite uncommon for docker-containers to clone the repo from remote.
Given that the Dockerfile comes with the repo, the sources are already there
and just need to be moved over with a

    ADD . /PATH/WHERE/GIT/WOULD/CLONE/TO

This also enables to make custom build with additional plugins installed into
your locally checked out repository.

Additionally to that, it enables to do per-branch builds using github actions
that can be pushed to docker hub automatically.

Dockerfile
```diff
@@ -2,54 +2,111 @@
...
+    && composer install --optimize-autoloader
+
+# Setup demo specific ENV
+ENV APP_ENV=dev \
```
xeroc, line 96: Usually, a `docker-compose.yml` would be used to set environmental variables.
If you put it in here, you cannot change it when using the container (I believe)

Dockerfile
```diff
@@ -2,54 +2,111 @@
...
+    DATABASE_URL=sqlite:////var/data/kimaidemo.sqlite
+
+# Create default SQLite database with a default user and compile it into the
+# image to offer some default setup to spool up incase of a demo.
+RUN set -ex \
```
xeroc, line 101: This should be done from outside the container, e.g. in a `docker-compose.yml`.
That would allow to use this container in production, and not come with pre
installed admin account like this.

Dockerfile
adrianrudnik: At least in docker-compose the [default behaviour](https://docs.docker.com/compose/environment-variables/) is that any injected env's take precedence over defaults declared via ENV in the dockerfile, so it should fine or do you see any edge case?

Dockerfile
adrianrudnik: Yeah, this is legacy and I just kept it this way. @kevinpapst can decide what he wants.

Pro: cloning a "default" setup: You can spool up the image and have a running copy without any additional cloning, mounting and configuration.

Contra: Image size gets bigger because a code snapshot is included.

You can always decide to mount a different volume over the codes. The thing is: This is for DEV only. So you either spool up an instance for demo/testing (where it would be nice if you do not need to check out and mount the same commit the image was built with) or you mount the sources anyway (overriding the given ones).

Not sure which way to go, so i kept the current one.

Dockerfile
adrianrudnik: This Dockerfile is, per description, **not** for production and only for development. https://github.com/tobybatch/kimai2 does the publishing towards docker hub and they are perfectly fine for that job. This Dockerfile should be for dev only, using the best and simplest way to allow for fast development/fixing/testing when required.

Dockerfile
kevinpapst: Keep in mind: I am not a docker user, so I can only share my own thoughts which are not based on experience or docker know-how.

Default generated `.env` contains `APP_ENV=dev`.
Environment variables overwrite `.env` settings.

If a user wants to change it by using locally mounted code, we would allow that if we remove it here?

kevinpapst: Sorry, default `.env` contains `prod` - forget my complete comment ^^

adrianrudnik: @kevinpapst Yeah, we changed it a while ago, because people made the same mistake, they used this image in PROD and a default of `dev` leads to a situation where every password is visible through the profiler/logs/messages towards the public. I would be OK with a default of `dev` though.

kevinpapst:
@adrianrudnik I am really sorry that your effort went unmerged again.
After a very long time I decided to ditch the Dockerfile completely.
I am simply not capable to maintain it and its better to remove it completely instead of leaving it broken in the repository.

Additionally I removed SQLite support, to be able to concentrate fully on MySQL (see above discussion regarding Doctrine)... - which I documented here: https://www.kimai.org/blog/2021/sqlite-and-ftp-support-removed/

All I can say is: thank you for your ideas and input 👍 which I really appreciate! And sorry for the time that you spent here!

adrianrudnik:
Kein Ding, I've been too busy to go further with this in the last few months sadly. Better to slim down then to further the technical debt :)

kevinpapst:
Thats exactly what I was thinking. 
Stop "wasting time" with technical solutions that are not part of the core product, but concentrate on the important features that people want.

Thanks again @adrianrudnik 😄

PR #2141, opened 2020-11-25
Added DateActivityInvoiceCalculator

Gagasas:
## Description
Added a DateActivityInvoiceCalculator for Invoices. It merges the logic of the ActivityInvoiceCalculator and the DateInvoiceCalculator for filtering between all activities and sum them up to one date.

## Types of changes

## Checklist

### DIFF
--- a/src/Invoice/Calculator/DateActivityInvoiceCalculator.php
+++ b/src/Invoice/Calculator/DateActivityInvoiceCalculator.php
@@ -1,10 +1,16 @@
 <?php
 
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace App\Invoice\Calculator;
 
 use App\Invoice\CalculatorInterface;
 use App\Invoice\InvoiceItemInterface;
-use App\Invoice\InvoiceItem;
 
 /**
  * Calculator for filtering activities and dates.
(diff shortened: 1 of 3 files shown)


src/Invoice/Calculator/DateActivityInvoiceCalculator.php
```diff
@@ -0,0 +1,48 @@
...
+        $date_array = explode(" ", $date);
+
+        // Merges both arrays together
+        $activity_date_array = array_merge($date_array, $activity_array);
+        // Creates from array a readable string for Kimai
+        $activity_date_string = implode(" ", $activity_date_array);
+
+        // Returns the new created string
+        return $activity_date_string;
+    }
```
kevinpapst, line 39:
```suggestion
    protected function calculateSumIdentifier(InvoiceItemInterface $invoiceItem): string
    {
        $id = '';

        if (null !== $invoiceItem->getBegin()) {
            $id .= $invoiceItem->getBegin()->format('Y-m-d');
        }

        $id .= '__';

        if (null !== $invoiceItem->getActivity()) {
            $id .= (string) $invoiceItem->getActivity()->getId();
        } else {
            $id .= 'NULL';
        }

        return $id;
    }
```

kevinpapst (changes requested):
I like the idea! But I'd be happier if we could use my code (becuase I think it is easier to read)

kevinpapst:
I am not going to merge the code as is.
As there was no feedback, I am closing this issue.

PR #2207, opened 2020-12-18
Update timesheet.html.twig - Remove signature

partoneoftwo:
I would like to propose a change to the timesheet template shipped by default in the project.
As a user I never need to have my timesheets signed by myself. I use Kimai because I am able to digitally handle timesheets and they go straight into my PDF generation workflow. They are never printed, and manually signed by me or my workers.

I am proposing this change upstream because I think this is a relic of the past, which needs to go.

By the way, this is also my first Pull Request to this project. Yay.

Cheers,
Thomas

## Description

## Types of changes

## Checklist

### DIFF
--- a/templates/invoice/renderer/timesheet.html.twig
+++ b/templates/invoice/renderer/timesheet.html.twig
@@ -111,12 +111,6 @@
         <div class="table-responsive">
             <table class="table">
                 <tbody>
-                <tr>
-                    <th style="padding-bottom: 60px">{{ 'invoice.signature_user'|trans }}</th>
-                </tr>
-                <tr>
-                    <th>{{ 'invoice.signature_customer'|trans }}</th>
-                </tr>
                 </tbody>
             </table>
         </div>


kevinpapst:
Thanks for your contribution 👍 

> I am proposing this change upstream because I think this is a relic of the past, which needs to go.

Kimai is used by so many different companies with different workflows in all kind of industries.

The reason that multiple of my customers use this workflow and template today should be enough argument, but I can also say that I myself used this template for years. It is a normal workflow (depending on the kind of job/contract your have) that someone has to sign your hours before payment.

Then as direct feedback to the PR, hopefully my reaction does not scare you off for the next time!
- You only deleted the tr's / rows, but left rest of the table
- You didn't remove the unused message from the translation files
- This is a breaking change and not a Bugfix (wrong checkbox) 
- There is no documentation update (wrong checkbox) 

I will not merge this change due to the reason that this template is used and that you can simply create your own invoice template instead: 
copy & paste the existing file into var/invoices/ , apply the change: done!

partoneoftwo:
I see you rejected the PR, which is fine.
I already have a script which resets this template on my server by copying in place my modified version over the upstream original file. But it would be nice to (1) not have to do this (2) for other users not to have this available. 

* There was no checkbox for improvement, so none of your categories fitted. Therefore I ticked the closest match which was bugfix.

I* f it is a breaking change then I might now understand the logic, which is fine.

* What kind of document change was needed? Is the signature part documented anywhere? Do you mean add to changelog? Is this functionally desribed anywhere?

kevinpapst:
> I already have a script which resets this template on my server by copying in place my modified version over the upstream original file.

Which is the wrong workflow: please read the invoice documentation how to create your own invoice template. Then you never have to do that again!
I guess this is the real problem here 😁 

Invoice templates are just "ideas" - most companies customize them in some way.

PR #2210, opened 2020-12-18
Tiny fix: use the correct name for the CustomEvent

strouble:
Just a super tiny change. I think it is not intended to use the "un-split" parameter.


kevinpapst:
Ups, yeah... looks right. I wonder where this is used.
Did you stumble upon that or were you searching for a bug?

strouble:
It is currently not used with more than one event name, so it's currently not a problem, but potentially in the future ...
I am currently looking into a performance improvement, where i could use the custom event and then stumbled across it.

kevinpapst:
Without looking at the code I would have said, that all the API action links are using this code:
https://github.com/kevinpapst/kimai2/blob/master/templates/timesheet/actions.html.twig#L35 (see `data-event` attribute)
I will test and report back

kevinpapst (approved):
Good finding 👍 mercy

PR #2212, opened 2020-12-18
Fix dropdowns on the bottom of a page

strouble:
## Description
If a dropdown menu is opened on the bottom of a page, it "leaves" the page and the user can not see all menu entries (eg. projects index page).
dropdown_fail

Fix/Hack: add `dropup` class to all dropdown menus leaving the page:
dropdown_fix

It is a known/common probllem with bootstrap 3 and can be problematic to all dropdown menus. I've used the `KimaiDatatable.js` to fix this, because the dropdown menus are used there and the fix must be re-applied after loading the page via ajax.

## Types of changes
- [x] Bug fix

## Checklist


assets/js/plugins/KimaiDatatable.js
```diff
@@ -79,11 +82,26 @@ export default class KimaiDatatable extends KimaiPlugin {
...
             },
             error: function(xhr, err) {
                 form.submit();
             }
         });
 
     }
+
+    fixDropdowns() {
+        // HACK: show dropdown menu upwards, if it "leaves" the page
```
kevinpapst, line 95:
```suggestion
        // show dropdown menu upwards, if it is outside the visible viewport
```

assets/js/plugins/KimaiDatatable.js
```diff
@@ -79,11 +82,26 @@ export default class KimaiDatatable extends KimaiPlugin {
...
+        jQuery(this.selector + ' [data-toggle=dropdown]').each(function() {
+            let parent = jQuery(this).parent();
+            let menu = parent.find('.dropdown-menu');
+
+            if (parent && menu) {
+                if ((parent.offset().top + parent.outerHeight() + menu.outerHeight()) > jQuery(document).height()) {
+                    parent.addClass('dropup');
```
kevinpapst, line 102: I'd say you should remove the "dropdown" class in that case. At least it works the same in my test and the example at 
https://getbootstrap.com/docs/3.4/components/#dropdowns-example
looks as if the dropdown class is not required.

assets/js/plugins/KimaiDatatable.js
```diff
@@ -79,11 +82,26 @@ export default class KimaiDatatable extends KimaiPlugin {
...
                 form.submit();
             }
         });
 
     }
+
+    fixDropdowns() {
+        // HACK: show dropdown menu upwards, if it "leaves" the page
+        jQuery(this.selector + ' [data-toggle=dropdown]').each(function() {
+            let parent = jQuery(this).parent();
```
kevinpapst, line 97:
```suggestion
            const parent = jQuery(this).parent();
```

assets/js/plugins/KimaiDatatable.js
```diff
@@ -79,11 +82,26 @@ export default class KimaiDatatable extends KimaiPlugin {
...
+            let menu = parent.find('.dropdown-menu');
```
kevinpapst, line 98:
```suggestion
            const menu = parent.find('.dropdown-menu');
```

assets/js/plugins/KimaiDatatable.js
```diff
@@ -79,11 +82,26 @@ export default class KimaiDatatable extends KimaiPlugin {
...
+
+            if (parent && menu) {
```
kevinpapst, line 100: I am not a fan of these implicit checks... I guess this case will rarely happen, I'd just orefer something like `parent !== undefined`.

kevinpapst (changes requested):
Wohooo, that is so cool! This bug is now open for a very long time and I tried it twice, but couldn't find a solution. My approach with drop was the same, but I failed to calculate which ones get the class.
Cool, very much appreciated 👍 

I just have some minor comments...

assets/js/plugins/KimaiDatatable.js
strouble, line 102:
```suggestion
                    parent.addClass('dropup').removeClass('dropdown');
```

assets/js/plugins/KimaiDatatable.js
strouble: i also want to implicitly check for empty objects/arrays, since this is most probably returned by jquery ...
If you want to make it explicit, then we have to check that too :thinking:

assets/js/plugins/KimaiDatatable.js
strouble: works for me as well

assets/js/plugins/KimaiDatatable.js
kevinpapst: I am a completely javascript noob and never used this kind of checks neither in JS nor PHP, so maybe I am just a bit over complicating here. If this surely works, fine for me as well.

kevinpapst (approved):

kevinpapst (approved):
Thank you very much @strouble 👍

PR #2238, opened 2021-01-03
make sure to generate a valid filename

kevinpapst:
## Description

in export and invoice

## Types of changes

## Checklist


src/Export/Base/PDFRenderer.php
```diff
@@ -119,6 +120,9 @@ public function render(array $timesheets, TimesheetQuery $query): Response
             $filename = 'kimai-export.pdf';
         }
 
+        $filename = new UnicodeString($filename);
+        $filename = $filename->collapseWhitespace()->ascii();
```
kevinpapst, line 124: Use FileHelper:: convertToAsciiFilename()

PR #2246, opened 2021-01-06
Add /api/projects/{id}/statistics endpoint

JEDIBC:
## Description
Add /api/projects/{id}/statistics endpoint to return App\Model\ProjectStatistic of a project

## Types of changes

## Checklist


kevinpapst:
Thank you, I really appreciate that!

I'm mega busy right now, so my feedback might take a bit longer 😞

kevinpapst:
After giving this multiple thoughts and as described in #2346 (which I closed for the same reason) I am going to close this PR.

I really really appreciate the detailed PR and your time that you spent with this.
But  I don't want to include API endpoints for statistics.
At least not for now, because the complete statistic system is not really in a stable state.

This can be included in a plugin, but not in the core right now.

Issue #1676, opened 2020-04-30
Vacation and Sick day request


EliyahuEFGL:
**Is your feature request related to a problem? Please describe.**
I can't find a way to easily track vacation and sick days. 

**Describe the solution you'd like**
A predefined sick day and vacation day. 

**Describe alternatives you've considered**
I set up a sick project with a sick activity and did the same for vacation days. 

If this is covered somewhere in the documentation and I missed it, please point me in that direction.

MarkusWz:
I like the idea on having special activites which are shown and calculated differently.
I would suggest to have vacation, absence other and flexi time.
The reason not to track sick leave but name it absence other or maybe just absence is that according to German rules, the sick leave has to be treated very sensible. Absence can also have other reasons, so it is better to track this way.
While vacation and absence count the "normal" working hours, flexi time would have to be substracted from the hour counting.
So the entry should be shown then as hours in the normal way, but in the database they would be counted as negative hours.
I guess this would be really usefull.

stereo9:
maybe this can be implemented in a transaction based "bank account" type of system

vacation days could be requested and if allowed the transaction will be done
that way the admin / supervisor can add vacation days

the system could then print a balance

| Date of Transaction | Purpose                                                                  | # Leave Days | Effect | Granted on | Granted by |
|---------------------|--------------------------------------------------------------------------|--------------|--------|------------|------------|
| 01.06.2020          | Welcome to the team! Just hired!                                         | 15           | +15    | 25.10.2020 | Admin      |
| 07.12.2020          | Absent                                                                   | 1            | 0      | 07.12.2020 | Admin      |
| 26.12.2020          | Vacation from 26.12.2020 - 29.12.2020                                    | 4            | -4     | 10.11.2020 | Admin      |
| 30.12.2020          | "Betriebsferien"                                                         | 2            | 0      | 01.12.2020 | Admin      |
| 01.01.2021          | You can take only 10 days to the next year. Balance 2020: 11; 10-11 = -1 | 0            | -1     | 01.01.2021 | Admin      |
| 01.01.2021          | Vacation Days 20201                                                      | 30           | +30    | 01.01.2021 | Admin      |
|                     |                                                                          | Balance      | 40     | 01.01.2021 | Admin      |

in an mvp everything could be done manually, thinking about the future all lot of this could be automatized by adding contract details to a profile

kevinpapst:
A very very long time ago this was requested. Now it is here 🎉 

[...]

PR #2249, opened 2021-01-08
Extend category functionality of Timesheets

jrester:
## Description
Now Timesheets can be created with a Category and the category can also be edited.
Timesheets can also be searched based on their Category.

Related to #1676.

## Types of changes

## Checklist


kevinpapst:
As discussed in chat, the plan for the category is a bit larger and the direction was not what I had in mind.

Thank you very much for your time and the contribution 👍 I really appreciate the input and would love to see other contributions/PRs in the future @jrester .

PR #2174, opened 2020-12-08
Allow just one number in the duration hours

ZeiP:
## Description
Currently when adding a new timesheet through the web UI the duration must be given in the format hh:mm. In most other systems you can enter the hours with just one number, ie. 2:30 for 2 hours 30 minutes.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [X] I verified that my code applies to the guidelines (`composer code-check`)
- [N/A] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [X] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

PR #2264, opened 2021-01-15
Improved duration selector

### LINKED ISSUE #1329: Duration mode: enter a value as minutes, not seconds
I've spotted the file src\Utils\Duration.php, which is of great added value, since it converts duration strings into seconds. In other words, entering, lets say *30* in the duration field, wil be accepted and converted in 30 seconds.

Although this might be of practical use for some businesses, precision by seconds is in our time records not used. For our way of time registration it would be much more convenient to convert these loose numbers into minutes, not seconds. Currently, if we want to register, lets say, 45 minutes, we need to type *0:45*. typing just *45* wil only register 45 seconds.

Convert duration strings into minutes by changing the Duration.php file accordingly.

Creating the option in the configurations tab is also fine, off course, if both the options *convert to seconds* and *convert to minutes* need to be possible, depending on the needs of the company in question.

For a test, I manipulated the Duration.php file by changing the fraction from seconds to minutes. I edited
*$hour = floor($seconds / 3600);*  
*$minute = floor(($seconds / 60) % 60);*

into
*$hour = floor($seconds / 60);*  
*$minute = floor(($seconds / 1) % 1);*

This, surprisingly did the trick, duration strings were converted to minutes, but, expectedly, was to simple as a fix, since all the times (also the ones already present) were multiplied by 60, making it look as we had been working sixty times harder :-).

Do you agree that it would be more useful to use conversion to minutes, instead of seconds for duration input?

If so, do you know in what way the Duration.php file needs to be edited in order to have te desired effect? Or do you see other possibilities to add this feature?

Many thanks

Willem

### LINKED ISSUE #1667: When entering time record, the "duration" field it pre-filled with 00:00 and has to be deleted before being filled
When you create a new time record (by pressing on the "+" button on the "my times" screen) the duration field is not empty: it contains the string "00:00". Also, it has focus. If you just type your duration (for example, "1h") the field shows "1h00:00". You have to select all and type over, or press "canc" or "backspace" until it's empty and then enter your duration. 

I expected it to be empty from the beginning, so that when I enter my duration, I don't have to edit the previous content of the field (that is, delete the "00:00" string before typing).

Also, after deleting its value, I see a greyed-out "00:00" that disappears as soon as I type something, so I suppose that the proper behaviour should be that the duration field starts with this greyed "00:00" string shown, and clears itself as soon as I start typing.

I see this issue on Firefox for Linux, but I suppose that it's not browser specific.
Kimai version is 1.8

### LINKED ISSUE #1701: Time selector should match time rounding
Time selector is always minute granularity, even when you configure times should be rounded to e.g. 15 minutes

Time selector should offer :00, :15, :30, :45 according to rounding configuration

make start and end time text fields or a group of buttons

**Additional context**
the minute scroll box is too long and is annoying to get to 45 minutes

### LINKED ISSUE #1702: duration input should be more flexible about values entered
compared to an excel spreadsheet that tracks time durations, the following should work:

1:00 -> 01:00
1 -> 01:00
:30 -> 00:30

don't make me type a leading 0 for no good reason

quick buttons for repeating values like 1hr meetings

**Additional context**

### LINKED ISSUE #1794: Give modals static backdrop
There have been so many times when I've accidentally clicked outside the modal after spending a ton of time typing a description of a time entry, which closes the modal and causes all of that work to be lost. 

Add data-backdrop="static" to twig templates for modals (templates/default/_form_modal.html.twig and possibly also templates/embeds/modal.html.twig)

Possible alternatives would be a confirmation dialog if you close a modal (by clicking close or clicking outside the modal) after making changes. That would require a nested/double modal, which might be more complicated depending how it's implemented.

**Additional context**
I track a lot of information in long time entries. I have no idea if others use Kimai this way, but for me, losing a lot of work because of a careless mouse click is very frustrating.

### LINKED ISSUE #1959: Rounding changes end time on timesheet entry edit
**Describe the bug**
When **Rounding of the duration in minutes** is used with a positive value, the **End** time is set to **Begin** + **Duration** when the entry is edited.

**To Reproduce**
1. Go to **Settings** /  **Time rounding** and set **Rounding of the duration in minutes (0 = deactivated)** to 15 (minutes).
2. Click on **My times** and create a task. 
3. Stop the task after a minute or so. The **duration** is correctly set to **15 minutes** automatically, based on rounding!
4. Edit the task, update the description and save.
5. The End time has been updated to **Begin** + **Duration**.

This is wrong, since the end time should not change when only the description is updated.
This also leads to overlapping tasks, even though the real end time does not overlap with a following task. Rounding should only be used to round the duration.

**Additional context**
 - Kimai version 1.10.1
 - PHP version 7.4.9

### LINKED ISSUE #2044: More mobile friendly hour/minute set value.
Its clumsy to write 00:00. Its more then usual where employees write only ”8” for 8 hours.
The system puts out 00:01.

Idd like a more friendly / failsafe hour input.

**Screenshot**
Request

kevinpapst:
## Description

Can be tested at https://demo-branch.kimai.org (eg. with `susan_super` and `kitten`).

- added dropdown to select durations (using duration rounding to define steps) (Fixes #1667, Fixes #2044) - will be displayed only if duration rounding is set to 5 or more minutes)
- minute dropdown for begin and end date time use rounding rules for value selection (Fixes #1701)
- BC break: treat numeric duration as decimal hours (not as seconds like before)
- duration field allows to enter one decimal instead of forcing two decimals for hours (Closes #2174)
- duration field allows to enter several formats: 8h30m = 08:30 = 8:30 = 8,5 (Fixes #1329, Fixes #1702)
- Bug: do not update end field upon editing an existing timesheet (Fixes #1959)
- do not close modal if form is dirty (Fixes #1794)
- added shadow and rounded corners to modals
- make minute dropdowns configurable (by default uses rounding rules)

**TODO**
- [x] tests
- [ ] documentation

## Types of changes

## Checklist


tests/API/TimesheetControllerTest.php
```diff
@@ -419,8 +419,8 @@ public function testPostActionForDifferentUser()
             'activity' => 1,
             'project' => 1,
             'user' => $user->getId(),
-            'begin' => ($dateTime->createDateTime('- 16 hours'))->format('Y-m-d H:m:0'),
-            'end' => ($dateTime->createDateTime())->format('Y-m-d H:m:0'),
+            'begin' => ($dateTime->createDateTime('- 16 hours'))->format('Y-m-d H:m'),
```
kevinpapst, line 422: ????

assets/js/plugins/KimaiAjaxModalForm.js
```diff
@@ -28,9 +28,21 @@ export default class KimaiAjaxModalForm extends KimaiReducedClickHandler {
 
     init() {
         const self = this;
+        this.isDirty = false;
 
         this.modal = jQuery('#remote_form_modal');
-        this.modal.on('hide.bs.modal', function () {
+        this.modal.on('hide.bs.modal', function (e) {
+            let msg = self.getContainer().getTranslation().get('modal.dirty');
+            console.log(self.isDirty);
```
kevinpapst, line 36: remove

assets/js/plugins/KimaiAjaxModalForm.js
```diff
@@ -139,6 +160,12 @@ export default class KimaiAjaxModalForm extends KimaiReducedClickHandler {
         // the new form that was loaded via ajax
         form = jQuery(formIdentifier);
 
+        this._isDirtyHandler = function(e) {
+            console.log(e);
```
kevinpapst, line 164: remove

assets/js/plugins/KimaiAjaxModalForm.js
```diff
@@ -44,7 +56,12 @@ export default class KimaiAjaxModalForm extends KimaiReducedClickHandler {
         });
         this.modal.on('shown.bs.modal', function () {
             // workaround for autofocus attribute, as the modal "steals" it
-            jQuery(self._getFormIdentifier()).find('input[type=text],textarea,select').filter(':not("[data-datetimepicker=on]")').filter(':visible:first').focus().delay(1000).focus();
+            let formAutofocus = jQuery(self._getFormIdentifier()).find('[autofocus]');
+            if (formAutofocus.length < 1) {
+                formAutofocus = jQuery(self._getFormIdentifier()).find('input[type=text],textarea,select');
+            }
+            formAutofocus.filter(':not("[data-datetimepicker=on]")').filter(':visible:first').focus().delay(1000).focus();
+            //jQuery(self._getFormIdentifier()).find('input[type=text],textarea,select').filter(':not("[data-datetimepicker=on]")').filter(':visible:first').focus().delay(1000).focus();
```
kevinpapst, line 64: remove

src/Form/TimesheetEditForm.php
```diff
@@ -159,30 +147,58 @@ protected function showCustomer(array $options, bool $isNew, int $customerCount)
...
+                'placeholder' => '0:00',
+            ],
+        ];
+
+        if ($autofocus) {
+            $durationOptions['attr']['autofocus'] = 'autofocus';
+        }
+
+        $duration = $options['duration_minutes'];
+        if ($duration !== null && (int) $duration >= 5) {
```
kevinpapst, line 188: ?

PR #2291, opened 2021-01-22
limit amount of items in calendar drag and drop boxes

kevinpapst:
## Description

- limit amount of items in calendar drag and drop boxes

## Types of changes

## Checklist


cariforef-normandie:
Hi again Kevin,

How to help ? I'm not familiar with Github processes, but I think the code here is good. Ok I do not help so far but it's a beginning ;-)

kevinpapst:
Are you familiar with Git / Coding? If so, you could pull the code and test it.
I can't remember where I left off and what is missing

Issue #2323, opened 2021-02-07
allow LDAP jpegPhoto via data:image/jpeg,base64, url


bash-stack:
**Is your feature request related to a problem? Please describe.**
currently most ldap servers do not spit a public avatar URL, 
but uers have `jpegPhoto` , 
trying to "errornoulsy" pull them into the avatar-adressfield ( config excerpt)
```
                          ## FAILS , WOULD NEED URL FROM LDAP OR BASE64 capabilites 
                           - { ldap_attr: jpegphoto, user_method: setAvatar }
```
not only misses the needed `data:image/jpeg,base64` string a browser would accept
also the string is somehow escaped to a billion escape sequences
 but:
 
```
 An exception occurred while executing 'INSERT INTO kimai2_users (username, username_canonical, email, email_canonical, enabled, salt, password, last_login, confirmation_token, password_requested_at, roles, alias, registration_date, title, avatar, api_token, auth) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params ["testnutzer", "testnutzer", "a@b.de,"a@b.de", 1, null, "", "2021-02-07 17:27:10", null, null, "a:0:{}", "Test Nutzer", "2021-02-07 17:27:09", null, "\xff\xd8\xff\ 
 .. TONS OF ESCAPES...
 ", null, "ldap"]:

SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xFF\xD8\xFF\xE0\x00\x10...' for column `kimai2`.`kimai2_users`.`avatar` at row 1
```

**Describe the solution you'd like**
* SOLUTION A: kimai stores a copy of `jpegphoto` in `public/` folder ( and maybe resync every X time period)
* SOLUTION B: kimai takes the `jpegPhoto` content transformed as ` data:image/jpeg,base64,$jpegPhoto`

kevinpapst:
#1405

bash-stack:
( maybe just add `setAvatarBase64() `as a function that sets the url to `data:image/jpeg,base64,$jpegPhoto` 
DANGER:  this could kill browsers when appearing in html forms ( like user edit page), inline-stuff in forms pushs e.g. firefox to 100% cpu load   )

Issue #1405, opened 2020-01-22
LDAP jpegphoto avatars support


vvirehead:
**Is your feature request related to a problem? Please describe.**
Built-in LDAP avatars are not supported.

**Describe the solution you'd like**
I would like to have jpegphoto LDAP attribute support for avatar purposes.

**Describe alternatives you've considered**
Alternative avatar system support (ie. gravatar, based on user email)

**Additional context**
Right now data from jpegphoto is loaded to web interface but it's not displayed properly.

local.yaml

```yaml
kimai:
    ldap:
        user:
            attributes:
                - { ldap_attr: jpegphoto, user_method: setAvatar }
```

kevinpapst:
As a http(s) URL should work, I guess this is a base64 encoded image?

g-mike:
Yes, I believe at least in OpenLDAP this is encoded as base64 string, the attribute is defined as JPEG, 
Syntax OID is 1.3.6.1.4.1.1466.115.121.1.28 (this is the only spec I could find)

In other services in web browser image starts as:
```html
<input type="hidden" name="new_values[jpegphoto][0]" value="/9j/4AA...Y//9k=">
```
and is later displayed as:

```html
<img src="view_jpeg_photo.php?dn=XXX;index=0&amp;attr=jpegphoto&amp;location=session" alt="Photo" width="300" height="300">
```

bash-stack:
( maybe just add `setAvatarBase64() `as a function that sets the url to `data:image/jpeg,base64,$jpegPhoto` 
DANGER:  this could kill browsers when appearing in html forms ( like user edit page), inline-stuff in forms pushs e.g. firefox to 100% cpu load   )

kevinpapst:
some logic should dump the avatars into a public accessible url.
then kimai can grab that url.
not building support for base64 images into kimai directly.

PR #2325, opened 2021-02-07
add setAvatarBase64

bash-stack:
Fix for #2323  and #1405 

Added 
   ` public function setAvatarBase64(?string $avatar)`
that should sync base64 jpegPhoto from ldap indo kimai

## Description

## Types of changes
- [ x] New feature (non-breaking change which adds functionality)

## Checklist
- [ x ] I verified that my code applies to the guidelines (`composer code-check`)
I DID NOT update the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [ x ] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -268,7 +268,7 @@ class User extends BaseUser implements UserInterface
     public function setAvatarBase64(?string $avatar): User
     {
         $prefix = 'data:image/jpeg,base64,';
-        $this->avatar = $prefix.$avatar;
+        $this->avatar = $prefix . $avatar;
 
         return $this;
     }
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -267,7 +267,8 @@ class User extends BaseUser implements UserInterface
 
     public function setAvatarBase64(?string $avatar): User
     {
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks for the input! The avatar field is really only meant to store an (external) URL => varchar(255)
The image needs to be saved as file and then the avatar field needs to be updated to point to that file.

bash-stack:
> Thanks for the input! The avatar field is really only meant to store an (external) URL => varchar(255)
> The image needs to be saved as file and then the avatar field needs to be updated to point to that file.
> 
> Edit: I am not a fan of base64 in that case. Too much traffic for every page load, which could otherwise simply be cached by the browser.

as written under the feature req's: inline shizzle in browser has to be masked ( i've seen a crm named like a capital that kills browsers when editing inline mailtemplates , firefox,chrome(/native webkit)-> all browsers heavily struggle )

kevinpapst:
Closing here. Importing LDAP avatars seems to be a missing feature, but it needs to be handled in a different way.

PR #2510, opened 2021-04-18
Make initial value for datepicker in exports configurable in user prefs

mluex:
## Description
If you don't invoice monthly, but e.g. project based or yearly, you often find yourself increasing or emptying the time range field in the export and invoice forms, e.g. setting 01.01.2021 as start to filter all non-exported entries.

This PR introduces a new user preference for this, which can be used to set the default value for the Time Range Datepicker. Default is still the current month, so this is not a breaking change.

However, the current year is now also available for selection. The default date picker would then not be _2021-04-01 - 2021-04-30_, but _2021-01-01 - 2021-12-31_.

The current decade can also be selected. Probably this is not interesting for many users though.

Simply a small convenience thing to avoid a few clicks ;-)

German translations included.

And of course: thanks a lot for this awesome tool! ;-)

## Types of changes

## Checklist


kevinpapst:
Thanks for the PR 👍 

Did you write that based on the latest master?
I think this should not be a new setting, but better be incorporated into the search bookmarks. 

We need to support relative date times in the toolbar daterange filters.
So instead of saving `2021-01-01 - 2021-12-31` from:
Bildschirmfoto 2021-04-18 um 18 52 36
we should support strings like `this year`/`this month`/`this week` as selector in the `Time range` field.

I assume this is "only" an updated version of the existing data transformer:
https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/DateRangeType.php#L123

mluex:
Yes, it is based on the latest master. But I've indeed overlooked the new Search Bookmarks feature, sorry ... cool thing!

I totally agree with you that this functionality can be better integrated in the bookmarks feature, good idea. Adding that to the DateRangeType transformer shouldn't be complicated.

On the other hand, the integration with the Datepicker could be a bit tricky. It would have to be able to interpret `this year` & co. as well, so that the correct dates are selected when opening the Datepicker (if `value="this year"` is set through the bookmark).
Furthermore, the string expressions should be somehow selectable in some kind of dropdown, ideally in the Datepicker itself.

If you want, you can close this PR for now. I'll open a new one as soon as I've had a chance to look at it.

kevinpapst:
Lets keep this PR open for now, at least as idea.

What I think should be done:
- keep the dates as view data, but store the string
- convert internally to the string to store it 

So "this year" is the model data, while the "2021-01-01 - 2021-12-31" is the view data.
That would mean we have a list of pre-made dates that need to be compared inside the transformer.

From a first glance, I'd say we should offer the same strings than the DateTimePicker offers as pre-made choices:
Bildschirmfoto 2021-04-19 um 09 08 20

kevinpapst:
Closing for now, as the concept does not match the search bookmark idea.
A new PR adding support date ranges like `this year` would be great!

PR #2561, opened 2021-05-09
added API route to fetch infos about installed plugins

kevinpapst:
## Description

- improved API documentation
- added API route to fetch infos about installed plugins

ping @AlexandrePTJ

## Types of changes

## Checklist


AlexandrePTJ:
This is perfect for me. I will be able to add plugins support depending on this result.

PR #2580, opened 2021-05-21
feat: adds support for setting description by http params

tobybatch:
## Description

## Types of changes

## Checklist


tobybatch:
@kevinpapst I'd like to add a test for this too, are there tests for the http param stuff? where? and I'll add one for this.

src/Controller/TimesheetAbstractController.php
```diff
@@ -183,6 +183,11 @@ protected function create(Request $request, string $renderTemplate, ProjectRepos
             $entry->setActivity($activity);
         }
 
+        if ($request->query->get('description')) {
+            $description = $activityRepository->find($request->query->get('description'));
```
kevinpapst, line 187: Copy & Paste

kevinpapst (changes requested):

kevinpapst:
https://github.com/kevinpapst/kimai2/blob/master/tests/Controller/TimesheetControllerTest.php#L279

tobybatch:
And that's why I wanted the pointer to the tests :)

Thanks Kev.

tobybatch:
@kevinpapst I'm getting

```
failed to open stream: Too many open files
```

While running tests, my system limit is high but php seems to be choking it.

kevinpapst:
Either:
- Push and let GitHub actions do the work
- or simply execute only one test with `vendor/bin/phpunit tests/Controller/TimesheetControllerTest.php`

As I dropped SQLite support, please `cp phpunit.xml.dist phpunit.dist` and change to true `<env name="BOOTSTRAP_RESET_DATABASE" value="true"/>`.
You need to create a db and user first, commands can be copy & pasted from here: https://github.com/kevinpapst/kimai2/blob/master/phpunit.xml.dist#L19

tobybatch:
Yeah, I'd got the new SQL test DB settings, that was fine. I was just trying to run the whole test suite.

I'll just run my test. TYhanks

tobybatch:
@kevinpapst Tests added, changes made, linting linted. I've switch it from a draft to a real PR.

kevinpapst:
What do you need that for exactly?
Did you try to inject HTML for a XSS attack?

tobybatch:
Added a test to check html injection can't be done. 

I'm moving my browser extension (https://github.com/tobybatch/kimai-browser-plugin) to use tags rather than a custom method for storing relationship between projects/activities and github tickets. I don't think XSS is a vulnerability as a user must be logged in and the description text is correctly decoded.

kevinpapst (approved):
I was a bit worried about possible security issues, but there is no real difference between get and post params, so good to merge. Thanks @tobybatch 👍

PR #2618, opened 2021-06-18
Markdown in one-liners and team timesheets

synaestic:
## Description
`MarkdownExtension::commentOneLiner()` added in af9dea9226c49517f529920c4063098d9a6405cd was implemented without markdown rending. This PR adds the functionality and ensures inline rendering with use of [Parsedown::line()](https://github.com/erusev/parsedown/blob/6598f3860c2698fe2f0f1bc98212fc01d0a1893c/Parsedown.php#L1130).

Also including a fix for team timesheet descriptions. But maybe i just miss the point if it was disabled on purpose.

## Types of changes

## Checklist

### DIFF
--- a/src/Twig/Runtime/MarkdownExtension.php
+++ b/src/Twig/Runtime/MarkdownExtension.php
@@ -89,6 +89,10 @@ final class MarkdownExtension implements RuntimeExtensionInterface
             $addHellip = true;
         }
 
+        if ($this->isMarkdownEnabled()) {
+            $content = $this->markdown->inline($content, false);
+        }
+
         $content = explode(PHP_EOL, $content);
         $result = $content[0];
 
--- a/src/Utils/Markdown.php
+++ b/src/Utils/Markdown.php
@@ -23,7 +23,6 @@ final class Markdown
     {
         $this->parser = new ParsedownExtension();
         $this->parser->setUrlsLinked(true);
-        $this->parser->setBreaksEnabled(true);
     }
 
     /**
@@ -34,7 +33,21 @@ final class Markdown
     public function toHtml(string $text, bool $safe = true): string
     {
         $this->parser->setSafeMode($safe);
+        $this->parser->setBreaksEnabled(true);
 
         return $this->parser->text($text);
+    }
+
+    /**
+     * @param string $text
+     * @param bool $safe
+     * @return string
+     */
+    public function inline(string $text, bool $safe = true): string
+    {
+        $this->parser->setSafeMode($safe);
+        $this->parser->setBreaksEnabled(false);
+
+        return $this->parser->line($text);
     }
 }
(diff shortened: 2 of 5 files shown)


kevinpapst (changes requested):
Hi @synaestic - I appreciate any contribution: thanks you 👍 

Unfortunately none of these changes can be added.
The `commentOneLiner` and `comment1line` twig filter are made to take a string and return a result that is safe to show in a table and rely that it is only one line. Even though you `disableBeaks()` using markdown could result in headlines, lists and other line breaking elements.
So it would break the purpose of it. I do understand that there is always some visual distraction if you use markdown extensively and then render these description fields in other places without markdown. Maybe we could use markdown and then strip_tags  to get rid of the markdown elements?

Team timesheets does not render markdown on purpose. If you think this should be changed, then it should be a configurable behavior.

synaestic:
Hi @kevinpapst,

Markdown rendering is limited to the defined [$InlineTypes](https://github.com/erusev/parsedown/blob/cb17b6477dfff935958ba01325f2e8a2bfa6dab3/Parsedown.php#L1031). We could reduce the permitted elements further (e.g. removing images). The main use case for me are links to external projects pages or issues in the first line of an entry.
To ensure consistent behavior i think it would make sense to render a user / personal timesheet list description inline too.
Currently the `timesheet.markdown_content` setting also permits [markdown for customer, project and activity comments #807](https://github.com/kevinpapst/kimai2/pull/807). I think so too. It would make sense to ensure someone with administrative rights explicitly enables markdown on all team elements.

While strip_tags would improve the readability of some elements i would prefer to signalize markdown support. Maybe a link to the supported elements in addition to a warning if markdown is disabled for team elements.

kevinpapst:
I don't mind changing this, I am just busy with other things currently.

Markdown support for timesheets should probably not influence the support for markdown in customer/projects/activities. 
If I remember correctly that was just a quickfix for the case when users are allowed eg. to create activities.

kevinpapst:
Hey @synaestic - would you jump and jump in again. Seems you have some good points here and I'd like to address them. Please let me know when we could work on this PR

synaestic:
I've seen your response and want to continue to work on this PR but time was rare and is still rare this week. I should have some time next week and can include all the required changes. It would be great if we can make a list of the things which need to be  implemented in this PR.

kevinpapst:
Awesome, lets do it when you are available. I'd prefer having a quick chat and then start when we have the same idea of the changes.

synaestic:
I have some time now and got a quick look into the issue. To sum up:

So, we add an additional configuration for customer/projects/activities and one for team timesheets. Markdown elements in tables will get parsed through Parsedown::line (except images). Should  team timesheets comments in tables also get rendered in one line? Should document the supported markdown elements (if i remember correctly `<h1>` have already been removed) and add a link near the description / comment fields? What do you think?

synaestic:
@kevinpapst I have (again) some time now. Please let me know what do you think.

kevinpapst:
I think this should be re-started for the next major release: https://github.com/kevinpapst/kimai2/pull/2902

I will not add it to the current 1.x release and also the next release will have much better CSS support, so we might be able to fix some issues in the frontend as well.

kevinpapst:
Needs re-write for https://github.com/kevinpapst/kimai2/pull/2902

PR #2637, opened 2021-06-30
Fix import issue where only one customer is created during import

kitzler-walli:
## Description
This fixes the customer creation during timesheet import.

## Types of changes

## Checklist


kevinpapst (changes requested):
Can you please remove the lines instead of commenting them. This looks like a test now...

Ok, after reading the code you might be right. 1,5 years old code is like reading code from a stranger 😁 But I believe this if has something to do with the `--customer` option. Did you use that? 

I will have to test it (the demo CSV would be really helpful) first.

kitzler-walli:
Just by reading the code it's clear that after the first run or after the first customer that had to be created the check for `$this->customerFallback` will no longer return null and therefore always return the same customer.
I didn't use the --customer option.

kevinpapst:
Yes, agree. But what happens now for new projects? Will there always be a new customer for a new project? 
AFAIR the idea was that imported projects get assigned to the same "fallback" customer and then have to be sorted out manually. I remember using the `--customer` option normally when doing imports the last times.

kitzler-walli:
The customer is identified by it's name - so every new project gets the same customer as long as it has the same customer name. If the customer name is different a new customer is created or retrieved if it already exists.
To be honest I didn't check the customer option - probably there is some extra work necessary for this but yesterday I wanted to move 15.000 timesheets with new projects and customers so I didn't check every option.

kevinpapst:
Ah see, I overlooked the support for the support if the `Customer` name. Yeah, probably too long ago that I wrote the code.
Okay, I will do some testing and come back with an update somewhen in the future. 
Thanks @kitzler-walli for your support until here 👍

kevinpapst:
PR can be closed, replaced by multiple other PRs:

- #2659 for the calendar drag&drop boxes layout
- #2291 for the dog&drop item amount
- #2660 for the importer

Thanks for the inspiration, very much appreciated 👍 🎉

PR #2698, opened 2021-08-06
Add a calendar configuration event

Jelle-S:
This way plugins can define subscribers in which they can alter the
configuration of the calendar. Coming from
https://github.com/digipolisgent/kimai_plugin_personal-calendar-pref/issues/1

## Description
Adds a calendar configuration event that is dispatched before rendering the calendar.

## Types of changes

## Checklist

### DIFF
--- a/src/Event/CalendarConfigurationEvent.php
+++ b/src/Event/CalendarConfigurationEvent.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+class CalendarConfigurationEvent extends Event
+{
+    /**
+     * @var array
+     */
+    private $configuration;
+
+    public function __construct(array $configuration)
+    {
+        $this->configuration = $configuration;
+    }
+
+    public function getConfiguration(): array
+    {
+        return $this->configuration;
+    }
+
+    public function setConfiguration(array $configuration)
+    {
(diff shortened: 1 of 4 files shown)


src/Event/CalendarConfigurationEvent.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function getConfiguration(): array
+    {
+        return $this->configuration;
+    }
+
+    public function setConfiguration(array $configuration)
+    {
+        $this->configuration = $configuration;
```
kevinpapst, line 33: Can you please use something like this:
```php
foreach ($configuration as $key => $value) {
  if (array_key_exists($key, $this->configuration) {
    $this->configuration[$key] = $value;
  }
}
```
Otherwise the plugins need to be clever enough to set all keys which might be added in the future.

src/Event/CalendarConfigurationEvent.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+namespace App\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+class CalendarConfigurationEvent extends Event
+{
+    /**
+     * @var array
```
kevinpapst, line 17:
```
     * @var array<string, string|int|bool|array>
```

kevinpapst (changes requested):
Thanks @Jelle-S looks good so far.
Two minor comments and could you please add a small test to verify the event object like this: https://github.com/kevinpapst/kimai2/blob/master/tests/Event/CalendarGoogleSourceEventTest.php

Jelle-S:
Thank you for your quick reply. Your review remarks were addressed.

kevinpapst (approved):
LGTM 👍 thanks for sharing!

PR #2747, opened 2021-09-03
added ProjectConstraint to add dynamic project validation

pkaltenboeck:
## Description
This PR allows Plugin-Developer to add custom ProjectConstraints by extending the newly created ProjectConstraint (like it's already possible for Timesheet Validation).

## Types of changes

## Checklist

## Background
I had the case where I had to make sure that the project name is unique overall. I wanted to solve the problem inside a plugin which isn't possible till now.


kevinpapst:
Thanks, I will have a lok. after my holiday.
Can you add tests for your changes? There is a test for the timesheet validator which you can use as inspiration.

kevinpapst (changes requested):
Thank you 👍 but as commented earlier, needs tests.

kevinpapst:
Thanks @pkaltenboeck 🚀

kevinpapst:
Can you share your use-case for that @pkaltenboeck ?

I am currently working on 3.0 and would like to remove this piece if code, as the new Symfony version does not allow to use `AutoconfigureTag` attribute on Constraint classes, see https://github.com/symfony/symfony/issues/62883

In my opinion you should be able to use a FormExtension like that:

```php
final class ProjectEditFormExtension extends AbstractTypeExtension
{
    public static function getExtendedTypes(): iterable
    {
        return [
            ProjectEditForm::class,
        ];
    }

    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'constraints' => [
                new MyCustomConstraint()
            ],
        ]);
    }
}
```

With `Timesheet` there is much more stuff going on at many places, which Projects likely do not need.

Issue #2748, opened 2021-09-05
Whitespace on line endings in created docx invoices


Input-BDF:
Using .docx invoice templates adds whitespaces on lineendings when using multiline values for customer address, contact, payment_details, ... Whitespace is added from second until penultimate line.

I am using templatevariables like
```
${customer.address}
${template.contact}
${template.payment_details}
```
 as follows:
```
Line1
Line2
Line3
```
Within document.xml from unpacked .docx the result is:
```
<w:t>Line</w:t><w:br/>
<w:t xml:space="preserve">Line2 </w:t><w:br/>
<w:t>Line3</w:t></w:r>

```
This is only noticeable when using right aligned text and happens on all textarea input fields using multiple lines. Other template formats like pdf or html are not affected 

 - Kimai version: 1.14.3 (already happend on 1.12)
 - PHP version: 7.3

kevinpapst:
And your problem is the extra space in `Line 2 `?

Input-BDF:
Exactly. First and last line is okay and without whitespace. Every line between gets an extra whitespace.

kevinpapst:
See https://github.com/kevinpapst/kimai2/blob/master/src/Invoice/Renderer/DocxRenderer.php#L38

Are you sure the value is not in your text box?

Input-BDF:
Double checked.
Maybe I should notice that I'm using Libre Office. But i rechecked with Word 365. Word is showing formating correctly but Whitespaces are also present.

I renamed .docx to zip and examined document.xml
Source there is showing an CR (marked as [CR] in the code below) where imho it should not be:

```
<w:t>ADLine1[CR]
</w:t><w:br /><w:t xml:space="preserve">ADLine2[CR]
</w:t><w:br /><w:t xml:space="preserve">ADLine3[CR]
</w:t><w:br /><w:t xml:space="preserve">ADLine4</w:t>
```
I'm not shure, but document.xml should only have <?xml... and <w:document... as Lines. Removing the CRs from document.xml repacking, renaming etc. And everything is as wanted.

It seems to by caused by using PHP_EOL since it is imho plattform specific on line endings: LF vs CRLF.

Input-BDF:
LOL. Found this: https://eidson.info/post/php_eol_is_broken

My php knowledge is out of date but i did a quick and dirty-fix via ssh.
```
$replace = preg_replace('/\n|\r\n?/', '</w:t><w:br /><w:t xml:space="preserve">', $replace);
```
[...]

PR #2756, opened 2021-09-10
prevent whitespace on line endings 

Input-BDF:
Using PHP_EOL is system specific. This adds unwanted whitespace in rendering docx-templates. See #2748


kevinpapst:
Thanks 👍 I'll have a closer look after my holidays. I believe there is a similar code snippet somewhere else, which could benefit from this adjustment as well.

I'll get in touch in about a week.

kevinpapst (approved):
One of these bugs where the actual fix is small, but the time to really find and discuss it so much more.

Thank you very much for investigating that issue 👍

PR #2778, opened 2021-09-19
Add option to copy the billable status of the timesheet

### LINKED ISSUE #2777: Get same billable status when starting again a time entry
When I restart an activity, the new timesheet entry always check the billable option.

I want the new entry to have the same billable option as the previous one.

Guillaume-Duc-95:
## Description
I added the possibility to also copy the field "billable" of a timesheet. This fixes #2777.

## Types of changes

## Checklist

### DIFF
--- a/src/API/TimesheetController.php
+++ b/src/API/TimesheetController.php
@@ -590,7 +590,7 @@ class TimesheetController extends BaseApiController
      *      required=true,
      * )
      *
-     * @Rest\RequestParam(name="copy", requirements="all|tags|rates|meta|description", strict=true, nullable=true, description="Whether data should be copied to the new entry. Allowed values: all, tags, rates, description, meta (default: nothing is copied)")
+     * @Rest\RequestParam(name="copy", requirements="all|tags|rates|meta|description|billable", strict=true, nullable=true, description="Whether data should be copied to the new entry. Allowed values: all, tags, rates, description, billable, meta (default: nothing is copied)")
      * @Rest\RequestParam(name="begin", requirements=@Constraints\DateTime(format="Y-m-d\TH:i:s"), strict=true, nullable=true, description="Changes the restart date to the given one (default: now)")
      *
      * @ApiSecurity(name="apiUser")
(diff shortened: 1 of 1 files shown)


src/API/TimesheetController.php
```diff
@@ -590,7 +590,7 @@ public function stopAction(Timesheet $id): Response
      *      required=true,
      * )
      *
-     * @Rest\RequestParam(name="copy", requirements="all|tags|rates|meta|description", strict=true, nullable=true, description="Whether data should be copied to the new entry. Allowed values: all, tags, rates, description, meta (default: nothing is copied)")
+     * @Rest\RequestParam(name="copy", requirements="all|tags|rates|meta|description|billable", strict=true, nullable=true, description="Whether data should be copied to the new entry. Allowed values: all, tags, rates, description, billable, meta (default: nothing is copied)")
```
kevinpapst, line 593:
```suggestion
     * @Rest\RequestParam(name="copy", requirements="all|tags|rates|meta|description", strict=true, nullable=true, description="Deprecated since 1.15. All data is copied always.")
```

kevinpapst (changes requested):
Good idea, thanks for the contribution 👍 
I'd prefer to simplify the code (don't know why there are so many options).

Can you add my suggestion and replace the entire if with:
```php
        if (null !== ($copy = $paramFetcher->get('copy'))) {
            if ($copy !== 'all') {
                @trigger_error('Setting the "copy" attribute in "restart timesheet" API to something else then "all" is deprecated', E_USER_DEPRECATED);
            }

            $copyTimesheet
                ->setHourlyRate($timesheet->getHourlyRate())
                ->setFixedRate($timesheet->getFixedRate())
                ->setDescription($timesheet->getDescription())
                ->setBillable($timesheet->isBillable())
            ;

            foreach ($timesheet->getTags() as $tag) {
                $copyTimesheet->addTag($tag);
            }

            foreach ($timesheet->getMetaFields() as $metaField) {
                $metaNew = clone $metaField;
                $copyTimesheet->setMetaField($metaNew);
            }
        }
```

kevinpapst (changes requested):
Thanks for the update, but unfortunately this is now a BC break.
By replacing `requirements="all|tags|rates|meta|description"` you broke the use case for all API users which are using one of the other allowed values. The framework validates the passed value and if someone now submits `copy:meta` it will fail with a validation error.
Can you please check my suggestion from before.

Guillaume-Duc-95:
Aw yes, you are right! It should be better now

kevinpapst (approved):

PR #2828, opened 2021-10-08
Extend orderNumber string to 50 characters (#2824)

### LINKED ISSUE #2824: Extend length of project orderNumber string
The project orderNumber string is limited to 20 characters. Sometimes costumers have complicated ways to create their numbers and we have hit the limit several times.

Can we extend the limit to some generic value e.g. 256?
I think I could try to do that with some guidance.

As far as I understand it would mean to:
- adapt 2 lines here: https://github.com/kevinpapst/kimai2/blob/b7cf9cd776593cb8b05c7d70109d04d4aaa4f86c/src/Entity/Project.php#L115
- define migration to enlarge fields in database

We have also purchased the MetaFieldsBundle and could create a custom field. However, I think our proposal could save some trouble in the future.

iusgit:
## Description
The database field containing the project order number is extended allow up to 50 characters. Closes #2824.

## Types of changes

## Checklist
- [x] I updated the documentation | nothing to  do


iusgit:
@kevinpapst  So, to do the testing (composer code-check) I need a running Kimai instance, right? Or would the workflow do it, too? Little guidance would be appreciated...

kevinpapst (approved):
LGTM 👍

kevinpapst:
Thanks @iusgit 🎉

PR #2865, opened 2021-10-18
upgraded composer packages, including DBAL 3

kevinpapst:
## Description


kevinpapst:
merged into #2902

PR #2934, opened 2021-11-15
improve calendar in duration mode (#968)

### LINKED ISSUE #968: Allow user changing the date for "duration with fixed start time" tracking mode
When using the 'duration_only' mode it is allways necessary to enter a time for the start of the entry. As my customers are not really interesseted in the exact time when the work was done, I really appreciate the new 'duration_fixed_start' mode. This really saves me some needless clicks and scrolls. So I configured 'default_begin: 08:00' in my local.yaml to avoid overlapping with the next day, for example when entering a 8 hour workday at 22:00.

So now comes the feature request. Unfortunately it is not possible to change the date anymore. Therefore it's only possible to track times for the same day. When I travel for example from tuesday to thursday, I usually track my times on friday for the last 3 days. With the current configuration it's not possible to track those times comfortable in 'duration_fixed_start' mode. Is it possible to add a new permission for users to select a different date while in this tracking mode? I think tracking times with the calendar view would definitely benefit from this additional configuration feature. At the moment when clicking on different dates in the calendar view, the times is always tracked for today.

delvalt:
## Description

Fix #968

Upon creating a timesheet from the calendar monthly view, set the timesheet time to 'default_begin' instead of current time in both duration timetracking modes.

In the user calendar view, replace the event start time by the duration in the event title.

## Types of changes

## Checklist

### DIFF
--- a/src/Controller/CalendarController.php
+++ b/src/Controller/CalendarController.php
@@ -62,6 +62,7 @@ class CalendarController extends AbstractController
             'can_edit_begin' => $mode->canEditBegin(),
             'can_edit_end' => $mode->canEditBegin(),
             'can_edit_duration' => $mode->canEditDuration(),
+            'is_duration_mode' => $mode->isDurationMode(),
         ]);
     }
 }
--- a/src/Timesheet/TrackingMode/AbstractTrackingMode.php
+++ b/src/Timesheet/TrackingMode/AbstractTrackingMode.php
@@ -34,7 +34,15 @@ abstract class AbstractTrackingMode implements TrackingModeInterface
             return;
         }
 
-        $start = DateTime::createFromFormat('Y-m-d', $start, $this->getTimezone($entry));
+        $startFormat = 'Y-m-d';
+
+        // keep default begin time in duration timetracking modes
+        if ($this->isDurationMode()) {
+            $startFormat .= ' H:i:s';
+            $start .= ' ' . $entry->getBegin()->format('H:i:s');
+        }
+
+        $start = DateTime::createFromFormat($startFormat, $start, $this->getTimezone($entry));
         if (false === $start) {
             return;
         }
--- a/src/Timesheet/TrackingMode/DefaultMode.php
+++ b/src/Timesheet/TrackingMode/DefaultMode.php
@@ -56,6 +56,11 @@ final class DefaultMode extends AbstractTrackingMode
         return true;
     }
 
+    public function isDurationMode(): bool
+    {
+        return false;
+    }
+
     public function create(Timesheet $timesheet, ?Request $request = null): void
     {
         parent::create($timesheet, $request);
--- a/src/Timesheet/TrackingMode/DurationFixedBeginMode.php
+++ b/src/Timesheet/TrackingMode/DurationFixedBeginMode.php
@@ -48,6 +48,11 @@ final class DurationFixedBeginMode implements TrackingModeInterface
         return false;
     }
 
+    public function isDurationMode(): bool
+    {
+        return true;
+    }
+
     public function create(Timesheet $timesheet, ?Request $request = null): void
     {
         if (null === $timesheet->getBegin()) {
--- a/src/Timesheet/TrackingMode/DurationOnlyMode.php
+++ b/src/Timesheet/TrackingMode/DurationOnlyMode.php
@@ -56,6 +56,11 @@ final class DurationOnlyMode extends AbstractTrackingMode
         return false;
     }
 
+    public function isDurationMode(): bool
+    {
+        return true;
+    }
+
     public function create(Timesheet $timesheet, ?Request $request = null): void
     {
         if (null === $timesheet->getBegin()) {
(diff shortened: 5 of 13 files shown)


kevinpapst:
Hi @delvalt
thanks for taking the time to put a PR together 👍 and sorry for the delay with my answer.

The tracking modes all have a `create` method, which receive new timesheets. 
This method defines the behavior for that mode, which should not be differ between calendar, API or regular UI.

Please have a look here: 
https://github.com/kevinpapst/kimai2/blob/master/src/Timesheet/TrackingMode/PunchInOutMode.php#L40
https://github.com/kevinpapst/kimai2/blob/master/src/Timesheet/TrackingMode/DurationOnlyMode.php#L59
https://github.com/kevinpapst/kimai2/blob/master/src/Timesheet/TrackingMode/DurationFixedBeginMode.php#L51

From a first glance I'd say you alter the behavior only in the calendar, right?

delvalt:
Hi @kevinpapst , thanks for your response!

From what I recall, yes, this only alters the calendar:
Upon creating a timesheet by clicking a day in the calendar's monthly view, the timesheet begin time is set to current time instead of 'default_begin' setting for duration modes.

The reason why I implemented the fix in `AbstractTrackingMode` was because I originally thought this issue was occurring for both `DurationOnlyMode` and `DurationFixedBeginMode`. As I write this comment, I realize it only affects `DurationOnlyMode`, so I guess it would be better to fix [`DurationOnlyMode::create`](https://github.com/kevinpapst/kimai2/blob/master/src/Timesheet/TrackingMode/DurationOnlyMode.php#L59) directly?

kevinpapst:
Not sure, as I'd say the `DurationOnlyMode` behaves as expected. It creates the record with start = now.
We could argue if that makes sense, but unfortunately it is out there and definitely used by companies.
So changing this behavior might break existing workflows.

Just imagine a company with night shifts using the DurationOnly mode. 
Changing the start time to a fixed date could/would change time logging for them.
Before your change they started tracking on the current day at 20:00 and stopped on the next day at 04:00.
With your change they suddenly start tracking at 08:00 and stop at 16:00 => which is not even near their working shift.

If you want to have the fixed start time, why don't you use `DurationFixedBeginMode`?

delvalt:
I'm not sure we're talking about the same thing here: the fix I want to implement only changes the default input value for the "From" field in the timesheet creation form on a `GET /timesheet/create` request (as described in #968), but doesn't alter the behavior of timesheet creation done in `POST /timesheet/create`, meaning the "From" field can still be customized by the user and won't be "forced" to 'default_begin' upon timesheet creation.

---

For example, if I'm in DurationOnlyMode with default_begin = 08:00, when I click a specific day on the calendar, I get the following HTML response:
kimai-timesheet-create-2

The default value of the "From" input field ignores the default_begin, which is inconsistent with what you get from clicking the [+] button to create a timesheet (which uses the default_begin value):
kimai-timesheet-create-1


I hope I have clarified the purpose of this PR.

If I'm missing something obvious about the process, I sincerely apologize for wasting your time on this minor fix 😅

delvalt:
No worries, my initial description was really ambiguous.

> And again: If you want to have the fixed start time, why don't you use DurationFixedBeginMode?

I'm not personally using Kimai at the moment (but might in the near future), I was looking for a nice Symfony project to contribute to during my spare time and jumped into #968 because it was tagged "good first issue"!
But yeah, I suppose using DurationFixedBeginMode in that case is the way to go.

kevinpapst:
Damned, I wrote a long answer and lost it - thanks to a browser hang.

I really appreciate your PR, especially as you don't use Kimai 👍 

Maybe it would be good if you use Kimai, otherwise you can easily miss some different simple solution. 

See: I hate rejecting PRs, because this usually means frustration and wasted times on both ends and I could really need some help here. But: this PR probably is a candidate for a "wont fix". Which is entirely my fault, because I haven't checked that label in a long while.

But I need time to think about and test it. Right now I believe this is more a logical problem or documentation issue and not necessarily a missing feature or bug. But I'll come back to it once I had the time to verify it.

Thanks again for your feedback and code and if you look for some more work, I am happy to look through the issues with you.

delvalt:
There's really no problem with you rejecting the PR, don't worry about it. I understand how difficult it can be to maintain such an app alone (or with very few contributors).

I don't have much free time at the moment, but I think I should be able to contribute a bit! Also, I'll start using Kimai to get a better grasp on the app (just need to install a prod environment first).
Do you have any idea of small PRs I could start working on?

kevinpapst:
I am really sorry for rejecting your first PR for Kimai 😭 

I added this sentence to the docs which clarifies the behavior of the mode
```
This mode does not use the `Default start-time` but always displays the current time as start for new entries.
```

The solution for #968 would probably be a new time-tracking mode.
But I closed the issue, as nobody else was interested in that feature for 2.5 years.

Thanks again for your contribution!!!! 👍

delvalt:
Don't worry about it 😉
This seems a good addition to have in the doc.

I take this opportunity to tell you that I've started to use Kimai to track my time, and I really enjoy it so far! Even though I use it to track only my own time, it helps me get organized in my work, so thanks for developing and maintaining this app. I hope I will find some time to contribute again to it!

Issue #2947, opened 2021-11-18
CSRF Tokens are not properly refreshed on some form submissions


tdozbun-reno:
**Describe the bug**
In 1.16.2, A number of form actions use CSRF tokens, 
but when the token is used/consumed, refreshToken is passed the value of the token instead of the ID of the token (by mistake?)
This means that the token is not refreshed immediately and can continue to be reused.
Getting a token with the same ID from CsrfTokenManager will provide the same value.

Most of the uses of CSRF tokens look something like the following (comments mine):
```php
    public function duplicateAction(Timesheet $entry, Request $request, string $token, CsrfTokenManagerInterface $csrfTokenManager): Response
    {
    //receive the request with $token as the value string of the token

        //create a token object with the ID of the token and the provided value of the token and validate it
        if (!$csrfTokenManager->isTokenValid(new CsrfToken('timesheet.duplicate', $token))) {
            $this->flashError('action.csrf.error');

            return $this->redirectToRoute('timesheet');
        }

        //if the token was valid, refresh the token
       //Symfony Documentation: "Generates a new token value for the given ID." 
       //    @param string $tokenId The token ID.
        // ISSUE: the token value string is passed instead of the ID 
        $csrfTokenManager->refreshToken($token); // <------------------

        //pass the submitted token value string...
        //Note: care should be taken not to reuse the token string and instead get a new one from the token manager.
        return $this->duplicate($entry, $request, 'timesheet/edit.html.twig', $token);
    }
```

**To Reproduce**
1. Submit any form successfully (which uses similar token code) 
2. Submit the form again with the same token string
3. See the lack of an error.

**Verifying the problem**
This can be verified by adding debug lines to the DoctorController::deleteLogfileAction or any similar action  (sorry for not using proper logging)

```php
public function deleteLogfileAction(string $token, CsrfTokenManagerInterface $csrfTokenManager): Response
    {
        if (!$csrfTokenManager->isTokenValid(new CsrfToken('doctor.flush_log', $token))) {
            $this->flashError('action.csrf.error');

            return $this->redirectToRoute('doctor');
        }
        var_dump('$token passed: '.$token); //<-----------------
        var_dump('token generated before refresh: '.$csrfTokenManager->getToken('doctor.flush_log'));//<-----------------

        $csrfTokenManager->refreshToken($token);
        var_dump('token generated after refresh: '.$csrfTokenManager->getToken('doctor.flush_log'));//<-----------------

[...]

PR #2948, opened 2021-11-18
Update CSRF Token IDs for Issue kevinpapst/kimai2#2947

tdozbun-reno:
## Description
This change passes the correct Token IDs to refreshToken; intended to patch issue #2947

## Types of changes

## Checklist
Code-check fails, but a fresh copy 1.16.2 from master also fails...
- [N/A] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [yes] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

### DIFF
--- a/src/Constants.php
+++ b/src/Constants.php
@@ -17,11 +17,11 @@ class Constants
     /**
      * The current release version
      */
-    public const VERSION = '1.16.2';
+    public const VERSION = '1.16.3';
     /**
      * The current release: major * 10000 + minor * 100 + patch
      */
-    public const VERSION_ID = 11602;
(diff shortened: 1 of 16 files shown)


kevinpapst:
Some bugs are so stupid, you really wonder how much alcohol was necessary... 😁 

I had the fixed almost ready locally using constants, but now I take yours instead.

Thanks for spotting that mistake 👍

kevinpapst (dismissed):
Looks good to me now 👍

kevinpapst (approved):

PR #2952, opened 2021-11-19
Add list ALLOWED_URL_SCHEMES for Markdown link

noobpk:
Fix bug Stored XSS via Markdown at the comment in Project
Disclousre: https://huntr.dev/bounties/89d6c3de-efbd-4354-8cc8-46e999e4c5a4/

## Description
When rendering to Markdown, the application does not filter and check the properties are valid, so when the user enters ``[XSS](javascript:alert(`document.domain`))`` it will render as `<a href="javascript:alert(document.domain)">XSS</a> `.

## Types of changes

## Checklist

### DIFF
--- a/src/Twig/Runtime/MarkdownExtension.php
+++ b/src/Twig/Runtime/MarkdownExtension.php
@@ -126,6 +126,25 @@ final class MarkdownExtension implements RuntimeExtensionInterface
      */
     public function markdownToHtml(string $content): string
     {
+        $ALLOWED_URL_SCHEMES = array("ftp", "ftps", "http", "https", "mailto", "sftp", "ssh", "tel", "telnet", "tftp", "vnc");
+        
+        $pattern = '/([\[\s\S\]]*?)\(([\s\S]*?):([\[\s\S\]]*?)\)/';
+        # regex check
+        preg_match_all($pattern, $content, $matches, PREG_SET_ORDER);
+        if($matches) {
+            foreach ($matches as $match) {
+                // get value of group regex
+                $scheme = $match[2];
+            }
+            // scheme check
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thank you @noobpk 👍 awesome work! I saw the huntr post as well, just need some time to explore the issue. 
I remember this code and the following method, which might be the better place to fix the problem:
https://github.com/kevinpapst/kimai2/blob/master/src/Utils/ParsedownExtension.php#L52

noobpk:
> Thank you @noobpk 👍 awesome work! I saw the huntr post as well, just need some time to explore the issue.
> I remember this code and the following method, which might be the better place to fix the problem:
> https://github.com/kevinpapst/kimai2/blob/master/src/Utils/ParsedownExtension.php#L52

Ah, I had a look but didn't see it used so I quickly wrote a filter for `markdownToHtml`. ^^!

noobpk:
When I debug I see comment template when render is using `md2html`
https://github.com/kevinpapst/kimai2/blob/56d02673eddacf617e15f617af39db51bb840eb3/templates/embeds/comments.html.twig#L38

It will call to `markdownToHtml` of `getFilters` function
https://github.com/kevinpapst/kimai2/blob/56d02673eddacf617e15f617af39db51bb840eb3/src/Twig/RuntimeExtensions.php#L44-L53

That's why I wrote more for `markdownToHtml` 😅. Sorry if that's not true
You can customize my fix if you feel it doesn't fit. And please confirm my report on huntr @kevinpapst . ♥️

kevinpapst:
Thank you overmuch, noch for your PR and the security notice!
I am using built-in functions of the library to filter user input.
The PR #2961 has all the infos.

PR #2972, opened 2021-11-25
fix two opening form tags

### LINKED ISSUE #2971: invalid generated code in html
**Describe the bug**
 would like to draw your attention to invalid generated code in html.  For example on timesheet and quick_entry page.
1. form inside another html form
2. < is "&lt ;" and > is "&gt ;")

 - Kimai version: 1.16.5
 - PHP version: 7.4.
 - Browser Firefox

kevinpapst:
## Description

Fixes #2971 

## Types of changes

## Checklist

### DIFF
--- a/templates/macros/search.html.twig
+++ b/templates/macros/search.html.twig
@@ -14,7 +14,7 @@
                     {% set orderHasError = form.orderBy.vars.errors|length > 0 or form.order.vars.errors|length > 0 %}
                     {% set orderBy = form_widget(form.orderBy) %}
                     {% set order = form_widget(form.order) %}
-                    {{ form(form) }}
+                    {{ form_rest(form) }}
                     <div class="form-group{% if orderHasError %} has-error{% endif %}">
                         {{ form_label(form.orderBy) }}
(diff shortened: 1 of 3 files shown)


BlackKerio:
@kevinpapst I don't know if you tried to fix all the bugs I pointed out before, but unfortunately the bug persists. 
In the index.html.twig template, the {{ form_widget(form.rows) }} line still generates in the tbody 
&lt;div id=&quot;quick_entry_form_rows&quot; data-prototype=&quot;
&quot;    
&quot;&gt;     
&lt;/div&gt;
and after the data-prototype="" statement, the kimai.theme.html.twig template inserts the table twice between the quotes before the 'div' termination. The first one is not in valid html ... the arrow characters and quotes are replaced by text form & lt; & gt; & quot;, so the table is not displayed. The second is in html and is visible.

kevinpapst:
No, I didn't try to fix the div inside table, as it doesn't cause any bugs I know off.

Unfortunately I don't exactly understand what you are saying. Which table is not displayed?

kevinpapst:
This is on purpose. It is the template (prototype) for new rows.

PR #2982, opened 2021-12-02
Per user reports for rate and internal rate in addition to reports of worked hours.

jonaswinkler:
## Description
The current weekly / monthly / yearly reports per user / for all users currently display working hours. I'd like to extend this feature to also display rate / internal rate in a similar fashion.

This is done by adding a drop down that the user can use to select between duration, rate and internal rate:

image

Duration is selected by default to not alter current behavior.

The displayed data will adjust accordingly:

image

Affected reports:

* All reports for one user
* All reports for all users

First of all, would you agree to have this feature in Kimai? If so, there are a couple more things I need to work on:

- [x] Unit tests
- [ ] Documentation
- [ ] Total rows still show "Working hours total" or similar

Questions I still have that may need resolving:

- [ ] I don't find the repetitive use of `isDuration ? value|duration : value|money` particularly beautiful. Is there a better way to do this? I have no experience with PHP at all.
- [x] ~Is there a formatter for rates that takes care of using the correct number of decimal places?~ Yes.
- [ ] Should this feature be guarded by certain permissions?

## Types of changes

## Checklist

### DIFF
--- a/templates/reporting/report_by_user.html.twig
+++ b/templates/reporting/report_by_user.html.twig
@@ -7,7 +7,7 @@
     {% embed '@AdminLTE/Widgets/box-widget.html.twig' %}
         {% import "macros/widgets.html.twig" as widgets %}
         {% block box_before %}
-            {{ form_start(form, {'attr': {'class': 'form-inline'}}) }}
+            {{ form_start(form, {'attr': {'class': 'form-inline', 'id': 'by-user-form'}}) }}
         {% endblock %}
         {% block box_after %}
             {{ form_end(form) }}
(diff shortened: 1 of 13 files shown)


templates/reporting/report_by_user.html.twig
```diff
@@ -19,11 +19,24 @@
...
+                        {{ form_widget(form.sumType) }}
+                    </li>
+                </ul>
+            </div>
         {% endblock %}
         {% block box_body_class %}{{ box_id }} table-responsive no-padding{% endblock %}
         {% block box_body %}
             {% set totals = {'totals': 0} %}
             {% set columns = 2 %}
+            {% set fieldName = form.sumType.vars.value %}
```
jonaswinkler, line 37: Not sure if this is the correct way to access this data. It's either 'duration', 'rate' or 'internalRate'.

templates/reporting/report_user_list.html.twig
```diff
@@ -22,6 +32,9 @@
             {% else %}
                 {% set absoluteTotals = 0 %}
                 {% set totals = {} %}
+                {% set fieldName = form.sumType.vars.value %}
+                {% set totalFieldName = 'total' ~ (fieldName | capitalize) %}
```
jonaswinkler, line 36: Use of capizalize formatter doesn't feel right here, but gets the job done.... is this acceptable?

kevinpapst:
HI Jonas, thanks!

Pretty large PR, which I need to check in detail once I have the time. Please standby for a while :)

kevinpapst:
Hi @jonaswinkler - it took a while, but I added your idea to master branch today.
Unfortunately and sorry for that(!) I didn't use your PR. 
I refactored a lot of things in the reports in the meantime and used a slightly different approach for rendering the values.
See https://github.com/kevinpapst/kimai2/pull/3155 for more technical infos.

Thanks again for your idea and code examples 👍 which I used a couple of times while adding the changes.
Will be available soon with the next release & update.

jonaswinkler:
Hi @kevinpapst.

Thanks for adding the feature! Very excited to see it in action.

No worries about not merging the PR. Part of me also just wanted to dabble with PHP a bit, and learning new frameworks and tools is always worth it.

kevinpapst:
Cloud will be updated soon.

PR #2995, opened 2021-12-06
link entities to invoices

Conni2461:
## Description

Continuation of #2414 (did rebase it) based on this discussion: https://github.com/kevinpapst/kimai2/discussions/2630

Not sure if thats the intended way

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [x] I verified that my code applies to the guidelines (`composer code-check`)
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [x] I agree that this code is used in Kimai and will be published under the [MIT license](https://github.com/kevinpapst/kimai2/blob/master/LICENSE)

PR #3080, opened 2022-01-20
link timesheets to invoices

kevinpapst:
## Description

Closes #2995 
Closes #2414 
See discussion https://github.com/kevinpapst/kimai2/discussions/2630

## Types of changes

## Checklist

### DIFF
--- a/src/Entity/Timesheet.php
+++ b/src/Entity/Timesheet.php
@@ -318,11 +318,27 @@ class Timesheet implements EntityWithMetaFields, ExportItemInterface
      * @ORM\OneToMany(targetEntity="App\Entity\TimesheetMeta", mappedBy="timesheet", cascade={"persist"})
      */
     private $meta;
+    /**
+     * @var Collection<Invoice>
+     *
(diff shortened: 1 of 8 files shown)


kevinpapst:
@Conni2461 is that a feature that you actually require or can you already work with the adjusted InvoiceCreatedEvent, which now allows to fetch the entities of the invoice?

Conni2461:
I'll need to check. I'll try to find some time for it this week.

Thanks for putting in all this work :)

kevinpapst:
There is not much enough interest to implement that fully

LiaraAlis:
Any chance to implement this after all? I'm happy to provide support here.

kevinpapst:
Nope. Both concept and goals are unclear and merely anyone wants it.

LiaraAlis:
To be honest I don't understand why not. It totally makes sense to link information to invoices that are needed to understand on what the invoice depends.
And it is required to make it possible to regenerate invoice documents or improve the invoicing process (generating credit documents and so on…).

Issue #3109, opened 2022-01-30
Colored Dots in Select2 Components


kulturbande:
**Is your feature request related to a problem? Please describe.**
I have several Projects that have the same name but different hourly rate. For that reason I use a color code to make easier, which is working fine in the timesheet view. But it is confusing if I'm creating the timesheet entry, because the projects have the same name.

**Describe the solution you'd like**
I would like to add a colored dot before the name on all entities that supports a color (Customer, Project, etc.). 

**Describe alternatives you've considered**
A custom CSS or an own plugin to patch the behavior.

**Screenshot**
![Aufnahme 2022-01-30 at 22 18 39@2x](https://user-images.githubusercontent.com/68833/151718368-ea089034-9da1-4324-9eb0-c62eaca3a1a6.png)

I already forked the project and was trying to add these function. I currently try to figure out how to use the webpack version in my Kimai Docker Setup.

kevinpapst:
You can check this code as example (color chooser): 
https://github.com/kevinpapst/kimai2/blob/master/assets/js/plugins/KimaiFormSelect.js#L124

I guess the code cannot be re-used, because the data project collection uses the `color` field.

And this activates the color renderer:
https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/ColorChoiceType.php#L77

Should be supported in all fields:
https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/ActivityType.php
https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/ProjectType.php
https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/CustomerType.php

kulturbande:
Thx for the hint. Currently I have some trouble to get Symfony deliver the newly created assets, but I will figure that out. I will provide a PR for that feature.

kulturbande:
I tried multiple things:

1. My first thought was, that the selects are API driven, because I saw the color field in the API responses. So I extended the Select Plugin https://github.com/kevinpapst/kimai2/blob/master/assets/js/plugins/KimaiFormSelect.js#L124
2. Then I took a look into the mentioned form types. I was able to extend the default data config to enable the color field in the select2 - component: https://github.com/kevinpapst/kimai2/blob/master/src/Form/Type/ActivityType.php#L88
[...]

PR #2902, opened 2021-11-03
Next major version 2 - Tabler, 2FA and more...

kevinpapst:
## Demo

Can be tested at https://demo-branch.kimai.org/de/login with the usual logins (eg `susan_super` and `kitten`).

Call for help with RTL languages: https://github.com/kevinpapst/kimai2/issues/3384

This CODE IS **NOT** LICENSED under the **MIT license** any longer, but it is **AGPL v3** now.

## Description

**Decisions / Done**
- [x] Migrate to tabler.io and one gazillion frontend updates
- [x] New configurable (in-browser editing) dashboard 
- [x] New report start screen
- [x] Remove `App\Utils\Translator` and `kimai.i18n_domains`
- [x] Updated to Symfony 6 and upgraded all composer packages to latest release
- [x] Store column visibility server side and 
  - [x] sticky "profile" per device for desktop and mobile, each can configure different columns
- [x] Upgrade to new Security (Passport) System
  - [x] Login Form
  - [x] API
  - [x] SAML
  - [x] LDAP (**needs more testing!**)
- [x] Remove Sweetalert2 in favor of Bootstrap Toasts & Modals
- [x] Remove `gd` extension requirement
- [x] Removed third-party mailer packages
- [x] Upgraded Chart.JS to v3.8
- [x] Removed deprecated code
- [x] migrate to FullCalendar >= 5.10.2
- [x] migrate from Moment.js to Luxon
- [x] migrate from daterangepicker to litepicker (dependency free and without locale files)
- [x] removed jQuery
- [x] delete settings for [Random colors](https://github.com/kevinpapst/kimai2/discussions/3446) and limited colos (default true for both now)
- [x] delete setting `OFF` for `Minute selection for Duration`
- [x] 2FA with [TOTP](https://symfony.com/bundles/SchebTwoFactorBundle/6.x/providers/totp.html)
- [x] True "Remember me" mode
- [x] New "welcome" wizard for first time users
- [x] expense plugin widgets
- [x] Notification support via NotificationsSubscriber
- [x] Context menu for calendar entries
- [x] Migrate to [webpack-encore 4](https://github.com/symfony/webpack-encore/releases/tag/v4.0.0)
- [x] new "system user" field for user - hide in reports and dropdowns
- [x] integrated "recalculate rates" logic from external bundle
- [x] add configured colors to tag dropdown
- [x] Confirm edit dialogs with keyboard shortcut 
- [x] context menu in all listing tables with the option to hide the "actions dropdown" column
[...]

PR #3167, opened 2022-02-20
Feature/colored select2 component issue 3109

kulturbande:
## Description
Show the color in the timesheet edit form for customer, project, and activity. This is a good way to select the correct entry if you have multiple e.g. activities with the same name (and maybe different rates). 

Issue Reference https://github.com/kevinpapst/kimai2/issues/3109

Example:
Aufnahme 2022-02-20 at 21 06 50@2x

## Types of changes

## Checklist


kevinpapst:
I like the changes in general and the try to simplify the code by re-using the color attribute.
But: I'd prefer if the existing markup for the "colored dot" would be re-used. 
Then we would need the support for `data-renderer=color-dot` and `data-renderer=color`.

kulturbande:
Sure, that would make sense. I was reusing only the previous select solution, but I can have a look if colored dot of the table is usable as well to make it "dryer".

kulturbande:
I updated the usage of the `color-dot`:

- use a background instead of an icon
- use the same icon - class for the select2 options
- moved the definition of the `.color-dot` - class to a new utils.scss (I couldn't find a proper place)

Data Table:
Aufnahme 2022-02-22 at 22 27 20@2x

Timesheet Edit:
Aufnahme 2022-02-22 at 22 27 35@2x

kulturbande:
@kevinpapst What do you think of the current implementation? Should it be improved or is it mergable?

kulturbande:
@kevinpapst I rebased the branch to have the latest main - changes included

kulturbande:
@kevinpapst Do you have any objections? I would rebase the master and resolve the current conflicts.

kulturbande:
No worries! Keep me posted, then I rebase the branch again.

kulturbande:
And thank you for your dedication!

kulturbande:
I rebased and updated the PR to the latest master changes

kulturbande:
I made a rebase with the latest master

kevinpapst:
I think the idea is good, so let me keep this for a while.

kulturbande:
Fair enough. I could also update to the latest master, but this is a bit exhausting, because the build increments changing quit offen.

kevinpapst:
Leave it as is, I will get back to it and resolve the conflicts

kevinpapst:
Short update:

I am working on the next major #2902 since quite a while and now we have several issues:
- it is using Tomselect instead of Select2 (I can adapt the code for it)
- the next major has the "random color" setting always active (which we should assume fr your PR as well)
- the API does return null on unset colors, which are likely the most

The last point is the biggest issue, as we cannot show a color upon select reload, which happens for project and activity if you choose a customer, or only activity if you select a project.

I wanted to replace all "null" with the color for the current object (it is not entirely random, but the color is calculated from the name, so it is "sticky random") in the AP, whichI I tried to do with with a JMS Serializer `SubscribingHandlerInterface`, but I couldn't get it to work. 
Now if someone has an idea how to do that, I will merge this PR into #2902

kevinpapst:
Closing, needs re-write for #2902

PR #3406, opened 2022-07-09
Budget graph in project details

kevinpapst:
## Description

Can be tested at http://demo-lts.kimai.org with `susan_super` and `kitten` / `admin` and `password` .

- show budget timeline for monthly (bar chart) and lifetime budgets (line chart)
- money chart will be shown if permitted, otherwise duration chart (if permitted)
- support rendering of negative durations

Slightly off-topic:
- do not copy description if restarting via calendar via drag&drop

## Types of changes

## Checklist


src/Calendar/TimesheetEntry.php
```diff
@@ -38,7 +38,8 @@ public function getData(): array
         }
 
         return [
-            'description' => $this->timesheet->getDescription(),
+            // restarting a timesheet should not copy the description - @version 1.21
+            //'description' => $this->timesheet->getDescription(),
```
pfrenssen, line 42: Why is this commented out? When restarting an activity it is now necessary to re-enter the description by hand?

kevinpapst: Because restarting an activity is not the same as restarting a record.
Same behavior as the difference between restart in timesheet listing and restart activity from top of the screen.
Could be made configurable at some point.

pfrenssen: Thanks, I see indeed that activities can still be restarted from the "Timesheets" page. I agree it makes sense because the descriptions are not visible in the calendar, and if the page is not refreshed after an activity is started in another window, the activities can be stale.

PR #3447, opened 2022-07-30
Croatian translation 20220730

milotype:
## Description
Added strings for "hr" (Croatian)

## Types of changes

## Checklist


kevinpapst:
Wow, strange that I missed that. Thanks for finding and fixing @milotype 👍

PR #3480, opened 2022-08-22
Added configuration for small business rule (German law § 19 UStG.)

LiaraAlis:
## Description
This PR adds a new setting to the invoice section. When this option is checked, tax calculation is disabled and a note is automatically added in HTML and PDF invoices. Tax is hidden from invoices and the tax rate option in invoice templates is disabled.
Please follow this discussion for more information: https://github.com/kevinpapst/kimai2/discussions/3478

## Types of changes

## Checklist

### DIFF
--- a/templates/invoice/renderer/default-pdf.pdf.twig
+++ b/templates/invoice/renderer/default-pdf.pdf.twig
@@ -5,13 +5,18 @@
 {% set currency = model.currency %}
 <html lang="{{ language }}">
 <head>
-    <meta charset="UTF-8">
-    <title>{% block title %}{{ model.invoiceNumber }}-{{ model.customer.company|default(model.customer.name)|u.snake }}{% endblock %}</title>
-    <style type="text/css">
-        {{ encore_entry_css_source('invoice-pdf')|raw }}
-    </style>
+    {% block head %}
+        <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


LiaraAlis:
If this feature does not fit for merged into the standard, I could also imagine offering this as a plugin. 😊

LiaraAlis:
I have seen that there is also such a regulation in Austria. Maybe you can also adapt/add the translations for this right away.

kevinpapst:
Thanks for your first contribution and the PR 👍 

Unfortunately I have to reject these changes: I do not want any country/regulatory specific rules in the core. 
I know that Kimai is used in more than 30 countries and that couldn't be handled properly.

What I imagined was much simpler. Showing the subtotal/tax row only `{%if tax > 0 %}`, which means adding one if block to each invoice templates.

But having a bundle with specific rules for Germany would be nice as well!

LiaraAlis:
@kevinpapst I totally understand this. I move this changes to a plugin and make it available for everyone with no charge. 😊
I changed this PR to follow your suggestion and additionally add some twig blocks in all invoices templates. So customization should be easier. 😁

(Btw: Is there a reason why the stylesheet in the export timesheet pdf is written inline and in the invoice pdfs the `encore_entry_css_source()` function is used?)

LiaraAlis:
Here you can find the first dev version of my plugin: https://github.com/LiaraAlis/kimai2-SmallBusinessRuleBundle

PR #3484, opened 2022-08-23
Allow to hide zero tax rows via plugin

LiaraAlis:
## Description
This PR adds some twig blocks in the invoice templates. Additionally it hides subtotal and tax rows in the table when no tax is used.
For more information see original PR: https://github.com/kevinpapst/kimai2/pull/3480

## Types of changes

## Checklist

### DIFF
--- a/templates/export/pdf-layout.html.twig
+++ b/templates/export/pdf-layout.html.twig
@@ -77,8 +77,8 @@
 {% endblock %}
 </head>
 <body>
-{% block pdf_footer %}
 <!--mpdf
+{% block pdf_footer %}
 <htmlpagefooter name="myfooter">
 <table style="border-top: 1px solid #000000; font-size: 9pt; padding-top: 3mm; width: 100%">
 <tr>
@@ -101,8 +101,8 @@
 </table>
 </htmlpagefooter>
 <sethtmlpagefooter name="myfooter" value="on" />
-mpdf-->
 {% endblock %}
+mpdf-->
 {% block body_start %}{% endblock %}
 {% block summary %}
     {% block summary_header %}
--- a/templates/export/pdf-layout.html.twig
+++ b/templates/export/pdf-layout.html.twig
@@ -77,8 +77,8 @@
 {% endblock %}
 </head>
 <body>
-<!--mpdf
 {% block pdf_footer %}
+<!--mpdf
 <htmlpagefooter name="myfooter">
 <table style="border-top: 1px solid #000000; font-size: 9pt; padding-top: 3mm; width: 100%">
 <tr>
@@ -101,8 +101,8 @@
 </table>
 </htmlpagefooter>
 <sethtmlpagefooter name="myfooter" value="on" />
-{% endblock %}
 mpdf-->
+{% endblock %}
 {% block body_start %}{% endblock %}
 {% block summary %}
     {% block summary_header %}
--- a/templates/invoice/renderer/default-pdf.pdf.twig
+++ b/templates/invoice/renderer/default-pdf.pdf.twig
@@ -16,8 +16,8 @@
     {% endblock %}
 </head>
 <body>
-<!--mpdf
 {% block pdf_header %}
+<!--mpdf
 <htmlpageheader name="header">
     <table class="header">
         <tr>
@@ -30,10 +30,10 @@
         </tr>
     </table>
 </htmlpageheader>
-<sethtmlpageheader name="header" page="ALL" value="on" show-this-page="1" />
 {% endblock %}
 
 {% block pdf_footer %}
+<sethtmlpageheader name="header" page="ALL" value="on" show-this-page="1" />
 <htmlpagefooter name="footer">
     <table class="footer">
         <tr>
@@ -51,8 +51,8 @@
     </table>
 </htmlpagefooter>
 <sethtmlpagefooter page="ALL" value="on" name="footer" show-this-page="1" />
-{% endblock %}
 mpdf-->
+{% endblock %}
 
 {% block wrapper %}
 <div class="wrapper">
--- a/templates/invoice/renderer/default-pdf.pdf.twig
+++ b/templates/invoice/renderer/default-pdf.pdf.twig
@@ -54,7 +54,6 @@
 {% endblock %}
 mpdf-->
 
-{% block body_start %}{% endblock %}
 {% block wrapper %}
 <div class="wrapper">
     {% block addresses %}
@@ -201,6 +200,5 @@ mpdf-->
     {% endblock %}
 </div>
 {% endblock %}
-{% block body_end %}{% endblock %}
 </body>
 </html>
--- a/templates/invoice/renderer/default-pdf.pdf.twig
+++ b/templates/invoice/renderer/default-pdf.pdf.twig
@@ -5,13 +5,18 @@
 {% set currency = model.currency %}
 <html lang="{{ language }}">
 <head>
-    <meta charset="UTF-8">
-    <title>{% block title %}{{ model.invoiceNumber }}-{{ model.customer.company|default(model.customer.name)|u.snake }}{% endblock %}</title>
-    <style type="text/css">
-        {{ encore_entry_css_source('invoice-pdf')|raw }}
-    </style>
+    {% block head %}
+        <meta charset="UTF-8">
+        <title>{% block title %}{{ model.invoiceNumber }}-{{ model.customer.company|default(model.customer.name)|u.snake }}{% endblock %}</title>
+        {% block styles %}
+        <style type="text/css">
+            {{ encore_entry_css_source('invoice-pdf')|raw }}
+        </style>
+        {% endblock %}
+    {% endblock %}
 </head>
 <body>
+{% block pdf_header %}
 <!--mpdf
 <htmlpageheader name="header">
     <table class="header">
@@ -25,6 +30,9 @@
         </tr>
     </table>
 </htmlpageheader>
+{% endblock %}
+
+{% block pdf_footer %}
 <sethtmlpageheader name="header" page="ALL" value="on" show-this-page="1" />
 <htmlpagefooter name="footer">
     <table class="footer">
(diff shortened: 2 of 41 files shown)


templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -126,22 +150,33 @@ mpdf-->
...
+                {% endblock %}
             </tr>
+            {% endblock %}
         {% endfor %}
+        {% endblock %}
         </tbody>
         <tfoot>
+        {% block items_foot %}
+            {% if model.calculator.tax > 0 %}
+            {% block items_foot_subtotal %}
```
kevinpapst, line 162:
```suggestion
            {% block items_footer_subtotal %}
```

templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -126,22 +150,33 @@ mpdf-->
...
             <tr>
                 <td colspan="4" class="text-right">
                     {{ 'invoice.subtotal'|trans }}
                 </td>
                 <td class="last text-right">{{ model.calculator.subtotal|money(currency) }}</td>
             </tr>
+            {% endblock %}
+            {% block items_foot_tax %}
```
kevinpapst, line 170:
```suggestion
            {% block items_footer_tax %}
```

templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -126,22 +150,33 @@ mpdf-->
...
             <tr>
                 <td colspan="4" class="text-right">
                     {{ 'invoice.tax'|trans }} ({{ model.calculator.vat }}%)
                 </td>
                 <td class="last text-right">{{ model.calculator.tax|money(currency) }}</td>
             </tr>
+            {% endblock %}
+            {% endif %}
+            {% block items_foot_total %}
```
kevinpapst, line 179:
```suggestion
            {% block items_footer_total %}
```

templates/invoice/renderer/default.html.twig
```diff
@@ -1,145 +1,186 @@
...
-            {% endif %}
-            {% if model.customer.number is not empty %}
-                <br>
-                {{ 'label.number'|trans }}: {{ model.customer.number }}
-            {% endif %}
-            {% if model.query.project is not empty and model.query.project.orderNumber is not empty %}
+    {% block addresses %}
+    <div class="row">
+        <div class="col-sm-5">
+            {% block addresses_to %}
```
kevinpapst, line 20:
```suggestion
            {% block address_recipient %}
```

templates/invoice/renderer/default.html.twig
```diff
@@ -1,145 +1,186 @@
...
+                {% if model.query.project is not empty and model.query.project.orderNumber is not empty %}
+                    <br>
+                    {{ 'label.orderNumber'|trans }}: {{ model.query.project.orderNumber }}
+                {% endif %}
+            </address>
+            {% endblock %}
+        </div>
+        <div class="col-sm-2"></div>
+        <div class="col-sm-5">
+            {% block addresses_from %}
```
kevinpapst, line 42:
```suggestion
            {% block address_sender %}
```

kevinpapst (changes requested):
Thank you @LiaraAlis 👍 

I found many small things that I would like to smooth out, mostly block names which should be consistent between export and invoice templates.

My review is only for the first template, can you adapt the other ones accordingly with the same changes (search and replace should work if you used the same names).

templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -5,13 +5,18 @@
...
+        {% block styles %}
+        <style type="text/css">
+            {{ encore_entry_css_source('invoice-pdf')|raw }}
+        </style>
+        {% endblock %}
+    {% endblock %}
 </head>
 <body>
+{% block pdf_header %}
 <!--mpdf
```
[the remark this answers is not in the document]
LiaraAlis: Oh, you're right! I will fix this with the next commit. :)

templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -150,14 +185,20 @@ mpdf-->
...
 
+    {% block payment_terms %}
     {% if model.template.paymentTerms is not empty %}
     <div class="paymentTerms">
         {{ model.template.paymentTerms|md2html }}
     </div>
     {% endif %}
+    {% endblock %}
 </div>
+{% endblock %}
```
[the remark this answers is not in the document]
LiaraAlis: I think we don't need this empty blocks, because when you override one block, you can call the parent (`{{ parent() }}`).
Or do you have something in mind that I'm not thinking about?

LiaraAlis:
Should we also add a empty `pdf_header` block in the export template to match the invoice templates?

templates/invoice/renderer/default-pdf.pdf.twig
```diff
@@ -93,18 +107,25 @@ mpdf-->
             {{ model.query.project.orderNumber }}
         {% endif %}
     </p>
+    {% endblock %}
 
+    {% block items %}
     <table class="items">
         <thead>
+            {% block items_head %}
```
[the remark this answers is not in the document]
LiaraAlis: I'm not sure if this is a good idea. Beacuse the `items_header` block in the export template means a title, while the `items_head` blocks in the invoice templates means the `<thead>`.
Do you have a better idea for naming?

LiaraAlis:
@kevinpapst Can you please check my comments above? After that I can complete your change requests. 😁

templates/invoice/renderer/default-pdf.pdf.twig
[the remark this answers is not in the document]
kevinpapst: Why does it feel harder to come up with proper variable names, compared to developing an entire bundle 😁 

I am fine with skipping the suggestion 👍

kevinpapst:
You know what, I am a little afraid of the merge problems that will show pushen merging this PR.
Not sure if I mentioned it before, I am working on a massive rewrite of Kimai: see #2902 

I see three options how to move forward: 
- you try to merge your changes with this linked branch to see if merging is easy possible 
- you only add the necessary changes for your feature in here and
- you move all optional changes to the other branch

I know this is annoying, but I try to keep the changes as small as possible currently to avoid further merge issues.
Do you have any opinion on this?

templates/invoice/renderer/default-pdf.pdf.twig
LiaraAlis: I would skip the footer changes too - same reason as for the header. :)

templates/export/pdf-layout.html.twig
```diff
@@ -101,8 +101,8 @@
 </table>
 </htmlpagefooter>
 <sethtmlpagefooter name="myfooter" value="on" />
-mpdf-->
 {% endblock %}
+mpdf-->
```
kevinpapst, line 105: See above

kevinpapst (changes requested):
Okay, I really love your input in the PRs and hate being a pain in the ass... but I have to reject this for now. Can you re-target the changes to #2902 instead and ONLY apply the minimum necessary changes here.

I am fine with the blocks for the tax = 0 change, but everything else doesn't feel right.

The problem are the pretty generic block names which might to layout issues in existing templates. I searched through the library of my customers templates and found some duplicates already eg `block items`.

I simply cannot guarantee that these new blocks will not lead to some weird BC breaks, which should be avoided at all cost.

So my proposal: 
- target the v2 branch (the three comments are for it)
- add only the tax=0 change only in this PR

LiaraAlis:
> Okay, I really love your input in the PRs and hate being a pain in the ass... but I have to reject this for now. Can you re-target the changes to #2902 instead and ONLY apply the minimum necessary changes here.

I go with you. But I suggest that we go back to the start and think carefully about the topic how the templates in v2 could be structured to make it consistent across all invoice and export templates.
I'm also not happy with these namings and I think it is more elegant to refactor the export and invoice templates in v2.
In my eyes the current templates are too static and inflexible. For example, not even a logo is possible without copying the template to a new one.

For this, to make my plugin better, I need a change for the subtotal and VAT row and also a block before the payment terms. Can we do this?

And a off topic question:
Is there a collection of different invoice templates everywhere? I would like to change my templates but I don't like the default ones.

templates/export/pdf-layout.html.twig
```diff
@@ -77,8 +77,8 @@
 {% endblock %}
 </head>
 <body>
-{% block pdf_footer %}
 <!--mpdf
+{% block pdf_footer %}
```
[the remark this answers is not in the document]
LiaraAlis: What exactly do you mean with "self-containing blocks"?

LiaraAlis:
@kevinpapst I've updated the PR and reverted all the changes that added blocks. So only the change about the tax is left.
So we should take about the v2 templates later. Should we create an issue for that?

kevinpapst:
> So we should take about the v2 templates later. Should we create an issue for that?

Yes please 👍 

I tried to get proper information online but I could only find out. that you can add a `0,00€ Tax` or `0% VAT` row without problems (as registered EU VAT company). And that you have to stick to all regular invoice rules, even with reverse charge invoices (which means having that row is mandatory). 

Honestly: no company ever complained, so I kinda think that this row is no trouble for registered companies.
So I am not sure if that change is legally a risk for some companies / scenarios. 

This PR is a never ending story, I am sorry about that! It is a sensitive topic...

We should probably add a method `public function isHideZeroTax(): bool` somewhere, which can be triggered by your plugin. That values should be passed down to the templates, so they can hide the tax row or leave it, depending on the setting.

I will look at it later and send an update.

LiaraAlis:
> This PR is a never ending story, I am sorry about that! It is a sensitive topic...

You have not to sorry about that. You are right, I also thought about this and searched. But I cannot find anything about this.
In my eyes, for small businesses printing 0% VAT is misleading. In reality there is not 0% tax, but none at all (or in programming language: `0 != null` 😁). But I don't know if this is true. I have to ask a tax consultant, but I do not have someone. 😅

> We should probably add a method `public function isHideZeroTax(): bool` somewhere, which can be triggered by your plugin. That values should be passed down to the templates, so they can hide the tax row or leave it, depending on the setting.

I like the idea. I think this way is the best to lower the risk.

LiaraAlis:
I've added the property to the `InvoiceModel.php`. I think it is a good place and I didn't want to add the property to a doctrine model, because it is only a internal property and not stored in database.
I also changed the check in the templates.

kevinpapst (approved):
That was a long journey, but I like where it brought us 😄 

Thank you @LiaraAlis 👍

Issue #3481, opened 2022-08-22
Use "Content-Disposition: inline" for PDF previews


LiaraAlis:
For me it is very annoying when you have to download the PDF invoice every time before you want to see the preview.
I would be very happy if the preview (for PDF files) is opened with "Content-Disposition: inline" so that the browser can open the PDF itself without downloading it. This should only be the case for the preview, not for the save function.
This could be added for invoices as well for the timesheets.

If you want I can make a PR for this. Or is there any reason that the download for the preview is necessary?

kevinpapst:
I am pretty sure there was some issue, but I honestly cannot remember why I have chosen the download. 

Let's give it a try 👍

PR #3486, opened 2022-08-23
Added inline disposition for PDF previews

LiaraAlis:
## Description
This PR changes the behavior how the files are sent to the client. If the file type can be shown inline, the corresponding renderer now supports inline disposition. This will automatically applied for the invoice preview function and for exporting pdfs if request is not set to mark the entries as exported.
If the invoice is saved or the entries are set to mark as exported, the files are sent as attachment as before.
See also: https://github.com/kevinpapst/kimai2/issues/3481

## Types of changes

## Checklist


kevinpapst:
Can you please mark the new interface `@deprecated remove me in 2.0`. 
We should move it to base exporter interface in #2902 .

LiaraAlis:
> And remove the duplicate interface & trait => use the one in exporter. In the next major these functions move closer together.

You mean that we use the `DispositionInlineTrait` and `DispositionInlineInterface` in `Export/Base` both for export and invoice renderers? Should I change the namespace or should I leave it in `Export/Base`?

LiaraAlis:
@kevinpapst I've pushed some changes, I hope it is as you meant.

src/Controller/ExportController.php
```diff
@@ -113,6 +114,11 @@ public function export(Request $request): Response
 
         $renderer = $this->export->getRendererById($type);
 
+        // display file inline if supported and `markAsExported` is not set
+        if ($renderer instanceof DispositionInlineInterface && !$query->isMarkAsExported()) {
+            $renderer->setDispositionInline(true);
+        }
+
         if (null === $renderer) {
```
kevinpapst, line 122: Probably the new code should go below the check for null

src/Controller/ExportController.php
```diff
@@ -113,6 +114,11 @@ public function export(Request $request): Response
 
         $renderer = $this->export->getRendererById($type);
 
+        // display file inline if supported and `markAsExported` is not set
```
kevinpapst, line 117: I understand the choice in invoice screen where it makes sense, because the invoice listing is shown.
But what was your idea behind the check in exporter screen?

I guess it is "if you mark as export, you want to either share the document with your client or put it in your archive, and a direct download saves a couple of clicks", am I correct?

src/Export/Base/DispositionInlineTrait.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function getDisposition(): string
+    {
+        return $this->disposition;
+    }
+
+    /**
+     * @param bool $useInlineDisposition
+     * @return void
+     */
```
kevinpapst, line 32: all redundant comments can be removed (redundant as in: already 100% clear through function definition)

src/Invoice/Renderer/JsonRenderer.php
```diff
@@ -10,13 +10,16 @@
...
 use App\Entity\InvoiceDocument;
+use App\Export\Base\DispositionInlineInterface;
+use App\Export\Base\DispositionInlineTrait;
 use App\Invoice\InvoiceFilename;
 use App\Invoice\InvoiceModel;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 
-final class JsonRenderer extends AbstractTwigRenderer
+final class JsonRenderer extends AbstractTwigRenderer implements DispositionInlineInterface
```
kevinpapst, line 19: I tried locally and JSON is not helping inline IMO. 
If someone uses JSON (some IT companies do for further processing) this is very likely via the InvoiceCommand. Same goes for TEXT invoice format.

src/Controller/ExportController.php
LiaraAlis: You're totally right. I did not realize this check. 😅

src/Controller/ExportController.php
LiaraAlis: Correct. That was my intention. Maybe I'm wrong here, but because there is no preview button, I thought about that if you don't check "mark as exported" you want to quick check you're timesheet. So in this case it is better to display the pdf inline.
But if this behavior is not logical enough for the user, I can remove this check. Otherwise there should be a separate button for this.

src/Invoice/Renderer/JsonRenderer.php
LiaraAlis: Yes, you're right. I realized that I've installed a json viewer extension in my chrome, therefore JSON is formatted pretty well. But this is not the standard.

src/Controller/ExportController.php
kevinpapst: I am not really using the export myself, so I can only make an educated guess. I'd say its good and if users do not get it, they will "post their complains" and we can work on improving it later on...

src/Invoice/Renderer/JsonRenderer.php
kevinpapst: Further test here revealed: XSLX and DOCX are downloaded no matter which disposition is used.
So in my opinion it is enough to add the interface only to the PDF renderer.

LiaraAlis:
@kevinpapst I've updated PR. I hope it is OK now.

kevinpapst (approved):
That is a nice add-on, thanks @LiaraAlis 👍

PR #3504, opened 2022-09-01
Fixes #3495: Quick entry during grace period

### LINKED ISSUE #3495: Weekly working hours: New entries do not take grace period into account
**Describe the bug**
When using the weekly working hours (quick entry), it doesn't seem to take the grace period into account. When I try do add an entry today (01 Sept 2022) for 01 Aug 2022, the input element is disabled

**To Reproduce**
Use these settings for your lockdown and grace periods:
image

And on the first day of a month, try create a time entry for the first day of last month. The input will be disabled, while existing time entries are still editable. Adding them through "My Times" or through the calendar does work.

**Logfile**
N/A

**Additional context**
 - Kimai version: 1.22.1
 - PHP version: 8.1.9
 - Device: Ubuntu 20.04.4 LTS (Focal Fossa) 64-bit Laptop 16 inch
 - Browser: Firefox 104.0

Jelle-S:
During the grace period, new time entries could not be added on the
first day of the lockdown period. This was due to an edge case where the
start time of the timesheet exactly matched the start time of the
lockdown period, and we checked for "greater than" instead of "greater
than or equal to".

## Types of changes

## Checklist

### DIFF
--- a/src/Timesheet/LockdownService.php
+++ b/src/Timesheet/LockdownService.php
@@ -86,9 +86,9 @@ final class LockdownService
         }
 
         // further validate entries inside of the most recent lockdown
-        if ($timesheetStart > $lockdownStart && $timesheetStart < $lockdownEnd) {
+        if ($timesheetStart >= $lockdownStart && $timesheetStart <= $lockdownEnd) {
             // if grace period is still in effect, validation succeeds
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):
Thank you @Jelle-S  👍

PR #3583, opened 2022-10-18
Naive implementation of non-accesskey hotkeys

### LINKED ISSUE #3572: Confirm edit dialogs with keyboard shortcut
When I create or edit entries I frequently find myself pressing ctrl+enter after entering the discription (expecting the dialog to save and close). Clicking the button or tabbing all the way down is the only way to save it, though.

A keyboard shortcut to trigger a dialog's non-cancel action would be nice. I find ctrl+enter convenient, but that's just my preference.

I'm up for a PR if this is a feature others would like to see as well.

j2L4e:
Submit form modals on ctrl+enter.
Keyup events are checked against preset criteria. If criteria are met and exactly one visible element matches the provided selector a click event is fired on that element.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

close #3572

### DIFF
--- a/assets/js/KimaiLoader.js
+++ b/assets/js/KimaiLoader.js
@@ -40,6 +40,7 @@ import KimaiTeamForm from "./forms/KimaiTeamForm";
 import KimaiCopyDataForm from "./forms/KimaiCopyDataForm";
 import KimaiDateNowForm from "./forms/KimaiDateNowForm";
 import KimaiNotification from "./plugins/KimaiNotification";
+import KimaiHotkeys from "./plugins/KimaiHotkeys";
 
 export default class KimaiLoader {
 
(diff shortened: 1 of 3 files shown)


kevinpapst:
Seeing the isVisible check reminds me why I linked the JS modal stuff initially.

This code is now listening always. 
The modal code is always in the pages source code and once a modal was active, the buttons stay there until a new modal is opened and they are replaced.

Would it be better, to activate the key listener only when a modal is shown and then deactivated once it is hidden?
Yes, that would be tighter coupling, but less space for "troubles".

kevinpapst (approved):
Thanks, I'll give it a try 👍

PR #3625, opened 2022-11-05
remove unnecessary translations

kevinpapst:
## Description

remove unnecessary "invoice renderer" translations

## Types of changes

## Checklist


kevinpapst:
added in #2902

PR #3647, opened 2022-11-23
Add accessibility attributes to sortable table headers.

### LINKED ISSUE #3646: Table sorting is not accessible
### Describe the problem

The sorting of datatables cannot be understood or controlled by accessibility software such as screen readers. This makes it impossible for users with visual / motor disabilities to change the sorting on the tables.

Implement the relevant accessibility attributes, ref. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort

_No response_

### Screenshots

_No response_

pfrenssen:
Fixes #3646

## Description
Implements the necessary `ARIA` and `role` accessibility attributes to make it possible to read and control table sorting with accessibility software.

## Types of changes

## Checklist

### DIFF
--- a/templates/macros/datatables.html.twig
+++ b/templates/macros/datatables.html.twig
@@ -74,7 +74,11 @@
                             {% if title != 'actions' and not headerOptions.orderBy is same as(false) %}
                                 {% if orderBy == headerOptions.orderBy %}
                                     {% set headerClass = headerClass ~ ' sortable sorting_' ~ (order) %}
-                                    {% set ariaSort = order ~ 'ending' %}
+                                    {% if order == 'asc' %}
+                                        {% set ariaSort = 'ascending' %}
+                                    {% else %}
+                                        {% set ariaSort = 'descending' %}
+                                    {% endif %}
                                 {% else %}
                                     {% set headerClass = headerClass ~ ' sortable sorting' %}
                                 {% endif %}
--- a/templates/macros/datatables.html.twig
+++ b/templates/macros/datatables.html.twig
@@ -89,7 +89,7 @@
                             {% elseif title is not empty and title != 'actions' %}
                                 {% set headerTitle = (translationPrefix ~ title)|trans({}, translationDomain) %}
                             {% endif %}
-                            <th data-field="{{ title }}" {% if not headerOptions.orderBy is same as(false) %}data-order="{{ headerOptions.orderBy }}" role="button" {% endif %}{% if ariaSort %}aria-sort="{{ ariaSort }}" {% endif %}class="{{ headerClass }}">
+                            <th data-field="{{ title }}" {% if not headerOptions.orderBy is same as(false) %}data-order="{{ headerOptions.orderBy }}" role="button" {% endif %}{% if ariaSort is defined %}aria-sort="{{ ariaSort }}" {% endif %}class="{{ headerClass }}">
                                 {% if headerOptions.html_before is defined %}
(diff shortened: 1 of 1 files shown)


templates/macros/datatables.html.twig
```diff
@@ -74,6 +74,7 @@
                             {% if title != 'actions' and not headerOptions.orderBy is same as(false) %}
                                 {% if orderBy == headerOptions.orderBy %}
                                     {% set headerClass = headerClass ~ ' sortable sorting_' ~ (order) %}
+                                    {% set ariaSort = order ~ 'ending' %}
```
kevinpapst, line 77: Isn't there a space before the `ending` missing? What does `ending` mean?
Have you tried that on any page? What happens if order == false?

templates/macros/datatables.html.twig
```diff
@@ -84,7 +85,7 @@
                             {% elseif title is not empty and title != 'actions' %}
                                 {% set headerTitle = (translationPrefix ~ title)|trans({}, translationDomain) %}
                             {% endif %}
-                            <th data-field="{{ title }}" {% if not headerOptions.orderBy is same as(false) %}data-order="{{ headerOptions.orderBy }}" {% endif %}class="{{ headerClass }}">
+                            <th data-field="{{ title }}" {% if not headerOptions.orderBy is same as(false) %}data-order="{{ headerOptions.orderBy }}" role="button" {% endif %}{% if ariaSort %}aria-sort="{{ ariaSort }}" {% endif %}class="{{ headerClass }}">
```
kevinpapst, line 88:
```suggestion
                            <th data-field="{{ title }}" {% if not headerOptions.orderBy is same as(false) %}data-order="{{ headerOptions.orderBy }}" role="button" {% endif %}{% if ariaSort is defined %}aria-sort="{{ ariaSort }}" {% endif %}class="{{ headerClass }}">
```

kevinpapst (changes requested):

templates/macros/datatables.html.twig
pfrenssen: The `aria-sort` property can have two values: `ascending` and `descending`. We already have similar wordings in the class, `sorting_asc` and `sorting_desc`. This is converting `asc` into `ascending` by adding an appropriate `ending` :)

I you prefer I can try to rewrite it in a more easy to understand way.

kevinpapst: Oha, I didn't understand that. A bit too magical I'd say. 

One of my issues is the next major release which is right before the door and which needs merging: 
https://github.com/kevinpapst/kimai2/pull/2902

Can you refactor to something like `if 'desc' in order` please.

pfrenssen: Fixed in 98c3498e2aadf11b8df71ffc11e19723ae23776c

pfrenssen: I will make a second MR that targets v2. I'm sorry I was not aware that this was being worked on, but it is looking great!

kevinpapst (approved):
Thanks @pfrenssen 👍

kevinpapst:
I assume there are many more issues with screenreaders.

If you want to add further aria support, please checkout #2902 - I am going to call out a feature freeze somewhen in December to be able to fully concentrate on the next major release.

pfrenssen:
Thanks! I will keep an eye out if I find other parts that can benefit from accessibility enhancements, but it is looking very good already. I have been using it daily for almost a year and having very little trouble accessibility wise.

PR #3666, opened 2022-12-02
Users API: fix non-admins can't search for all users

ymarkus:
## Description
I noticed that if a non-admin user with permission `view_users` calls the API path `/api/users`, it will only return an array with one object: the user themselves.
Deleting fixes it as far as I can see.

## Types of changes

## Checklist

### DIFF
--- a/src/API/UserController.php
+++ b/src/API/UserController.php
@@ -95,7 +95,6 @@ final class UserController extends BaseApiController
     public function cgetAction(ParamFetcherInterface $paramFetcher): Response
     {
         $query = new UserQuery();
-        $query->setCurrentUser($this->getUser());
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thank @ymarkus 👍 
But what does the same user see in the frontend? How is the team setup in your installation?
Removing the line disables all permission checks and is very likely wrong.

kevinpapst:
Tested on demo.kimai.org - everything behaves correct.

Assign "view_user" to ROLE_USER, logion with `john_user/kitten` - you see only this user in the frontend.
Same goes for the API endpoint.
I don't see a bug. 
Happy to talk about it though in an issue.

ymarkus:
Thanks for the feedback, I've opened an issue! :)

Issue #3767, opened 2023-01-17
Sort language list by display name


itkfm:
### Describe the problem

The language list for setting the preferred language in the *preferences* page is currently sorted by the underlying “language code” value (at least that’s what I presume that it’s ordered by), not what is shown.

This leads for example to different “flavors” of English being sorted in an order like `F`, `F`, `M`, `U`, `G`, `G`.

### Describe the solution you'd like

Order those entries by their display name.

### Screenshots

![image](https://user-images.githubusercontent.com/53079668/212977525-9c0408a9-3979-4c81-8be8-e27bedf0cdc0.png)

itkfm:
I hope this is not a duplicate (I’m sorry if it is!), unfortunately the search results (for “language”) were ten a penny.

PR #3906, opened 2023-03-08
Natural sorting of language dropdown display values

adrianrudnik:
## Description
As mentioned in #3767 displayed names of the languages in the LanguageType dropdown are not sorted naturally.

This has some considerations that I skipped, also detailing about the reasoning:

- Differences in sorting due to the active locale, like DIN 5007-1 (o,ö are equal and should come before p), DIN 5007-1 variant 2, EN 13710 can not be considered, as PHP lacks an implementation for this. Currently natsort (or an UTF-8 codepoint approach) lacks the distinction between `Deutsch (Schweiz)` and `Deutsch (Österreich)` and sort it wrong (Österreich being behind Schweiz).
- Testing is fiddly with the `LocaleService` depending on both configuration and `\Locale` so I had to mock a specfic structure in the test.

## Types of changes

## Checklist

### DIFF
--- a/src/Form/Type/LanguageType.php
+++ b/src/Form/Type/LanguageType.php
@@ -29,11 +29,13 @@ final class LanguageType extends AbstractType
         $choices = [];
         foreach ($this->localeService->getAllLocales() as $key) {
             $name = ucfirst(Locales::getName($key, $key));
-            $choices[$name] = $key;
+            $choices[$key] = $name;
         }
 
+        natcasesort($choices);
+
         $resolver->setDefaults([
-            'choices' => $choices,
+            'choices' => array_flip($choices),
             'label' => 'language',
             'choice_translation_domain' => false,
         ]);
--- a/tests/Form/Type/LanguageTypeTest.php
+++ b/tests/Form/Type/LanguageTypeTest.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Tests\Form\Type;
+
+use App\Configuration\LocaleService;
+use App\Form\Type\LanguageType;
+use Symfony\Component\Form\FormExtensionInterface;
+use Symfony\Component\Form\PreloadedExtension;
+use Symfony\Component\Form\Test\TypeTestCase;
+
+class LanguageTypeTest extends TypeTestCase
+{
+    public function testNaturalSortOfLanguageDisplayNames(): void
+    {
+        $form = $this->factory->create(LanguageType::class);
+
+        $options = $form->getConfig()->getOptions();
+
+        $this->assertArrayHasKey('choices', $options);
+        $this->assertIsArray($options['choices']);
+
+        $choices = array_values($options['choices']);
+
+        $this->assertEquals(
+            expected: ['de', 'de_CH', 'de_AT', 'en'],
+            actual: $choices,
+            message: 'Natural order of language names is not correct'
+        );
+    }
+
+    /**
+     * @return iterable<FormExtensionInterface>
+     */
+    protected function getExtensions(): iterable
+    {
+        $languageSettings = [
+            'de' => [
+                'date' => 'dd.MM.yy',
+                'time' => 'HH:mm',
+                'rtl' => false,
+            ],
+            'de_AT' => [
+                'date' => 'dd.MM.yy',
+                'time' => 'HH:mm',
+                'rtl' => false,
+            ],
+            'de_CH' => [
+                'date' => 'dd.MM.yy',
+                'time' => 'HH:mm',
+                'rtl' => false,
+            ],
+            'en' => [
+                'date' => 'M/d/yy',
+                'time' => 'h:mm a',
+                'rtl' => false,
+            ]
+        ];
+
+        return array_merge(parent::getExtensions(), [
+            new PreloadedExtension([
+                new LanguageType(new LocaleService($languageSettings))
+            ], [])
+        ]);
+    }
+}


kevinpapst:
Thanks 👍 

I actually never though about the order to be honest ... but I am not particular happy about all non ASCII languages being sorted at the end of the list now.

I find that more "discriminating" than the current approach based on the locales (which obviously makes no sense to any user). 

Probably that dropdown needs more content and we should simply prefix each entry with the language name in the current locale. Not sure if that is wise from a UX perspective... but it would simplify sorting^^

adrianrudnik:
Yeah, those get sorted to somewhere else. Right now they get injected by `AppExtension::setLanguageFormats` as `kimai.languages` parameter, coming from `services.yaml` `parameters.app_locales` which is a hand sorted string.

Should this be a hand sorted thing or was there any reason to reduce the dropdown to its current values?

Right now, using the [PHP 8.2.1 intl extension](https://onlinephp.io/c/7c618), there are 593 locales with an english name attached.

I'm not sure if translating that dropdown provides any benefit to the user. Once a wrong language is set, it can get pretty hard to find German, like coming from `tiếng Việt`, german would be `tiếng Đức` which is neither easy to find nor easy to google.

Maybe just using the english ones and sort them naturally? Right now the `intl` extension does not seem to required, so that would also be hurdle.

adrianrudnik:
Ah sorry, it is not hand sorted, just hand pasted I guess, the list gets ksorted later.

kevinpapst:
> Should this be a hand sorted thing or was there any reason to reduce the dropdown to its current values?

I generated the list of locales, but can't find the code that did it. I might have lost it when merging 2.0 into main.
And I really can't remember what I did to reduce the list. Unfortunately ... 😁 
I think I tried to generate a list of unique settings, not to cover all possible locales. 

You are right, translating the list really brings rally no benefit for the user. That's why it is like it is now.

I tried several ways of sorting the list, but all of them have an issue with the none Ascii languages (greek, chinese, vietnamese, japanese, russian, ukrainian, korean, hebrew, farsi, arabic, czech).
That's almost a third of all languages. 
Too many to ignore I would say.
But I am not sure if there is even a proper sort algorithm existing that can sort such a list properly, especially considering the RTL languages farsi/arabis/hebrew and chinese + korean + japanese.
Maybe it is okay to have them at the end of the list 🤷 
For me this is a less a technical issue (we could manually sort the list one) but more missing know-how on my end.

Maybe adding choice groups for locales would be a nice way to sort the list and then sort by region codes.
Or even manually adjust the list in the config file once?

adrianrudnik:
What approach is more to your liking? It goes over the scope of the ticket, just to feel the base of your future idea of it. Not sure if spending some more of your thoughts or time on this topic is worth it, or if I should prototype an approach.

__Keeping it as-is__

It really is of no utter importance, just fine details that might end up in technical debt later, either way.

__Enhance the manual aspect of it__

I do not know much about the `app_locales` parameter and what depends on it. The `translations.yaml` does not limit anything and there is no [enabled_locales](https://symfony.com/doc/current/reference/configuration/framework.html#enabled-locales) active as far as I can see.

__Make it system agnostic__

Drop manual stuff and make everything based on the `\Locale` v2 features available to PHP. The stuff injected into `LocaleService` can be generated in full from the intl package:

`date` can be generated by `(new IntlDateFormatter('en_UK', IntlDateFormatter::SHORT, IntlDateFormatter::NONE))->getPattern();` resulting in `M/d/yy` matching your manual config.

`time` can be generated by `(new IntlDateFormatter('en_UK', IntlDateFormatter::NONE, IntlDateFormatter::SHORT))->getPattern();` resulting in `h:mm a` matching your manual config.

`rtl` is the only one I have no clue about, I think its an app decision, not adhering to a specific ISO definition or?

Maybe filter out PRIVATE and GRANDFATHERED locales and just support every locale maintained somewhere else.

__Side note__

Just found https://www.php.net/manual/de/collator.sort.php, that might enable us to sort texts to the current locale, as a another approach.

kevinpapst:
The list is not (randomly) generated by hand. I found the code again:
https://github.com/kimai/kimai/blob/main/src/Command/RegenerateLocalesCommand.php
Only the locales that have at least a certain percentage of translations should be displayed.
Enabling all locales will just raise support issues about missing translations... no thanks 😁 

There are not many RTL languages in this world and they hard coded in the command.

Never stumbled upon `enabled_locales`, impossible to keep up with all the changes in SF.
But we can/should set it, might improve the caches a little bit.
The `app_locales` parameter is used in the route definition. So besides the aspect of fetching the correct formats, it is used to prevent that users can switch to a non-supported locale.

I already tried Collator while testing the PR, didn't change anything for me / asme result than your proposed changes. But please go ahead an re-test.

I really enjoy the technical discussion 👍 so don't get me wrong. 
I am just not sure if we can properly "fix" anything here regarding the sorting.

Lets check some apps... Slack does it like that:
Bildschirm­foto 2023-03-10 um 18 24 22
Sorting the non-ascii languages at the end of the list, but than the list is much easier anyway.
That's FreshRSS (seems they sort by locale as well):
Bildschirm­foto 2023-03-10 um 18 26 39
And finally Spotify: 
Bildschirm­foto 2023-03-10 um 18 31 44
Putting english on top, to make it simpler to find? Weird. 
Anyway, the rest seems to be sorted by locale as well (see Croation = hr and Hungarian = hu appear right after another)


We could do a sorting that is a bit more sophisticated and only sorts the region codes inside the locale (e.g. so all english are properly sorted). But I do not think that merging the currently proposed sorting is making the list of locales "better". Just my gut feeling obviously.

adrianrudnik:
`enabled_locales` could be a nice thing, if you want to enforce a support subset that you know that work. That again depends on what the locale solves.

Do you use the locale for translations only, or do you work with the region part (used for different date and time formattings) or even subtags like `fr_FR@currency=EUR`? If it is for translations only, you could even skip on things like `English (UK)` and `English (US)` if they are not maintained separatly and not present in `translations/`.

In your examples, looking over it:

- Slack is uses `Native language name (native region name)` as display name and sorts it, leaving the unicode characters at the end, where they belong, if you visit with a browser using `Accept-Language` being `de` or `en`.
- FreshRSS sorts by iso code of the language but shows native language name. `日本語` shown, sorted as `jp`. `Ελληνικά` shown, sorted as `el`.
- Spotify sorts the same way, `فارسی (Persian)` shown, `fa` sorted.

But coming back to the user issue, it is real. Just looking at `English` seems misleading, as the first subset is arab emirates, which I would expect to be under `United` as the united states. Also coming back to the topic above, is there any reason to have more than `English` there? I only see a single file translation file for it and I'm not sure you use anything else based on the `language` setting.

kevinpapst:
> or do you work with the region part (used for different date and time formattings)

That.

> you could even skip on things like English (UK) and English (US) 

Nope, (reason as above), e.g. 
en_GB => 24 hours
en_US => AM/PM

> Just looking at English seems misleading, as the first subset is arab emirates, which I would expect to be under United as the united states. 

Yes, true. I believe the simplest way forward is:
1. Sort the region codes under the locale, so "English (Austria)" comes before "English (United Arab Emirates)"
2. Touching the sort order of the main locales is not necessary, as it will not make anything better. At least I don't understand what and why.

adrianrudnik:
Ok, will modify the code accordingly!

kevinpapst:
Pleas re-open when you had the time to work on it.

Issue #3693, opened 2022-12-19
Team restrictions ignored at quick start activities


simarilius:
### Describe the issue

- In the quick start dropdown / or restart activities dropdowns, it is possible to restart activities in which the User is not (or no longer) part of the team, but has entries for this activity
 
 In our case, the "Abwesenheit" customer is managed by me and the plugin and the user should only be able to see their entries (that works) but not be allowed to create any. 

In the normal dropdown, the constraint works :)

### I already tried

- [X] I've read and searched [the documentation](https://www.kimai.org/documentation/).
- [X] I've searched for similar issues in this repository.
- [X] I've searched for similar issues in [the discussions](https://github.com/kevinpapst/kimai2/discussions).

### Kimai version

1.29.1 

### Which PHP version are you using?

7.4

### How do you run Kimai?

Self-Hosted

### Screenshots

![image](https://user-images.githubusercontent.com/7790769/208477176-0ef2e5c0-608b-45de-866a-9bcb6cc74791.png)
![image](https://user-images.githubusercontent.com/7790769/208477207-d21d09a5-230b-4288-8378-127cd6e77b3d.png)

kevinpapst:
Thanks for sharing 👍 

Will be fixed with 2.0.8

PR #3914, opened 2023-03-09
Release 2.0.8

kevinpapst:
## Description

- Javascript supports parsing negative durations (needs refactoring in the future)
- bump luxon dependency
- prevent API from triggering 2FA
- Include item name in delete dialog to make sure that the correct "thing" is deleted
- fixes permission issue in recent activities #3693

## Types of changes

## Checklist


omiT-84 (approved):
Negative duration seems to be working correctly.
Weekly view shows the correct substraction of the negative times.
afbeelding
Also confirmed export and invoice showing correct times.
I have not yet found the time to test the delete yet, but code looks sound.

omiT-84 (dismissed):
Approved.
I switched back and forward and the check went to comment instead of aprove in the last message. Sorry for the noise.

kevinpapst:
Thanks for testing 👍 
Not yet ready for merge, but good to know that it works!

PR #4173, opened 2023-07-07
[TASK] Use one letter as minimum for name

simonschaufi:
## Description
A time tracking system should not force me to have at least 3 letters for a project, activity or customer name if the system can handle it without breaking. This is especially important if you are importing data from other systems as there can be project names which are just 2 letters long (which is the case for me)

## Types of changes

## Checklist

### DIFF
--- a/src/Entity/Activity.php
+++ b/src/Entity/Activity.php
@@ -54,7 +54,7 @@ class Activity implements EntityWithMetaFields, EntityWithBudget
      */
     #[ORM\Column(name: 'name', type: 'string', length: 150, nullable: false)]
     #[Assert\NotBlank]
-    #[Assert\Length(min: 3, max: 150)]
+    #[Assert\Length(min: 1, max: 150)]
     #[Serializer\Expose]
     #[Serializer\Groups(['Default'])]
     #[Exporter\Expose(label: 'name')]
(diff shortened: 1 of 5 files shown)


kevinpapst:
Hi Simon, I love you for our common past and I understand this is frustrating.
But for now I have no plans to change that behavior, which can ultimately never be reverted and I am not yet there to go this road.
You have shown how easy it is to patch that to your likings and you / your customer can use a fork for this change.
I hope you understand that I am going to close this unmerged.

simonschaufi:
Can you please give me a good reason why you set this minimum? It doesn't make any sense to me to restrict people from  using any name they want. I understand that project names with one char doesn't make sense but if I insist on using project names like "a", "b" and "c", why not? Fun fact: there are companies which are called "X" :smile: Guess what I'm taling about :smile:

simonschaufi:
Hey Kevin, can you please give me a good reason why you set this minimum? The system won't break by using one letter names. It doesn't make any sense to me to restrict people from using any name they want. I understand that project names with one char doesn't make sense but if I insist on using project names like "a", "b" and "c", why not? Fun fact: there are companies which are called "X" :smile: Guess what I'm talking about :smile:

PR #4200, opened 2023-07-24
Fixes username field assertion to use field length of 180 instead of 60

### LINKED ISSUE #4199: Userprofile update not possible with username length >60
### Describe the issue

When length of the username in the database is >60 it is not possible to update the profile. An error occurs stating "This value is too long. It should have 60 characters or less.".

In our case, this happens because we use authentik as a SAML provider. The saml provider automatically maps the identifier provided by authentik (which is a 64 character base64 string) to the username. Therefore the user is created with a username > 60 characters in length but can never be edited because of the length restriction.

This can easily be fixed by setting the max-value in Entity/User.php:162 from the current 60 to the actual field length of 180

### I already tried

### Kimai version

2.0.28

### How do you run Kimai?

Docker

8.1

### Logfile

_No response_

### Screenshots

Screenshot 2023-07-24 at 12 53 08
Screenshot 2023-07-24 at 12 52 43

lukasassmann:
## Description

Fixes #4199

## Types of changes

## Checklist


kevinpapst:
Closing in favor of #4245 (which increases the max length to 64)

PR #4234, opened 2023-08-12
Weekly hours: warn on validation error - closes #4233

### LINKED ISSUE #4233: Warn when form validation error occurs when saving in  'Weekly hours'
### Describe the problem

I noticed that if I leave Weekly hours form open in my browser tab and enter hours later (for example tomorrow) and click save page is just refreshed but added hours are not saved. 

This behavior caused me to loose some hours I thought I logged. 

After debugging I figured out that this is caused by expired CSRF token.

Error popup which will clearly inform user that error happened, and data are not saved.

_No response_

### Screenshots

_No response_

stzoran1:
## Description
Adding an error message popup when quick entry form is not valid. This is preventing silent save failure.

## Types of changes
- UX improvement - fixes #4233

## Checklist

### DIFF
--- a/src/Controller/QuickEntryController.php
+++ b/src/Controller/QuickEntryController.php
@@ -181,7 +181,13 @@ final class QuickEntryController extends AbstractController
 
         $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
+        if ($form->isSubmitted()) {
+
+            if(!$form->isValid()){
+                $this->flashError('action.update.error', $form->getErrors());
+                return $this->redirectToRoute('quick_entry', ['begin' => $begin->format('Y-m-d')]);
+            }
+
             /** @var QuickEntryWeek $data */
             $data = $form->getData();
 


kevinpapst:
Thank you very much @stzoran1 for your help 👍 

That part of the code is really interesting:
I will have to look into it in depth I believe, because the form in general shows validation issues without problems.

I'll get back to you once I had the time, might need some days.

stzoran1:
@kevinpapst for easier and faster testing try loading Weekly hours page, add some new time in existing row and then from browser's Developer tools change value in input with id quick_entry_form__token. This will mimic expired CSRF token. Then click on Save button. You will notice page is reloaded but new entry is not saved, and all this happen without any warning. My solution is introducing warning so user is aware error happened and entry not saved.

kevinpapst:
This form is really a nightmare. Thanks for pointing out the missing errors!
I fixed the bug in a less workflow-disturbing way in this commit:
https://github.com/kimai/kimai/commit/636422e34293cd7155c22050b6a65faca85208b6

You will see the error and can re-submit the form afterwards.

Thanks for your contribution 👍

PR #4249, opened 2023-08-17
Fix: Change client_payload to client-payload in website.yaml

cngJo:
## Description
Fixes a small typo in the `website.yaml` workflow to ensure the `client-payload` is properly passed to the `workflow_dispatch`.

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/website.yaml
+++ b/.github/workflows/website.yaml
@@ -32,4 +32,4 @@ jobs:
           token: ${{ secrets.WEBSITE_ACCESS_TOKEN }}
           repository: kimai/www.kimai.org
           event-type: kimai_release
-          client_payload: '{"kimai_version": "$kimai_version"}'
+          client-payload: '{"kimai_version": "$kimai_version"}'


kevinpapst:
I applied several improvements and aded the validation logic:
https://github.com/kimai/kimai/blob/main/.github/workflows/website.yaml

Way too complicated for such a simple task of updating a single version number 🚀  but I like it anyway 😁

Thanks again!

cngJo:
Looks good, but wouldn't that technically still allow everyone to trigger the `workflow_dispatch` over at `kimai/www.kimai.org` and passing a valid kimai Version?

But I guess the risk is relatively small and so is the impact, since the worst thing that can happen is that someone discovers this and tries to spam pull requests. 

... maybe we shouldn't talk in a public issue tracker about this 🫣

I also like the solution, and complicated solutions for easy things are the most fun anyway, aren't they? 🤓

kevinpapst:
> Looks good, but wouldn't that technically still allow everyone to trigger the workflow_dispatch over at kimai/www.kimai.org and passing a valid kimai Version?

I am afraid that might be possible 🤷 
To reduce risk I added the validation logic there as well. 

> But I guess the risk is relatively small and so is the impact, since the worst thing that can happen is that someone discovers this and tries to spam pull requests.

And each run shows the person behind the triggering token: 

Bildschirmfoto 2023-08-17 um 23 46 58

So I could ban the user immediately.

> ... maybe we shouldn't talk in a public issue tracker about this 🫣

Hahaha, better talk about it and verify if there is a solution instead of hiding it

PR #4261, opened 2023-08-24
Change isWeekend() check to isWorkingDay() and fallback to Saturday/Sunday as weekend

### LINKED ISSUE #4260: [Visual glitch] Weekend days highlight in `Weekly hours` view are wrong if first day of week is Sunday
### Describe the issue

I think the title says it all. This is a pretty simple visual glitch.

When the user preference uses `Sunday` as first day of week, the `Weekly Hours` view highlight the last two columns, which are `Fri` and `Sat`.

### I already tried

### Kimai version

2.0.26

### How do you run Kimai?

Docker

8.1

### Screenshots

image

sharky98:
## Description
`format('w')` always returns `0` for Sunday and `6` for Saturday. The user configuration will not affect the result.

NOTE: I did not check the side-effects where "isWeekend()" is used elsewhere.

fixes #4260

## Types of changes

## Checklist
    - I edited the code removal from Github editor. I have no PHP dev environment.

### DIFF
--- a/src/Twig/LocaleFormatExtensions.php
+++ b/src/Twig/LocaleFormatExtensions.php
@@ -118,6 +118,7 @@ final class LocaleFormatExtensions extends AbstractExtension implements LocaleAw
 
         return $this->locale;
     }
+
     public function isWeekend(\DateTimeInterface|string|null $dateTime): bool
     {
         if (!$dateTime instanceof \DateTimeInterface) {
@@ -126,7 +127,7 @@ final class LocaleFormatExtensions extends AbstractExtension implements LocaleAw
 
         $day = (int) $dateTime->format('N');
 
-        if (!array_key_exists($day, $this->dayCache)) {
+        if (!\array_key_exists($day, $this->dayCache)) {
             $isWeekend = ($day === 6 || $day === 7);
 
             /** @var User|null $user */
--- a/src/Twig/LocaleFormatExtensions.php
+++ b/src/Twig/LocaleFormatExtensions.php
@@ -124,20 +124,6 @@ final class LocaleFormatExtensions extends AbstractExtension implements LocaleAw
 
         $day = (int) $dateTime->format('w');
 
-        if ($this->fdowSunday === null) {
-            /** @var User|null $user */
-            $user = $this->security->getUser();
-            if ($user !== null) {
-                $this->fdowSunday = $user->isFirstDayOfWeekSunday();
-            } else {
-                $this->fdowSunday = false;
-            }
-        }
-
-        if ($this->fdowSunday) {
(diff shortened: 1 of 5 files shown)


kevinpapst:
Can you elaborate, maybe it is too late and I don't get it...

I am not living in a place where Sunday is the first day of the week, so I might be mistaken.
But I thought that if Sunday is FDOW than your weekend is friday (5) and saturday (6).
While if Monday is FDOW your weekend is saturday (6) and sunday (0).

Hence the check for different numbers if a user has FDOW = sunday.

sharky98:
I am in Canada, where we use Sunday as the first day of the week. However, this has nothing to do with the weekend days, they are still Saturday and Sunday. This only affects how the calendar is shown.

https://www.timeanddate.com/calendar/days/first-day-of-the-week.html

https://www.wheniscalendars.com/wp-content/uploads/2022/10/august-2023-calendar-printable-by-month-wk-a4-ls-color-wheniscalendars-com_Page_08-1024x724.png

https://calendarkart.com/wp-content/uploads/2023/02/August-2023-Monday-Start-Calendar-1536x1086.jpg

kevinpapst:
Quote from the article:

> For this reason, many Muslim countries had their weekend on Thursday and Friday and would start the week on Saturday.

Which would be okay as third option, but then:

> This changed in 2000 when Jordan shifted to the Friday-Saturday weekend. Since then, most countries in the Middle East and other Muslim-majority countries have switched to the Friday-Saturday weekend. They start their work week on Sunday.

Argh, which is like it is currently handled. Considering that most paying customers do not come from middle east countries I don't mind switching to your proposal. But why can't it be easy for once ???

sharky98:
Yeah, I understand. In a future feature, maybe having a configurable off day? Knowing that more and more people are working 4 days a week and such. It should cover all cases where weekends are not Saturday and Sunday.

kevinpapst:
That is actually a brilliant idea 👍  

Instead of concentrating on the classical approach of a 5-day week we should check whether someone has a working-hour configuration (the `Employment contract`) for the given day: 
Bildschirmfoto 2023-08-31 um 10 19 54

Which should be fairly easy, but we need to reintroduce the `Security` which you just removed, because we need to access the current user.

src/Twig/LocaleFormatExtensions.php
```diff
@@ -115,30 +117,28 @@ public function getLocale(): string
 
         return $this->locale;
     }
-
+
```
kevinpapst:
```suggestion

```

kevinpapst (approved):
Thanks for the idea and input @sharky98 👍

PR #4387, opened 2023-10-31
Simplify code and change bad wording

simonschaufi:
## Description
Simplify code and change bad wording

## Types of changes

## Checklist

### DIFF
--- a/src/Utils/StringHelper.php
+++ b/src/Utils/StringHelper.php
@@ -30,7 +30,7 @@ final class StringHelper
     public static function sanitizeDDE(string $text): string
     {
         // see #3189
-        if (\strlen($text) === 0) {
+        if ($text === '') {
             return $text;
         }
 
@@ -43,7 +43,7 @@ final class StringHelper
         }
 
         if ($sanitize) {
-            // trying to prevent fucking Microsoft "feature" DDE
+            // trying to prevent horrible Microsoft "feature" DDE
             $text = "' " . $text;
         }
 


src/Utils/StringHelper.php
```diff
@@ -30,20 +30,20 @@ public static function ensureMaxLength(?string $string, int $length): ?string
...
         if (\in_array($text[0], self::DDE_PAYLOADS)) {
             $sanitize = true;
-        } elseif (stripos($text, 'DDE') !== false) {
+        } elseif (\stripos($text, 'DDE') !== false) {
             $sanitize = true;
         }
 
         if ($sanitize) {
-            // trying to prevent fucking Microsoft "feature" DDE
+            // trying to prevent Microsoft "feature" DDE
```
kevinpapst: that is not bad wording, that is one of the the worst "features" that MS has ever produced in his "office" products. I hate it and wasted way too many precious hours of my life, because ... well read about DDE injection and you will fall from your chair.

src/Utils/StringHelper.php
```diff
@@ -20,8 +20,8 @@ public static function ensureMaxLength(?string $string, int $length): ?string
             return null;
         }
 
-        if (mb_strlen($string) > $length) {
-            $string = mb_substr($string, 0, $length);
+        if (\mb_strlen($string) > $length) {
```
kevinpapst: The codestyles are purely managed by php-cs-fixer.
The build will fail if they are broken, no need for manual adjustments.
You can use `composer codestyle-fix` to apply them.

src/Utils/StringHelper.php
```diff
@@ -30,20 +30,20 @@ public static function ensureMaxLength(?string $string, int $length): ?string
     public static function sanitizeDDE(string $text): string
     {
         // see #3189
-        if (\strlen($text) === 0) {
+        if ($text === '') {
```
kevinpapst, line 33: as the method receives a string, this should never be null. what does it change?

kevinpapst:
Oh and Thank you for your PRs! :)

src/Utils/StringHelper.php
simonschaufi: No, i had to adjust it manually because this command didn't do it! Strange that PHP CS Fixer wants to revert it back to `mb_strlen` but all other places (`\in_array`...) are all namespaced and there it works. What's different there?

src/Utils/StringHelper.php
simonschaufi: This is a micro optimization because comparing pure strings is faster than calling a method. The result is the same.

src/Utils/StringHelper.php
simonschaufi: I know that MS has done some really eval things already but i would rather call it "horrible" or whatever but not "fu***"

src/Utils/StringHelper.php
kevinpapst: There is a link somewhere in the php-cs-fixer rule, that leads to a place where the explanation is better than I could ever repeat, but here is a link to start with: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/3048

src/Utils/StringHelper.php
kevinpapst: Yeah true, latter one is actually to good to bring it up in one sentence with Microsoft 🙈 👿

src/Utils/StringHelper.php
simonschaufi: ah ok, quoting https://github.com/symfony/symfony/pull/32717#issuecomment-514664927: Only some functions can be optimized but not all. Learned something new ;)

src/Utils/StringHelper.php
simonschaufi: Code should not be a place to express your feelings even though you "wasted way too many precious hours of my life" which I feel very sorry about.

kevinpapst: Why? This is my code.

simonschaufi:
Dear Kevin, I would have reverted the wording back if you would have told me and then you could still merge the micro optimization but just closing without commenting is a little bit rude. First you say thank you for this PR and now you just close it...

src/Utils/StringHelper.php
simonschaufi: Sure, if you want to leave it there, you can tell me to change it back.

kevinpapst:
I am closing, because there is no benefit in this PR and I want to prevent both of us wasting more time than necessary.

Issue #4503, opened 2023-12-20
Docker test script fails (currently deactivated)


kevinpapst:
### Describe the problem

The last release failed partially to build Docker images:
https://github.com/kimai/kimai/actions/runs/7241462128

Additionally many tags weren't pushed:
<img width="759" alt="Bildschirmfoto 2023-12-20 um 16 58 21" src="https://github.com/kimai/kimai/assets/533162/0e2dc7d8-93ce-46ab-9812-1a0a62a428f2">

Unfortunately many of them are prod tags. 
So many are now missing the 2.6.0 build: https://hub.docker.com/r/kimai/kimai2/tags

Might be fixed by #4504

My biggest issue is that I do not know how to test this.

kevinpapst:
@tobybatch my potential fix did not work. i tried to reproduce locally, wasn't able to. This is thrown in Docker build "Test lite":

<img width="758" alt="Bildschirmfoto 2023-12-26 um 16 51 02" src="https://github.com/kimai/kimai/assets/533162/e3d268a3-6cd0-45d2-b02f-69606e531d71">


Now if I change the BASTABASE_URL locally and remove the entire cache before calling `bin/console kimai:version` I am also able to produce an error (but not the same one).

With `DATABASE_URL=sqlite:///%kernel.project_dir%/var/data/kimai.sqlite`

<img width="771" alt="Bildschirmfoto 2023-12-26 um 16 54 36" src="https://github.com/kimai/kimai/assets/533162/aa0a40b6-bf79-44c2-b1b8-df0bf8a1c1f3">

With `DATABASE_URL=`

<img width="1018" alt="Bildschirmfoto 2023-12-26 um 16 55 18" src="https://github.com/kimai/kimai/assets/533162/4cb81027-ca4c-4407-936f-79e0a73db8ba">

With `DATABASE_URL=mysql://kimai:kimai@127.0.0.1:3307/kimai2xxxxxxx?charset=utf8mb4` (invalid port and missing serverVersion!)
<img width="767" alt="Bildschirmfoto 2023-12-26 um 16 56 27" src="https://github.com/kimai/kimai/assets/533162/3767c133-ac48-4852-913a-f36bf6d159d4">

No matter what I do, an error only pops up, if Doctrine tries to query for the server version. 
And I do not understand why that should happen, as I changed the DATABASE_URL to include all required information.

What is that test-lite.sh doing anyway? Can we remove it and re-run the scripts manually?
[...]

PR #4541, opened 2024-01-08
Hotfix to re-enable PHP/Kimai test

Not-Code:
## Description
This hotfix reenable /opt/kimai/bin/console kimai:version during CI/CD, closing issue #4503 and takes in consideration https://github.com/kimai/kimai/pull/4504/commits/48d86d41b284396e6fff277aad81305b28d39c4e. It should be analyzed if DATABASE_URL defined in the Dockerfile is still mandatory.

A later PR may fix this issue in a more complete way (for example spinning up a real mysql database), but in the meantime it permits to execute the kimai:version test.

## Types of changes

## Checklist

### DIFF
--- a/.docker/test-lite.sh
+++ b/.docker/test-lite.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 if [ -z "$DATABASE_URL" ]; then
-  export DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
+  DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
 fi
 
 /opt/kimai/bin/console kimai:version
--- a/.docker/test-lite.sh
+++ b/.docker/test-lite.sh
@@ -1,13 +1,14 @@
 #!/bin/sh -e
 
-# Test PHP/Kimai
-# This does not work currently, as a real database is required
-# see https://github.com/kimai/kimai/issues/4503
-#/opt/kimai/bin/console kimai:version
-#if [ $? != 0 ]; then
-#  echo "PHP/Kimai not responding"
-#  exit 1
-#fi
+if [ -z "$DATABASE_URL" ]; then
+  export DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
+fi
+
+/opt/kimai/bin/console kimai:version
+if [ $? != 0 ]; then
+  echo "PHP/Kimai not responding"
+  exit 1
+fi
 
 # Test FPM CGI
 if [ -f /use_fpm ]; then


.docker/test-lite.sh
```diff
@@ -1,13 +1,14 @@
...
-#/opt/kimai/bin/console kimai:version
-#if [ $? != 0 ]; then
-#  echo "PHP/Kimai not responding"
-#  exit 1
-#fi
+if [ -z "$DATABASE_URL" ]; then
+  export DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
+fi
+
+/opt/kimai/bin/console kimai:version
```
kevinpapst, line 7: `-z` tests the current process for an existing variable?

We should be able to do it like that as well:
```suggestion
DATABASE_URL=mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40 /opt/kimai/bin/console kimai:version
```

kevinpapst: oh and thank you 😄

Not-Code: -z checks if the variable is unset. The basic idea, from the script point of view, is:

if someone gives me a DATABASE_URL than that's ok, you should know better. Otherwise a DATABASE_URL is necessary so I'm filling a fake one.

Kimai already has a similar logic in the [startup ](https://github.com/kimai/kimai/blob/8eeb9c120a674532a75622b0ba322e4bd1f50a04/.docker/startup.sh) script (i.e. the entrypoint of normal execution):

```
  # Parse sql connection data
  if [ ! -z "$DATABASE_URL" ]; then
    DB_TYPE=$(awk -F '[/:@]' '{print $1}' <<< "$DATABASE_URL")
    DB_USER=$(awk -F '[/:@]' '{print $4}' <<< "$DATABASE_URL")
    DB_PASS=$(awk -F '[/:@]' '{print $5}' <<< "$DATABASE_URL")
    DB_HOST=$(awk -F '[/:@]' '{print $6}' <<< "$DATABASE_URL")
    DB_PORT=$(awk -F '[/:@]' '{print $7}' <<< "$DATABASE_URL")
    DB_BASE=$(awk -F '[/?]' '{print $4}' <<< "$DATABASE_URL")
  else
    DB_TYPE=${DB_TYPE:mysql}
    if [ "$DB_TYPE" == "mysql" ]; then
      export DATABASE_URL="${DB_TYPE}://${DB_USER:=kimai}:${DB_PASS:=kimai}@${DB_HOST:=sqldb}:${DB_PORT:=3306}/${DB_BASE:=kimai}"
    else
      echo "Unknown database type, cannot proceed. Only 'mysql' is supported, received: [$DB_TYPE]"
      exit 1
    fi
  fi
```

But if you prefer to override the value in any case, that's also possible. From a quick verification on the codebase the test-lite.sh appears to be called only by the CI/CD process.

Not-Code: I verified the suggestion and it works. I would still keep the **if** verification (my impression is that I want to permit override from docker run command or whatever) so I would suggesto to modify it like this:

```suggestion
if [ -z "$DATABASE_URL" ]; then
  DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
fi
/opt/kimai/bin/console kimai:version
```
But last word is yours. Tell me how you prefer it and I will update it accordingly. Thank you.

kevinpapst (approved):
Thank you @Not-Code 👍

.docker/test-lite.sh
kevinpapst: Thanks for your detailed testing and response. I like that!
Sounds as if your initial suggestion was already the best way to move forward. 
Not sure whether we want to use that script in another fashion, but its good to have the option to do so.

Thanks for all your input here. Very much needed, especially in the docker context, where I lack knowledge!

Not-Code: Thank you too. I feel that we still got a better answer than my initial suggestion because removing the export command makes more sense for this script.
I will update the PR with an additional commit later today.

kevinpapst: Right?! In my experience discussions in PRs usually lead to better solutions and/or knowledge transfer. Both results are good 😄

kevinpapst: Oh, and I will wait for your change before merging ...

kevinpapst (approved):
Lets see if it works on the next release 😁

PR #4586, opened 2024-01-24
split docker to use base image for faster builds

tobybatch:
## Description
Building fpm and apache, for dev and prod, for amd and arm = 2 * 2 * 2 builds and the x-platform building was *very* slow. In my forked repo it was just timing out. In this PR we switch to use a intermediate build (tobybatch/kimai-base:{fpm-apache}) which does not need to rebuilt every time kimai is built. This speeds up the kimai built and the base image will ony need to be rebuiolt when we move PHP versions.

I'll start on the docs now.

## Types of changes

## Checklist


kevinpapst:
The link for reference: https://github.com/tobybatch/kimai-base/tree/main

Everything that makes it faster is highly appreciated.

I might have asked that before, but why don't we use some pre-built PHP image from an "official" source? 
I can't believe that we have to roll that by ourselves.

kevinpapst:
Maybe @Not-Code wants to chime in, I am really the last person who should review Docker changes 😁

Not-Code:
Yes, gladly. I will post a complete opinion later this evening.

Not-Code:
Ok, so my first impression is that if we approve this PR we won't actually change the total build time of Kimai (from an E2E perspective it will remain exactly the same), but we will simply add a new dependency outside the control of Kimai project that will be mandatory for Kimai and will essentially split the build process in 2 parts (and so it will speed up Kimai build process because only half build will be performed).

Now below my main concerns.

**TRUST**
I now understand that @tobybatch is the main maintainer of the Kimai docker images, but I would consider someone who is simply looking at the available docker images for kimai for the first time. It will find the image maintained by Kimai and the image maintained by LinuxServer. If that was someone like me, and looking at the docker layers found `FROM tobybatch/` it would probably move to another repository.
The Kimai project has automatic trust because it's the image published by the project maintainer, not so much tobybatch/. So if this is the direction that we want to move on, I think that, at least, the intermediate image should be part of the Kimai repository.

**SECURITY**
I think that it's in the best interest of Kimai project to understand the lifecycle of the intermediate image. While it's true that the intermediate image doesn't need to be rebuilt every time that kimai is build, it should be at least tied to the release cycle of the php base image.
With the current process we are 100% sure that every time that kimai is built we are getting the latest version of the php container. For the 8.2-apache-bookworm it would be this:

image

(you can see that the digests are the same so 8.2-apache-bookworm is currently just another tag for 8.2.15-apache-bookworm).

If we suppose that the new intermediate image was built like 7 months ago we would instead ship the below image:

image

with many additional security issues.

---

Now, talking on generic ideas that will speed up the Kimai E2E build process.

**CACHING**
There's nothing that we can do to improve the build during npm/apache or different architecture targets. The only wasted time is when in the same server and architecture we are building dev/prod targets. If we found a way to cache the intermediate layers between the dev/prod build targets we should get the same improvement (I'm always talking of E2E time) of the current PR.

**LAYER OPTIMIZATION**
I'm not sure about this point because I tend to think that when someone does something complicated it does that for a good reason.
But looking at this:

```
FROM ${BASE}-php-ext-base AS php-ext-gd
RUN docker-php-ext-configure gd \
        --with-freetype && \
    docker-php-ext-install -j$(nproc) gd

# php extension intl : 15.26s
FROM ${BASE}-php-ext-base AS php-ext-intl
RUN docker-php-ext-install -j$(nproc) intl

# php extension ldap : 8.45s
FROM ${BASE}-php-ext-base AS php-ext-ldap
RUN docker-php-ext-configure ldap && \
    docker-php-ext-install -j$(nproc) ldap

# php extension pdo_mysql : 6.14s
FROM ${BASE}-php-ext-base AS php-ext-pdo_mysql
RUN docker-php-ext-install -j$(nproc) pdo_mysql

# php extension zip : 8.18s
FROM ${BASE}-php-ext-base AS php-ext-zip
RUN docker-php-ext-install -j$(nproc) zip

# php extension xsl : ?.?? s
FROM ${BASE}-php-ext-base AS php-ext-xsl
RUN docker-php-ext-install -j$(nproc) xsl

# php extension redis
FROM ${BASE}-php-ext-base AS php-ext-redis
RUN yes no | pecl install redis && \
    docker-php-ext-enable redis

# php extension opcache
FROM ${BASE}-php-ext-base AS php-ext-opcache
RUN docker-php-ext-install -j$(nproc) opcache

...

COPY --from=php-ext-xsl /usr/local/etc/php/conf.d/docker-php-ext-xsl.ini /usr/local/etc/php/conf.d/docker-php-ext-xsl.ini
COPY --from=php-ext-xsl /usr/local/lib/php/extensions/no-debug-non-zts-20220829/xsl.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/xsl.so
# PHP extension pdo_mysql
COPY --from=php-ext-pdo_mysql /usr/local/etc/php/conf.d/docker-php-ext-pdo_mysql.ini /usr/local/etc/php/conf.d/docker-php-ext-pdo_mysql.ini
COPY --from=php-ext-pdo_mysql /usr/local/lib/php/extensions/no-debug-non-zts-20220829/pdo_mysql.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/pdo_mysql.so
# PHP extension zip
COPY --from=php-ext-zip /usr/local/etc/php/conf.d/docker-php-ext-zip.ini /usr/local/etc/php/conf.d/docker-php-ext-zip.ini
COPY --from=php-ext-zip /usr/local/lib/php/extensions/no-debug-non-zts-20220829/zip.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/zip.so
# PHP extension ldap
COPY --from=php-ext-ldap /usr/local/etc/php/conf.d/docker-php-ext-ldap.ini /usr/local/etc/php/conf.d/docker-php-ext-ldap.ini
COPY --from=php-ext-ldap /usr/local/lib/php/extensions/no-debug-non-zts-20220829/ldap.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/ldap.so
# PHP extension gd
COPY --from=php-ext-gd /usr/local/etc/php/conf.d/docker-php-ext-gd.ini /usr/local/etc/php/conf.d/docker-php-ext-gd.ini
COPY --from=php-ext-gd /usr/local/lib/php/extensions/no-debug-non-zts-20220829/gd.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/gd.so
# PHP extension intl
COPY --from=php-ext-intl /usr/local/etc/php/conf.d/docker-php-ext-intl.ini /usr/local/etc/php/conf.d/docker-php-ext-intl.ini
COPY --from=php-ext-intl /usr/local/lib/php/extensions/no-debug-non-zts-20220829/intl.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/intl.so
# PHP extension redis
COPY --from=php-ext-redis /usr/local/etc/php/conf.d/docker-php-ext-redis.ini /usr/local/etc/php/conf.d/docker-php-ext-redis.ini
COPY --from=php-ext-redis /usr/local/lib/php/extensions/no-debug-non-zts-20220829/redis.so /usr/local/lib/php/extensions/no-debug-non-zts-20220829/redis.so
COPY --from=php-ext-opcache /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini  /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini
```

I wondered many times if it is really impossible to collapse those operations into a single RUN layer and 2 COPY commands.

Not-Code:
Expanding on the caching point. Maybe we can refer to this [Cache management with GitHub Actions](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache) and adding something like:

          cache-from: type=registry,ref=kimai/kimai2:buildcache-${{ matrix.server }}
          cache-to: type=registry,ref=kimai/kimai2:buildcache-${{ matrix.server }},mode=max

to the docker/build-push-action would do the trick.
To verify if docker/build-push-action@v5 is mandatory and variable substitution is supported, but you should get the idea.

kevinpapst:
What about using this instead: 
https://hub.docker.com/_/php/
https://github.com/docker-library/php

They have Apache and FPM images and ways to install extensions without these copy commands.

.github/workflows/docker.yaml
```diff
@@ -49,8 +49,6 @@ jobs:
                       kimai/kimai2:${{ matrix.server }}-${{ matrix.type }}
                       kimai/kimai2:${{ matrix.server }}-${{ env.kimai_version }}-${{ matrix.type }}
                   push: true
-            - name: Test Lite
-              run: docker run --rm --entrypoint /assets/test-lite.sh kimai/kimai2:${{ matrix.server }}-${{ matrix.type }}
```
kevinpapst, line 53: Delete `assets/test-lite.sh` as well?

Not-Code:
> What about using this instead: https://hub.docker.com/_/php/ https://github.com/docker-library/php
> 
> They have Apache and FPM images and ways to install extensions without these copy commands.

We are already using them. Taking as an example apache (the same is valid for nginx) the base image of _${BASE}-php-ext-base_ is the one you suggested:

`FROM php:${PHP_VER}-apache-bookworm AS apache-php-ext-base`

All the stuff you see is called multi-stage build. The scope of using multi-stage build is to reduce the total image of the final image.

For example, let's assume that we were shipping a C software, in that case you would need the C compiler to compile it after downloading the source from Github.
Still the final image would just needed the compiled executable and neither ther source or the C compiler.
So in this case you can use a multi-stage build where in the first image you compile the software and in the second you just move the compiled executable.

This is what is happening in Kimai. We are basically starting a new image for every extension, compile it, and moving it to the base image with the COPY command.
The point is that doing this 8 time is probably really slow when a QEMU simulator needs to be used for a non-native architecture. Compacting all the commands in just one RUN layer (if possible) should speed up (in theory) the non-native compilation.
Still going in this direction would require regression testing of all extensions.

So I was asking for confirmation because if things were split in this way instead of using a single layer there should be a reason, I think.

.github/workflows/docker.yaml
Not-Code: With all the important changes of this PR this was something that I wanted to highlight later, but I agree that the test should not be removed.
It actually works and gives Kimai container some coverage, so no real reason to remove it (at least for me).

Not-Code:
I tried to do a build with layer optimization (source at the end of the comment) and see what would happen to the build time. I run a multi-architecture build (amd64/arm64) targeting apache/prod image. Those are the build times on my machine.

**OPTIMIZED DOCKERFILE**
real	48m39.702s
user	67m56.631s
sys	6m44.219s

**ORIGINAL DOCKERFILE**
real	52m45.118s
user	75m10.467s
sys	8m0.296s

So we would have a small improvement but nothing amazing (and this is just a build time test, I didn't verify that things are actually working).

Most of the time is lost during the non-native architecture (in my case arm64) because all the php extensions that we are using are compiled during the build (and the compilation is very slow).

The native architecture build is instead pretty fast.
**OPTIMIZED DOCKERFILE (only amd64)**
real	5m25.636s
user	4m7.626s
sys	1m45.318s

So I will wait an answer to my original concerns.

```
FROM ${BASE}-php-ext-base AS php-ext
RUN docker-php-ext-configure gd --with-freetype && \
    docker-php-ext-configure ldap && \
    docker-php-ext-install -j$(nproc) gd intl ldap pdo_mysql zip xsl opcache && \
    yes no | pecl install redis && \
    docker-php-ext-enable redis

...

COPY --from=php-ext \
     /usr/local/etc/php/conf.d/docker-php-ext-xsl.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-pdo_mysql.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-zip.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-ldap.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-gd.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-intl.ini  \
     /usr/local/etc/php/conf.d/docker-php-ext-redis.ini \
     /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini \
     /usr/local/etc/php/conf.d/

COPY --from=php-ext \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/xsl.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/pdo_mysql.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/zip.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/ldap.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/gd.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/intl.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/redis.so \
     /usr/local/lib/php/extensions/no-debug-non-zts-20220829/
```

kevinpapst:
Toby was maintaining the Kimai docker images [for years under his repo](https://github.com/tobybatch/kimai2), we just moved it over here. I can't recall that anyone ever complained or asked. 

But I agree that having everything under the kimai namespace would be better. I guess it wouldn't be a problem to place the "intermediate Dockerfile" under this repo as well, right?

Testing is easier with single calls, but we know they work, right? And if I understand it correct, less layers are better, so merging the calls seems like a good idea. 

please ignore my other question regarding the php base image, I demonstrated once again how much I do not know about docker 😁

Not-Code:
It is understandable, from my point of view, that no one complained or asked, because I would never do that and would simply walk away. As a user who receives work that others release into the community as a free software, I have no say in how they manage the process. However, it is in my power to decide whether that is good enough for me or not.

So, while I cannot say that every user out there would think like me, I can at least share my perspective.

Now, if you look at Kimai registry (that as you said just moved here), we can see that it has over 5M pulls:

image

Instead one of the images that is necessary to use nginx and is out there from more than 3 years has only over 100K pulls:

image

and this happens while there is more interest in nginx over apache in the last years:

image


So either people don't like Kimai with nginx or the fact that the image that Kimai is using for nginx has the following issues:

1. Non official repository
2. Basically unmantained (no releases over 3 years)
3. With many security issues

is preventing them in using nginx.

Below you can see a scan with Trivy:

```
2024-01-27T10:12:00.617Z	WARN	This OS version is no longer supported by the distribution: alpine 3.11.6
2024-01-27T10:12:00.617Z	WARN	The vulnerability detection may be insufficient because security updates are not provided

docker.io/tobybatch/nginx-fpm-reverse-proxy:latest (alpine 3.11.6)
==================================================================
Total: 56 (UNKNOWN: 0, LOW: 2, MEDIUM: 16, HIGH: 34, CRITICAL: 4)
```

So yeah. If the idea is to transform the current image of Kimai in what happened with `tobybatch/nginx-fpm-reverse-proxy:latest` I don't think that this is in the best interest of Kimai project.

Also @tobybatch already applied the change because the current release of version 2.10 on docker hub has been overriden by him and it seems like it is already using his base image.

image

tobybatch:
@Not-Code You're right, the problem I had is testing the build pushed the images. If we trigger a build in the main branch, it will overwrite those images with the monolithic build.

The tobybatch nginx image is only used in the example docker compose file, it was never meant to go into production.

I'll look through @Not-Code comments later but I was already thinking that we should fold the base image into this repo too.

I'll update the PR layer today.

tobybatch:
> I might have asked that before, but why don't we use some pre-built PHP image from an "official" source? I can't believe that we have to roll that by ourselves.

We need to build addition libraries for things like GD, but the base images are built from the core php image. It's building those extension for x-platform that's _really_ slow.

.github/workflows/docker.yaml
tobybatch: That test ran `./bin/console kimai:version` and started apache/fpm. The test was run after the push so was pretty much useless other than to notify that something was wrong.

I'm happy to trust that the apache configs in the base images are tested upstream. I guess we could have hosed the apache config but it's not changed in 3 years.

tobybatch:
> Ok, so my first impression is that if we approve this PR we won't actually change the total build time of Kimai (from an E2E perspective it will remain exactly the same), but we will simply add a new dependency outside the control of Kimai project that will be mandatory for Kimai and will essentially split the build process in 2 parts (and so it will speed up Kimai build process because only half build will be performed).

The x-platform builds for multi arch images take > 2 hours on the github runners. We have 3000 minutes and we are using ~1000 minutes per build. Unless we move to a hosted runner solution we *can't* keep the build as it is.

tobybatch:
> **TRUST** I now understand that @tobybatch is the main maintainer of the Kimai docker images, but I would consider someone who is simply looking at the available docker images for kimai for the first time. It will find the image maintained by Kimai and the image maintained by LinuxServer. If that was someone like me, and looking at the docker layers found `FROM tobybatch/` it would probably move to another repository. The Kimai project has automatic trust because it's the image published by the project maintainer, not so much tobybatch/. So if this is the direction that we want to move on, I think that, at least, the intermediate image should be part of the Kimai repository.

I suggest we add the build stage image into this repo. Originally I didn't want to do that as it would bloat up KP's repo with more docker build files. I think a later comment by @kevinpapst suggest that too. I'll update the PR with a new docker file and add an action for use to build that image whenever we want. Maybe there is a way to detect when the upstream images release new versions.

tobybatch:
> **SECURITY** I think that it's in the best interest of Kimai project to understand the lifecycle of the intermediate image. While it's true that the intermediate image doesn't need to be rebuilt every time that kimai is build, it should be at least tied to the release cycle of the php base image. With the current process we are 100% sure that every time that kimai is built we are getting the latest version of the php container. For the 8.2-apache-bookworm it would be this...

That's a really good point. We should think about how to address this. Kind of the same response as above

tobybatch:
> **CACHING** There's nothing that we can do to improve the build during npm/apache or different architecture targets. The only wasted time is when in the same server and architecture we are building dev/prod targets. If we found a way to cache the intermediate layers between the dev/prod build targets we should get the same improvement (I'm always talking of E2E time) of the current PR.

Unless I've missed something (and I'm really open to a solution I haven't seen) the containers for the runners are destroyed after each run. That link you added deserves some investigation, I found this:

Selection_386

And didn't fancy relying on experimental features.

tobybatch:
> 
> **LAYER OPTIMIZATION** I'm not sure about this point because I tend to think that when someone does something complicated it does that for a good reason. But looking at this:
> 
> ```
> FROM ${BASE}-php-ext-base AS php-ext-gd
> RUN docker-php-ext-configure gd \
>         --with-freetype && \
>     docker-php-ext-install -j$(nproc) gd
> 

..... CLIPPPED......

> COPY --from=php-ext-opcache /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini  /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini
> ```
> 
> I wondered many times if it is really impossible to collapse those operations into a single RUN layer and 2 COPY commands.

When using buildx these images should build concurrently. If we inline them they will build consecutively. Well that's my understanding anyway.

tobybatch:
> Expanding on the caching point. Maybe we can refer to this [Cache management with GitHub Actions](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache) and adding something like:
> 
> ```
>       cache-from: type=registry,ref=kimai/kimai2:buildcache-${{ matrix.server }}
>       cache-to: type=registry,ref=kimai/kimai2:buildcache-${{ matrix.server }},mode=max
> ```
> 
> to the docker/build-push-action would do the trick. To verify if docker/build-push-action@v5 is mandatory and variable substitution is supported, but you should get the idea.

The link you found does indeed suggest that this isn't experimental. I'll give it a try.

.github/workflows/docker.yaml
tobybatch: and that command is run during the build so the build will fail if it doesn't complete.

https://github.com/kimai/kimai/blob/main/Dockerfile#L291

tobybatch:
> I tried to do a build with layer optimization (source at the end of the comment) and see what would happen to the build time. I run a multi-architecture build (amd64/arm64) targeting apache/prod image. Those are the build times on my machine.
> 
> **OPTIMIZED DOCKERFILE** real 48m39.702s user 67m56.631s sys 6m44.219s
> 
> **ORIGINAL DOCKERFILE** real 52m45.118s user 75m10.467s sys 8m0.296s
> 
> So we would have a small improvement but nothing amazing (and this is just a build time test, I didn't verify that things are actually working).
> 
> Most of the time is lost during the non-native architecture (in my case arm64) because all the php extensions that we are using are compiled during the build (and the compilation is very slow).
> 
> The native architecture build is instead pretty fast. **OPTIMIZED DOCKERFILE (only amd64)** real 5m25.636s user 4m7.626s sys 1m45.318s
> 
> So I will wait an answer to my original concerns.
> 
> ```
> FROM ${BASE}-php-ext-base AS php-ext
> RUN docker-php-ext-configure gd --with-freetype && \
>     docker-php-ext-configure ldap && \
>     docker-php-ext-install -j$(nproc) gd intl ldap pdo_mysql zip xsl opcache && \
>     yes no | pecl install redis && \
>     docker-php-ext-enable redis
> 
> ...
> 
> COPY --from=php-ext \
>      /usr/local/etc/php/conf.d/docker-php-ext-xsl.ini \

.... CLIPPED ....

>      /usr/local/lib/php/extensions/no-debug-non-zts-20220829/
> ```

I get roughly the same numbers on my local as well, but mine is AMD not ARM. If you run it many times you'll get varying set of results. I've had it complete as fast as 40 minutes and take as long 75 minutes but none of that actually matters as it's the speed on the github runners that actually matter and there the build can take up to 3 hours.

tobybatch:
So this produces a number of actions, all for me I think.

 - [x] I'll move the intermediate image into this repo. @kevinpapst I'll _hide_ it in the .docker folder off the root
 - [x] Add a workflow dispatch (manual workflow) to do the same.
 - [x] I'll do 1 & 2 with the old nginx image but as I pointed out that was only included to give an example of how to start the FPM image. 
 - [ ] Address this when I update the docker documentation. It's still that problem with people using the FPM image without knowing enough about what they are doing.
 - [ ] Investigate the caching layer, it if't stable, then try and use it.

@Not-Code Thank you for your time and such an in depth analysis of the build process.

kevinpapst:
I can't help myself, but whoever is using FPM standalone, should really know what to do. 
I'm impressed that the image got 100k downloads.

I addressed the nginx issue by [adjusting the docs](https://www.kimai.org/documentation/docker-compose.html):
- moved Apache to the top of Docker Compose examples
- added a fat warning that your image is only an examples and NOT for meant for production.

kevinpapst:
@tobybatch did we scare you away? 😁

tobybatch:
@kevinpapst not at all.

I pushed some changes, moved the dependant docker into this repo and kind of expected @Not-Code to comment again.

Other than that, we're just deciding on whether this is a good idea.

We're a little stuck if we don't do this and want to offer multiplatform images as the build time is just too high.

I'm looking at self hosted runners for a client in my day job. That may be an option in the future.

Not-Code:
Sorry for the delay but I've been busy with work lately.
Overall, I think Toby's latest changes are a good answer to my previous concerns. I will try to summarize below in points.

---

> When using buildx these images should build concurrently. If we inline them they will build consecutively. Well that's my understanding anyway.

Agreed, if during buildx build all those images are build concurrently then it makes sense to keep things as they are. I'm using podman build with manifest and, in my case, everything was built sequentially.

> assets/test-lite.sh ... and that command is run during the build so the build will fail if it doesn't complete.

I didn't notice that it was already used in the Dockerfile. Agreed on this too.

> kimai/kimai-nginx new image

It's a good idea to add an nginx-fpm image to kimai repository. It will help new users who want to use nginx instead of apache.

> The x-platform builds for multi arch images take > 2 hours on the github runners. We have 3000 minutes and we are using ~1000 minutes per build. Unless we move to a hosted runner solution we can't keep the build as it is.

This is the critical point. I don't have a good enough solution for this, but if we are using too many minutes per build a solution should actually be found (and I would not drop support for ARM because it will damage the project distribution).
So my general idea is that we should go ahead with the PR and change it, if necessary, in the future.

---

To improve the current PR, I would suggest the following changes if you both agree.

- To make sure the base image is built and not add it to anyone's responsibility (I don't think that we really care when it is built, we only care that it is built al least 1 time in a period) we could change the trigger of that worflow to a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).
- I would move the build of the new container kimai/kimai-nginx from the base image to the current kimai workflow for the following reasons: it makes sense, at least to me, that it's lifecycle is bounded to the kimai image (it's an end user image), and it should not take so much time even in cross platform compilation.
- Is this label correct in the new nginx image `LABEL "com.example.vendor"="Neontribe"`?

tobybatch:
* [x] I'll think about the base image. I wish there was a hosted trigger for "some one else's repo has changed". 
 * [x] I have (in the last 4 years) found a better way of doing this, I'll need to update the docker compose example file and include a nginx config file
 * [x] We can drop that label. It's the name of my company but I cashed out of that last Xmas. :)

kevinpapst:
> I cashed out of that last Xmas. :)

Uuuuh, Congratulations! That explains your long absence 😁

tobybatch:
@Not-Code I've switched up the nginx container to load a prep'd conf file rather than use a container with one built in.  Tidied out the Neontribe ref and put the base image on a schedule.

@kevinpapst will want a sign off before he merges this.

tobybatch:
> > I cashed out of that last Xmas. :)
> 
> Uuuuh, Congratulations! That explains your long absence 😁

Yeah, now I'm an employee, not a director. Good and bad. Mainly Good :smile:

kevinpapst:
Can we not add the docker-compose file in this PR? 
I already see some points I'd like to address if we are going to promote some kind of "best practice" here. 
This will end in a long discussion and this issue is already long...

tobybatch:
OK, I've pulled the docker compose file. I'll look at the docs site ASAP. 

I think we do want a `docker-compose.yml` in the repo eventually, it's the simplest quick start solution.

kevinpapst:
> I've pulled the docker compose file. 

Thanks! 

> I think we do want a docker-compose.yml in the repo eventually, it's the simplest quick start solution.

I agree, but your example was a dev image IMO and other stuff we might not want to promote as best-practice (e.g. phpmyadmin) for production usage. And you know that people do not read docs if they don't have to: they see a docker-compose file and just use it... 

So the PR is ready for a final look by @Not-Code ?

Not-Code:
I logged in tonight with the intention of clicking approve, but unfortunately I think that .docker/Dockerfile.nginx was accidentally deleted in commit 59edd086fb624989c77f98172cb9e5159ad6b427.
Other than that everything seems fine. When .docker/Dockerfile.nginx is restored I will try a manual build, just to be safe, and approve 😄

Not-Code:
Adding on the comment above.
This schedule [* * 5 * *] seems uncorrect:

image

This is the correct form (assuming that 04:00 AM is fine):

image

[source](https://crontab.guru/)

tobybatch:
@Not-Code The nginx docker was deliberately removed. We aren't shipping it as it leads people who don't understand it using it in prod envs.

tobybatch:
@Not-Code  Fixed crontab.  The nginx docker file was removed deliberate as people who don't understand how to use it end up using tit for prod envs. 

I'll provide an example in the docs when I get time to update those.

.github/workflows/docker-base.yaml
```diff
@@ -0,0 +1,53 @@
...
+                  tags: |
+                    kimai/kimai-base:${{ matrix.server }}
+                  push: true
+
+
+            - name: Build
+              uses: docker/build-push-action@v4
+              with:
+                  context: .docker
+                  file: .docker/Dockerfile.nginx
```
kevinpapst: @tobybatch but the nginx file is still used here

.docker/Dockerfile.kimai-base
```diff
@@ -0,0 +1,210 @@
...
+    cgi-fcgi -bind -connect 127.0.0.1:9000 || exit 1
+
+
+
+###########################
+# apache base build
+###########################
+
+FROM php:${PHP_VER}-apache-bookworm AS apache-base
+ARG KIMAI
```
Not-Code: This ARG should be removed otherwise the following warning is raised during the build:
`WARN[0000] missing "KIMAI" build argument. Try adding "--build-arg KIMAI=<VALUE>" to the command line`

.docker/nginx_default.conf
```diff
@@ -0,0 +1,34 @@
+server {
```
Not-Code: With the removal of the nginx fpm container is this configuration file still needed?

Not-Code (commented):
@tobybatch
I tried to do a local build and the build was fine. However you decide to close the last 2 points I raised, I will give automatic approval.

tobybatch:
@Not-Code I have fixed those two issues

kevinpapst:
This thread is super long, so I don't read everything again... but why don't we build the new base image on every release? We could first trigger the base image and then trigger the release image once the base image is done. Or did we want to prevent that?

Anyway: considering that @Not-Code gave an automatic approval, I think this is ready to merge?!

tobybatch:
I think it can be merged. 

There are issues with the time it takes to rebuild the base image every time you tag. We'll run out of minutes. I'm looking at self hosted runners for a business client. That may be a solution.

kevinpapst:
Thank you 👍 

But afaik open source projects have no limit on minutes.
It is only counted on private repos (that's what I see in my profile).

SaturnusDJ:
Good improvement regarding build speed.

Building 2.14.0 on very crappy hardware single arch AMD64:
Old: 04:16 min / 753 MB.
New: 01:56 min / 853 MB.

PR #4624, opened 2024-02-09
Use duration instead of durationBillable for time budget

danielegobbetti:
## Description
Use the ``duration`` field for Working Hours calculations, since ``durationBillable`` field is already used for the Billable calculations

I am unsure if this should be considered a bug-fix or a breaking change (rule https://xkcd.com/1172/ applies :-) ), but for my use case this is a fix, hence marking as such.

The behavior before this change is not considering working hours that are not billable in the statistics against the time budget for an activity, as per the following screenshot:
before

With the changes applied, the same activity looks as follows:
after

I consider this an improvement since the "Working Hours" are a subset of the "Time Budget", and the "Billable" hours are a subset of the "Working Hours".

## Types of changes

## Checklist

### DIFF
--- a/templates/embeds/budgets.html.twig
+++ b/templates/embeds/budgets.html.twig
@@ -32,7 +32,7 @@
                         </tr>
                         {% set totalPercentReached = 100 %}
                         {% if entity.timeBudget > 0 %}
-                            {% set totalPercentReached = (stats.durationBillable / (entity.timeBudget / 100)) %}
+                            {% set totalPercentReached = (stats.duration / (entity.timeBudget / 100)) %}
                         {% endif %}
                         <tr>
                             <td>{{ durationTrans|trans }}</td>
@@ -64,7 +64,7 @@
                         <tr>
                             <th colspan="3" class="text-nowrap text-end">
                                 {% if totalPercentReached < 100 %}
-                                    {{ 'stats.percentUsedLeft'|trans({'%percent%': totalPercentReached|number_format(2), '%left%': (entity.timeBudget - stats.durationBillable)|duration}) }}
+                                    {{ 'stats.percentUsedLeft'|trans({'%percent%': totalPercentReached|number_format(2), '%left%': (entity.timeBudget - stats.duration)|duration}) }}
                                 {% else %}
                                     {{ 'stats.percentUsed'|trans({'%percent%': totalPercentReached|number_format(2)}) }}
                                 {% endif %}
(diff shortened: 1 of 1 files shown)


kevinpapst:
The good old heat-bar, classic!

Why would non-billable duration count towards the budget?

1. The title in the first graph is actually wrong, I forgot to fix that one.
2. Are you trying to achieve a calculation for a pre-paid activity? Is that why all the work is non-billable?
3. That breaks post-paid calculation, which is only based on billable hours. 

The best solution would be, if that behavior would be configurable OR we add another budget type (so we cannot only choose `monthly` but also `pre-paid`) which would then change the calculation as in your PR here.

danielegobbetti:
Let me begin answering your question:

> Why would non-billable duration count towards the budget?

My thought was that every hour should count towards the time budget, whereas billable hours should count towards the monetary budget.

I might have made the wrong assumption, so I'll share how I mapped the UI fields to my mental model.

As both a monetary value and a time amount can be entered in the budget field, I thought of using this fields as follow:
- monetary value: amount of hours * hourly rate as agreed with the customer
- time amount: amount of hours I am planning to actually work on this activity (including "internal" tasks that I can't/don't want to charge the customer for)

Of course this comes from my understanding of the labels on the GUI, and it's just one point of view (and if one of the strings is not accurate, my mental model is even more probably wrong).


Practically speaking the money budget is the quote as approved by the customer, the time budget isn't (might be equal, might be bigger). This also allows for fixed price activities which require preparation, as per example below.

<details>
<summary>Practical example, with made up numbers: a (half-day) workshop by a customer for which 500 EUR are charged, including the preparatory works and wrap up documentation *click to expand*</summary>

Currently I map this to an activity which has a monetary budget of 500 EUR and a time budget of 10 hours, but the actual workshop will last maybe 3, maybe 4 hours.
Then I add a price (within this activity) for 500 EUR and mark this as "fixed price".
I enter all the preparatory hours as non billable; when I hold the workshop I enter one billable activity; I enter all the followup hours for wrapping up as non billable.

</details>

I'm not sure if you would consider this kind of activity a "pre-paid" one, but I might try a more rigorous modeling of it to see how it could fit in the data model (I have to say that I can fit my way of working very well in kimai as it is, I considered this a misbehavior of the GUI only)

kevinpapst:
Fair enough, if you don't want to sign a CLA. 
I opened #4839 so the info won't get lost.

danielegobbetti:
> Fair enough, if you don't want to sign a CLA.

That's the point, yes. (And to be clear: this is your project and I respect the decision you made. And exactly for this respect I preferred not to bring up an argument over CLA)

kevinpapst:
Decisions need to be questioned, so if you have links that discuss CLAs, I am happy to read arguments against them. 

You can always share patch files, so I can see our ideas.

Issue #4591, opened 2024-01-26
Missing ARM64/AArch64 docker images


KhalilSantana:
### Describe the problem

It seems the image `kimai/kimai2` does not yet have an ARM64 build, resulting in the container not starting on ARM64 machines (see below). Please consider creating multi-arch images

```
%      docker run --rm --name kimai-test \
         -ti \
         -p 8001:8001 \
         -e "DATABASE_URL=mysql://kimai:kimai@${HOSTNAME}:3399/kimai?charset=utf8mb4&serverVersion=5.7.40" \
         kimai/kimai2:apache
Unable to find image 'kimai/kimai2:apache' locally
apache: Pulling from kimai/kimai2
2f44b7a888fa: Pull complete 
3a95dcec6035: Pull complete 
e22afa33f327: Pull complete 
1aa61ea11ee8: Pull complete 
ea67ae2bde33: Pull complete 
ff5356c6468f: Pull complete 
5e91cb51e886: Pull complete 
935e88b21d5f: Pull complete 
8202e883ada1: Pull complete 
be40d3507d33: Pull complete 
5fac4c1a6ab6: Pull complete 
03fcb105f703: Pull complete 
87c6b0d7fa4d: Pull complete 
5a4d39ff5c27: Pull complete 
358cde348a7d: Pull complete 
3167e857138f: Pull complete 
acc7d11bbaba: Pull complete 
f72772c907f1: Pull complete 
b6ea6dfe6a27: Pull complete 
98952f5e41c7: Pull complete 
d93f9609a123: Pull complete 
ef31493d1f32: Pull complete 
712cdd660b1a: Pull complete 
aa0271954c32: Pull complete 
a0b874f9dfd9: Pull complete 
73b91592966b: Pull complete 
6a696c0687ad: Pull complete 
852aa803a153: Pull complete 
101e3edab22d: Pull complete 
65c43769547b: Pull complete 
76b356115116: Pull complete 
1b239230c995: Pull complete 
0a1a5df59ab8: Pull complete 
ab21ecd70cc7: Pull complete 
fef353c7e062: Pull complete 
e8b5ca5073dd: Pull complete 
40d60e29e1a2: Pull complete 
43d427d09ee3: Pull complete 
316c2f97476e: Pull complete 
Digest: sha256:903c85e1c01b96aafe0a12e57febf687013c1bf8b331801fb274828a3672b272
Status: Downloaded newer image for kimai/kimai2:apache
[...]

PR #4640, opened 2024-02-14
✨ add OCI docker labels

pxlfrk:
## Description
This PR solves #4638 by adding the missing OCI labels dynamically in the CI workflow-file.

Labels have been added as a workflow-variable to keep maintenance efforts low. Whereas possible, dynamic variables have been used.

I also updated the used github-actions to their recent version, most of them were based on Node 16 which will be [deprecated soon](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/).

## Types of changes

## Checklist
- [x] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation)) --> _not applicable_

### DIFF
--- a/translations/email.es.xlf
+++ b/translations/email.es.xlf
@@ -46,6 +46,22 @@
         <source>absence_created_supervisor_subject</source>
         <target state="translated">Nueva ausencia | %absence_user% | %absence_type%</target>
       </trans-unit>
+      <trans-unit id="wwLx._J" resname="absence_created_supervisor_message" xml:space="preserve">
+        <source>A new absence was created for %absence_user%.
+
+%absence_type% - %absence_comment%
+
+%absence_list%
+
+Please review them at: %approval_url%</source>
+        <target state="translated">Se ha creado una nueva ausencia para %absence_user%.
+
+%absence_type% - %absence_comment%
+
+%absence_list%
+
+Por favor, compruébelo en: %approval_url%</target>
+      </trans-unit>
     </body>
   </file>
 </xliff>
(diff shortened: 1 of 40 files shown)


Not-Code:
Thanks, for the interesting PR.

I dwelt upon this a bit and, as far as I understand, we are adding the new action docker/metadata-action with the scope to label and tag images in conjunction with docker/build-push-action, to handle OCI dynamic label values.

Looking at the action documentation, it seems to me, that the primary scope of the action is to define with more flexibility (and retrieving git info) the tags/labels of the docker/build-push-action:

```
        name: Build and push
        uses: docker/build-push-action@v5
        with:
          context: .
          push: ${{ github.event_name != 'pull_request' }}
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
```

but in kimai we are prevented from using the new action at 100%, because we are building with a multistage build and that isn't currently supported out of the box by docker/metadata-action.

If I got everything right, I don't see a specific reason to add a new component in the build that will add additional dependencies, increase the complexity, and add maintenance overhead only for the following fields:

```
org.opencontainers.image.created=${{env.NOW}}
org.opencontainers.image.version=${{github.ref_name}}
org.opencontainers.image.revision=${{github.sha}}
```

I got the point of the PR, but I would just add static labels in the Dockerfile. The creation date can be retrieved directly from the image or layers, the version is already part of the current tags (and someone using latest is not really interested in the version) and the sha should be uninmportant to end users with the version information.

Instead if the new action is able to drive docker/build-push-action targets and solve #4591 too, then I'm cheering 😄 .

PR #4768, opened 2024-04-10
Add command to stop time exceeding timesheets

binsky08:
## Description
Currently there's no option to automatically stop timesheets, you forgot to stop manually.
If you have something like a company rule for a maximum work time per timesheets (or you're just forgetful :D), the implemented command helps enforcing that.

---

This pull request adds a command that stops timesheets with a duration longer than the specified amount of hours (as command argument), if they do not yet have an end value set.

---

Example usage would be a cron job that runs the command once a hour like:
`bin/console kimai:timesheet:stop-time-exceeding 12` to disallow timesheets longer than 12 hours.

## Types of changes

## Checklist

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -543,6 +543,11 @@ parameters:
             path: src/Command/ResetTestCommand.php
 
         -
+            message: "#^Parameter \\#1 \\$value of function floatval expects array|bool|float|int|resource|string|null, mixed given\\.$#"
+            count: 1
+            path: src/Command/TimesheetStopTimeExceeding.php
+
+        -
             message: "#^Access to an undefined property SimpleXMLElement\\|false\\:\\:\\$file\\.$#"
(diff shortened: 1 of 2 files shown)


kevinpapst (changes requested):
Thanks for taking the time to share your solution 👍 

I have several issues with this.
A command can only be used from a cron-jobs, an API method would be better to make it more versatile.

Then there are already both an API method that allows to find active timesheets and one to stop a timesheet.
There is a system configuration for the max allowed duration of a timesheet.

Can you please move the logic to a service method, and use the [existing system configuration](https://github.com/kimai/kimai/blob/main/src/Configuration/SystemConfiguration.php#L367) as default option (instead of a required duration input) we can talk .

kevinpapst:
Please re-open if you want to make the adjustments.

Issue #3846, opened 2023-02-13
Request for "SAML group" to "Kimai Team" mapping feature


TSmigielski:
### Describe the problem

I have integrated a SSO service with SAML for my company's Kimai instance. And everything works flawlessly, except for the Team mapping. Which would be an amazing feature to have.

### Describe the solution you'd like

I've looked through the code a bit, and as far as I can tell, this feature isn't implemented yet, which is a bummer, hence the "Request".

kevinpapst:
You can help testing this: #4770 to see if it works. 
I have no matching setup and I am not able to verify the solution from @the-glu

TSmigielski:
Unfortunately I've left that company in November last year. I wish you the best of luck though, Kimai was one of my favorite tools to use.

kevinpapst:
Hahaha, all the best to you! 

Closing here, as the PR is already open.

PR #4770, opened 2024-04-10
First implementation of SAML teams mapping

the-glu:
## Description
Hello,

This is a basic implementation of #3846: It add possibility to map kimai's team from a SAML attribute.
It's also allow setting the leadership 'attribute' of a user in a team.

This PR currently lacks tests and documentations, I wanted first to be sure the implementation direction is ok :)

Feedback is welcome, my PHP is a bit rusty ^^'

===> PR for documentation: https://github.com/kimai/www.kimai.org/pull/423

## Types of changes

## Checklist


kevinpapst:
Thanks @the-glu for giving this a try 👍 

Code-wise it looks good (haven't checked details, but the overall view is great), . 
Maybe we should add a big try & catch - so login doesn't fail in case of issues with the teams.

Now as you mentioned: tests are missing and documentation.
Can you share how you tested this?

I'd like to try it, but I don't have a Keycloak setup, so I need either Azure AD or Google Workspace. 
How does one setup Teams on IDP side?

mastertIT:
@the-glu what's the syntax in the local.yaml config for this?

from the code it looks like another line in the mapping section?

would like to test this.

the-glu:
Hello,

Great, I will add tests and documentation as soon as possible :)

I tested with our internal SSO but it should works with any SSO. If you like I can make you a Keycloak available somewhere with some basic configuration.

@mastertIT 

For the syntax in the local.yaml config for testing:

```
kimai:
    saml:
        provider: google
        activate: true
        title: Login with SAML
(...)
        roles:
(...)
        teams:
            resetOnLogin: true
            attribute: groups
            mapping:
                - { saml: testgrsp, kimai: 7, leader: false }
(...)
```

Saml attribute is the value in the attribute list (that you specify on the line before), kimai is the team id on the kimai side (edit a team and check the URL for the ID), leader is if the user should be leader or not of the team :)

mastertIT (approved):
works in my lab

kevinpapst:
An approval is too early 😁 but great feedback 🚀

src/Configuration/SamlConfiguration.php
```diff
@@ -50,6 +50,21 @@ public function isRolesResetOnLogin(): bool
         return $this->configuration->isSamlRolesResetOnLogin();
     }
 
+    public function getTeamsAttribute(): ?string
+    {
+        return $this->configuration->getSamlTeamsAttribute();
+    }
+
+    public function getTeamsMapping(): array
```
kevinpapst, line 58:
```suggestion
    /**
     * @return array<int, array{'saml': string, 'kimai': int, 'leader': string}>
     */
     public function getTeamsMapping(): array
```

Please check if that is correct.
Then add the same for the Interface and then you can remove the PHPStan ignores as well.
The PHPStan baseline can only be reduced, but no new ones should be added.

src/Saml/SamlProvider.php
```diff
@@ -91,6 +93,71 @@ private function hydrateUser(User $user, SamlLoginAttributes $token): void
...
 
+        $teamAttribute = $this->configuration->getTeamsAttribute();
+        $teamMapping = $this->configuration->getTeamsMapping();
+
+        // extract user teams from a special saml attribute
+        if (!empty($teamAttribute) && $token->hasAttribute($teamAttribute)) {
+            $teamMap = [];
+            foreach ($teamMapping as $mapping) {
+                $field = $mapping['kimai'];
+                $team = $this->teamRepository->findById($field)[0];
```
kevinpapst, line 104: I do not like repository magic methods. Please change that to:
```suggestion
                $team = $this->teamRepository->findOneBy(['id' => $field]);
                if ($team !== null) {
                ...
```

src/Saml/SamlProvider.php
```diff
@@ -91,6 +93,71 @@ private function hydrateUser(User $user, SamlLoginAttributes $token): void
...
+                if (\array_key_exists($teamName, $teamMap)) {
+                    $teams[] = $teamMap[$teamName];
+                }
+            }
+
+            // If we need to reset teams on login, remove team that user should not have
+            if ($this->configuration->isTeamsResetOnLogin()) {
+                // For all teams of the user, we check that the user should be present
+                foreach ($user->getTeams() as $userTeam) {
+                    $shouldBeInTeam = false;
```
kevinpapst, line 123: Do we even need the boolean? Can't you replace the `break` by a `continue 2`?

kevinpapst (changes requested):
Thanks! A few small things to check.

mastertIT:
is there any movement on the changes needed to this?

can i make the changes and push them upstream?

the-glu:
> is there any movement on the changes needed to this?
> 
> can i make the changes and push them upstream?

Yes sorry, I don't have much time those day to work on this :/

I gave you access to the upstream repository so you may do the fixes if you have time.

Thanks :)

PR #4802, opened 2024-04-17
Docker: Fix TLS certificate verification for LDAP authentication

### LINKED ISSUE #4796: LDAP login with TLS broken again
### Describe the problem

It seems that PR #4586 reverted the fix provided in #4598. Therefore, the login via LDAP + TLS is not possible anymore beginning with Release 2.14.0. The login credentials are always rejected as incorrect.

Manually running `apt-get install libldap-common` inside the container fixes the issue. We are using the apache version of the image (`kimai/kimai2:apache-2.15.0-prod`).

### Describe your setup and add your Docker compose file (redact your credentials)

- OS: Debian 11.9
- Docker version: 26.0.1
- Docker compose version: 2.11.1

### Command used to run the container

_No response_

abrain:
Fixes #4796 

## Description
This PR adds back the Debian package `libldap-common` into the Apache base Docker image. This allows the verification of TLS certificates when using LDAPS for authentication.

## Types of changes

## Checklist

### DIFF
--- a/.docker/Dockerfile.kimai-base
+++ b/.docker/Dockerfile.kimai-base
@@ -144,6 +144,7 @@ RUN apt-get update && \
         bash \
         haveged \
         libicu72 \
+        libldap-common \
         libpng16-16 \
         libzip4 \
         libxslt1.1 \


kevinpapst:
Probably I am confused because I don't exactly understand in which layer which libs are installed and especially why  ... 
but doesn't that belong to `fpm-php-ext-base` and `apache-php-ext-base` ?

abrain:
The layers derived from `${BASE}-php-ext-base` are used to install php extensions, but are not merged into the final image. Only certain files are copied over. The `libldap-common` library, however, needs to be present in the final image.

But, it is true, that technically only the file `/etc/ldap/ldap.conf` is required to be there. The rest of the package is only documentation (see [file list](https://packages.debian.org/bookworm/all/libldap-common/filelist)). So it would be possible to do a similar approach and install it in the `apache-php-ext-base` layer and only copy over the .conf file. But I'm not sure that would gain much. Maybe @tobybatch has an opinion on that.

I haven't checked the LDAPS situation with the `fpm` image in practice (maybe I can switch our test system to that image real quick). But the issue appeared because the Apache-based `php` image dropped this library when moving to Debian Bookworm. So I assume only the Apache image is affected.

kevinpapst:
See, I am clueless when it comes to Docker... I need @tobybatch for the rescue 😁

kevinpapst (approved):
OK, I am gonna merge, so this issue is fixed.

In case someone has the same issue with nginx, I hopefully remember what to do 😁 

Thanks @abrain 👍

kevinpapst:
see also https://github.com/kimai/kimai/pull/4598 and https://github.com/docker-library/php/issues/1194

PR #4780, opened 2024-04-14
Release 2.16

kevinpapst:
## Description

- Remove `view_own_timesheet` permission check from Wizard controller
- Adds a string formatter for the general export function - thanks @popalltheshells
- Adds a twig function to sanitize data against DDE attacks - thanks @popalltheshells
- Fix just another doctrine deprecation
- allow to use english format for DateRanges even if user uses different locale
- added macro to simplify cross-links to search results
- cross-link "filter timesheets" view from weekly reports
- activate Taiwanese locale
- re-use existing validation translation from Symfony
- improve min_day and max_day handling
- added missing redirect after saving API token
- make sure that access tokens expiry date is in the future - fixes #4808 
- allow to delete api tokens for other users - fixes #4809
- use core date filters to show last_usage and expiry date for API tokens
- update all frontend dependencies to latest minor version
- upgrade to chart.js 4 - closes #4815 
- fix loading `local.yaml` as last config
- prepare icon refactoring: move animation class to template, use icon function instead of deprecated filter
- bump composer packages (fixes: translations are missing if marked as "needs-translation")
- remove all expired session on logout
- disable CSRF on logout again, causing too many unnecessary problems
- only match request if a `Bearer` token is passed, previously empty Authorization header were matched as well
- use rate limiter to reduce risk of Session prediction attack attempts by allowing only 250 requests with invalid Session IDs in 1 hour from one IP (thanks to Ignatius Michael from DeepCove Cybersecurity AKA @popalltheshells for raising awareness)

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [x] I verified that my code applies to the guidelines (`composer code-check`)
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [x] I agree that this code is used in Kimai (see [license](https://github.com/kimai/kimai/blob/main/LICENSE))

PR #4815, opened 2024-04-24
Default-Farben hinzugefügt

stevenbuehner:
Es ist zwar nicht getestet, sollte aber funktionieren.
Wäre es möglich die Default-Farbpalette von ChartJs aufzunehmen?

### DIFF
--- a/assets/chart.js
+++ b/assets/chart.js
@@ -23,6 +23,7 @@ import {
     Legend,
     Title,
     Tooltip,
+	Colors,
 //    SubTitle
 } from 'chart.js';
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
That doesn't work yet, as Kimai is still on Chart.js 3.x and the default colors are available since 4.x:
https://www.chartjs.org/docs/latest/general/colors.html#default-color-palette

I'll have to upgrade first: 
https://www.chartjs.org/docs/latest/migration/v4-migration.html

kevinpapst:
Upgraded to Chart.js 4 and added default colors in my PR for the next version. See #4780 - thanks for the hint.

PR #4840, opened 2024-05-06
New System Configuration for Timesheet's description field optional or mandatory.

adambarath:
## Description
Add opportunity users to make timesheet's description field mandatory.

## Types of changes

## Checklist


kevinpapst:
Thank you. The problem with this approach is: you cannot enforce it. People can simply click "restart" (or use the calendar or the weekly hours) and it won't work.

See https://www.kimai.org/documentation/plugin-custom-content.html#javascript

Bildschirmfoto 2024-05-06 um 14 37 26

kevinpapst:
Thanks for taking the time to share that code. 
Your approach has a few issues that make it incompatible with the core. 
So I am not going to merge that.

But: you can write a small plugin that adds this feature via a global form listener.
Or you simply use the shared javascript code.

PR #4853, opened 2024-05-13
Update Util.php

habibhaidari1:
Proposal to fix rate rounding issues on invoices. Since the invoices show the duration as a decimal number with two decimal points.

Example: 
assuming that the hourly rate is 34.34 and the duration is 25 minutes Current algorithm calculates 1500/3600*34.34 = 14.31 € proposal algorithm calculates 0.42 (number on the invoice) * 34.34 = 14.42 €

It's dirty but works for me. I don't expect that to be part of the repo

## Description

## Types of changes

## Checklist

### DIFF
--- a/src/Timesheet/Util.php
+++ b/src/Timesheet/Util.php
@@ -27,7 +27,7 @@ final class Util
      */
     public static function calculateRate(float $hourlyRate, int $seconds): float
     {
-        $rate = $hourlyRate * ($seconds / 3600);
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thank you 👍 

I will work on the invoices in the near future and then come back to this topic here.

kevinpapst:
It took quite a while, but I am not refactoring the invoice module and this change is quite an important point.
Thanks for sharing @habibhaidari1 🚀

PR #4592, opened 2024-01-27
use skopeo to preserve multi-arch lists during the tag operation

Not-Code:
## Description
This is an attempt to resolve #4591 using skopeo instead of docker: skopeo has the capability to clone list of images.

ubuntu-latest runner that Kimai is actually using comes with skopeo already [preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) and skopeo should fallback to credentials created in step _Login to DockerHub_.

Before approving this PR, @tobybatch or @kevinpapst should verifiy that everything is fine because I'm not able to test a complete workflow. 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I verified that my code applies to the guidelines (`composer code-check`)
- [ ] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [x] I agree that this code is used in Kimai (see [license](https://github.com/kimai/kimai/blob/main/LICENSE))

PR #4856, opened 2024-05-14
feat: stop multiarch images being lost on re tagging docker images

tobybatch:
## Description
When we re-tag docker images we loose the arm64 versions of the multi arch

## Types of changes

## Checklist


tobybatch:
@Not-Code do you see a draw back over this? It's a stack overflow fix, I've not used it before.

https://stackoverflow.com/questions/71470604/re-tagging-multi-arch-docker-image

kevinpapst:
Probably not the right place, but have you seen that the base images built in about 18 minutes? So far less problematic than the expected two hours: https://github.com/kimai/kimai/actions/workflows/docker-base.yaml

tobybatch:
@kevinpapst I'll build an all in one workflow that we can test for timing. I'm looking at caching in builds at work ATM as well.

kevinpapst (approved):
I never used any of those actions 🤷  ... so 👍 😁 

The only things that jump in my eyes are: 
- we have a lot of redundant tags
- why do we install Install buildx in the step before at line 22 if we don't use it afterwards?

kevinpapst:
Is this ready? Having the arm64 version would be great actually and I want to create a new release.

Does this superseed #4592 ?

kevinpapst:
ping @tobybatch is this WIP or ready for merge?

tobybatch:
> I never used any of those actions 🤷 ... so 👍 😁
> 
> The only things that jump in my eyes are:
> 
> * we have a lot of redundant tags
> * why do we install Install buildx in the step before at line 22 if we don't use it afterwards?

I've kept all the old tags that we have every used. I guess we could slash the number of tags to:
 
 * fpm
 * apache
 * apache-dev
 * fpm-dev

And

 * alias fpm -> latest

The old images are so that people who have used a tag won't break when we don't update the images. I guess we could drop them now.

buildx provides the x-platform support in the `docker build` command, it's invisible to the actual command.

tobybatch:
This should be ready to merge. I'll update the branch.

kevinpapst (approved):

kevinpapst:
Thanks, I'll answer in the other issue later

PR #4943, opened 2024-06-29
fixed case of `AS` keyword for consistency in Dockerfile

chinmaypurav:
## Description
When building the image I realised an inconsistency with the naming with `AS` keyword in dockerfile.

It was flagged off like this

image

I had received similar warning when using Sonarqube analysis in one of my projects. It SHOULD BE uppercase.

## Types of changes

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,10 +33,10 @@ RUN rm -r tests
 
 
 # FPM base
-FROM kimai/kimai-base:fpm as fpm-base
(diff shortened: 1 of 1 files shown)


kevinpapst:
Hm ... https://docs.docker.com/reference/dockerfile/#format

> The instruction is **not** case-sensitive. However, convention is for them to be UPPERCASE to distinguish them from arguments more easily.

Anyway: thanks for the PR!!

PR #4976, opened 2024-07-16
Add support for special characters in DATABASE_URL

robjuz:
## Description
Add support for special characters in DATABASE_URL

## Types of changes

## Checklist

### DIFF
--- a/.docker/dbtest.php
+++ b/.docker/dbtest.php
@@ -1,9 +1,9 @@
 <?php
-$DB_HOST = $argv[1];
-$DB_BASE = $argv[2];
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):

kevinpapst:
Thanks @robjuz 👍

kevinpapst:
@robjuz did you test this one? Now after having a second look because of #4997 I wonder if we should have used `urlencode` instead.

robjuz:
I may not have tested the ‘+‘ charakter. I will double check it and provide a PR

PR #5000, opened 2024-07-29
better support for special characters in DATABASE_URL

robjuz:
## Description
This is a fix for the previous implemetation with handling for the '+' charakter

## Types of changes

## Checklist

### DIFF
--- a/.docker/dbtest.php
+++ b/.docker/dbtest.php
@@ -1,6 +1,6 @@
 <?php
-$DB_HOST = urldecode(str_replace('+', '%2B', $argv[1]));
-$DB_BASE = urldecode(str_replace('+', '%2B', $argv[2]));
+$DB_HOST = urldecode($argv[1]);
+$DB_BASE = urldecode($argv[2]);
 $DB_PORT = $argv[3];
 $DB_USER = urldecode(str_replace('+', '%2B', $argv[4]));
 $DB_PASS = urldecode(str_replace('+', '%2B', $argv[5]));
--- a/.docker/dbtest.php
+++ b/.docker/dbtest.php
@@ -2,8 +2,8 @@
 $DB_HOST = urldecode($argv[1]);
 $DB_BASE = urldecode($argv[2]);
 $DB_PORT = $argv[3];
-$DB_USER = urldecode(str_replace('+', '%2B', $argv[4]));
-$DB_PASS = urldecode(str_replace('+', '%2B', $argv[5]));
+$DB_USER = urldecode($argv[4]);
+$DB_PASS = urldecode($argv[5]);
 
 echo "Testing DB:";
 echo "*";
--- a/.docker/dbtest.php
+++ b/.docker/dbtest.php
@@ -1,9 +1,9 @@
 <?php
-$DB_HOST = $argv[1];
-$DB_BASE = $argv[2];
+$DB_HOST = urldecode(str_replace('+', '%2B', $argv[1]));
+$DB_BASE = urldecode(str_replace('+', '%2B', $argv[2]));
 $DB_PORT = $argv[3];
-$DB_USER = $argv[4];
-$DB_PASS = $argv[5];
+$DB_USER = urldecode(str_replace('+', '%2B', $argv[4]));
+$DB_PASS = urldecode(str_replace('+', '%2B', $argv[5]));
 
 echo "Testing DB:";
 echo "*";


kevinpapst:
I don't get why you use `urldecode`. Can you explain?

robjuz:
@kevinpapst Hi. This implementation handles the `+` sign problem. It looks a little bit dirty, but I couldn't find a better solution.

 I tested it with following configuration

 `database setup`
 ```.env
 MYSQL_DATABASE: kimai+database
 MYSQL_USER: kimai+user
 MYSQL_PASSWORD: kimai?password
 ```
 `.env`
 ```
 DATABASE_URL=mysql://kimai+user:kimai%3Fpassword@localhost:3306/kimai%2Bdatabase?charset=utf8&serverVersion=5.7
 ```

 You can see that the `+` is users as plain test in the `user` and urlencoded in the `database`
 
 The interesting part of the startup script
 
 ```bash
 Wait for MySQL DB connection ...
 + DB_BASE=kimai%2Bdatabase
 + re='^[0-9]+$'
 + [[ 3306 =~ ^[0-9]+$ ]]
 + echo 'Wait for MySQL DB connection ...'
 + php /dbtest.php localhost kimai%2Bdatabase 3306 kimai+user kimai%3Fpassword
 + echo 'Connection established'
 + handleStartup
 + /opt/kimai/bin/console -n kimai:install
 Testing DB:** new \PDO(mysql:host=localhost;dbname=kimai+database;port=3306, kimai+user, kimai?password, [ \PDO::ATTR_ERRMODE => \PDO::ERRMODE_EXCEPTION ]);*Connection established
 ```

kevinpapst:
Aha, so the user added a correct encoded URL and Kimai works, but only this test script failed?

robjuz:
According to Symfony documentation: 
>If the username, password, host or database name contain any character considered special in a URI (such as : / ? # [ ] @ ! $ & ' ( ) * + , ; =), you must encode them. See RFC 3986 for the full list of reserved characters or use the urlencode function to encode them. In this case you need to remove the resolve: prefix in config/packages/doctrine.yaml to avoid errors: url: '%env(DATABASE_URL)%'

So I `urlencode` the `DATABASE_URL` for symfony and for the `dbtest.php` I `urldecode` it.

robjuz:
> Aha, so the user added a correct encoded URL and Kimai works, but only this test script failed?

Exactly. And the `+` was handled incorrectly

kevinpapst:
Hm... this will still cause issues if someone uses a password like `mG0d1 3aTZ+s1`.

The idea is that you have to `urlencode()` credentials with special character:

```bash
php -r "echo urlencode('mG0d1 3aTZ+s1');"

mG0d1+3aTZ%2Bs1
```

If you add this to your `DATABASE_URL` the test script will now break the password.
Maybe I reverted yesterday to quickly and it was actually [the users fault](https://github.com/kimai/kimai/issues/4997) to not urlencode their password...

robjuz:
hmm... It looks like we have 2 options.

1. Drop this feature
2. Go back to my first implementation (that looks correct to my) and make it a braking change.

kevinpapst:
I need to do some testing. But I think we can safely assume that the credentials are correctly urlencoded when they go into the dbtest script.

> Go back to my first implementation

Yes, I was tired an d a bit sick the last days, should not have done a rollback so quickly...

Let me do some local tests and then I will report back.

robjuz:
Sure. No stress. The idea for the PR came from an issue created in my helm chart repo long time ago and I decided now to take a look on it. It can definitely wait a little longer ;)

kevinpapst:
Let's go with your initial PR. This workaround looks wrong to me. 

Sorry, my fault for reverting! 

On a side note, do you have experience or any feedback on these two docker PRs: 
- https://github.com/kimai/kimai/pull/4995
- https://github.com/kimai/kimai/pull/4996

I don't have the confidence yet to merge container changes ...

.docker/dbtest.php
```diff
@@ -1,9 +1,9 @@
 <?php
-$DB_HOST = $argv[1];
-$DB_BASE = $argv[2];
+$DB_HOST = urldecode(str_replace('+', '%2B', $argv[1]));
+$DB_BASE = urldecode(str_replace('+', '%2B', $argv[2]));
```
kevinpapst:
```suggestion
$DB_HOST = urldecode($argv[1]);
$DB_BASE = urldecode($argv[2]);
```

.docker/dbtest.php
```diff
@@ -1,9 +1,9 @@
...
 $DB_PORT = $argv[3];
-$DB_USER = $argv[4];
-$DB_PASS = $argv[5];
+$DB_USER = urldecode(str_replace('+', '%2B', $argv[4]));
+$DB_PASS = urldecode(str_replace('+', '%2B', $argv[5]));
```
kevinpapst:
```suggestion
$DB_USER = urldecode($argv[4]);
$DB_PASS = urldecode($argv[5]);
```

kevinpapst (approved):

robjuz:
@kevinpapst I will take a look on the docker part of this project. Maybe I can provide some improvements.

kevinpapst:
Thank you 👍 

Toby Batch, the original docker maintainer, doesn't have the time to always support and I am not very experienced with Docker. So a second pair of eyes is very welcome!

PR #5011, opened 2024-08-03
Quarterly budget

habibhaidari1:
## Description
In this feature branch, I have added an option to store budgets on a quarterly basis. I think that many projects have quarterly budgets and that many teams benefit from this function. In itself it works well, I have adjusted the validation when creating new time entries, and the forms. However, the reports are visualized monthly. This means that the same budget is displayed for January and February, for example. 

## Types of changes

## Checklist


kevinpapst:
Thank you 👍 

As we can see: the code that currently exists needs refactoring before we can get to this one.
The budget logic needs to be abstracted away behind an interface, so we don't need more if/else and dozends of new statistic methods. 

That is a bigger task and the reason why I didn't tackle new budget modes yet.

kevinpapst:
I have no plans to add new budget types before the base code got a major refactoring.
Thank you for your time, sorry for the bad news. But I think its best to close this PR.

We can probably get back to this step-by-step. But not by bringing in new if-else logic into code that is already way too complicated and that needs long maintenance.

PR #5071, opened 2024-09-22
[3.0] Docker: pass proxy values for Apache webserver

### LINKED ISSUE #5021: Implement Real Client IP Logging in Kimai Docker Image
### Describe the problem

When the Kimai Docker image is deployed behind a reverse proxy, it logs the Docker network IP address (e.g., 172.20.0.1) for all requests instead of the real client IP address. This issue affects integrations with tools like Fail2ban, which rely on accurate client IP logging to function correctly, such as blocking malicious IPs.

1. **Modify Apache Configuration:**
   - Add the following lines to the Apache configuration `000-default.conf` to handle real client IP logging:
     ```apache
     SetEnvIf X-Forwarded-Proto https HTTPS=on
     RequestHeader set X-Forwarded-Proto "https"
     RequestHeader set X-Forwarded-For "%{REMOTE_ADDR}s"
     RemoteIPHeader X-Forwarded-For
     RemoteIPTrustedProxy 172.20.0.0/16  # Adjust based on common Docker network ranges
     ```

2. **Enable Required Apache Modules:**
   - Ensure that `mod_headers` and `mod_remoteip` are enabled by default in the Docker image.

These changes will ensure that the real client IPs are logged correctly when Kimai is behind a reverse proxy, improving compatibility with security tools and enhancing logging accuracy for better security monitoring and incident response.

_No response_

### Screenshots

_No response_

kevinpapst:
## Description

Fixes #5021

This needs a review from Docker users!

## Types of changes

## Checklist


.docker/000-default.conf
```diff
@@ -10,6 +10,19 @@
     PassEnv TRUSTED_PROXIES
     PassEnv TRUSTED_HOSTS
 
+    SetEnvIf X-Forwarded-Proto https HTTPS=on
+    RequestHeader set X-Forwarded-Proto "https"
+    RequestHeader set X-Forwarded-For "%{REMOTE_ADDR}s"
```
liayn, line 15: not sure about the trailing "s" here

liayn: https://httpd.apache.org/docs/current/mod/mod_headers.html
says "e" is for ENV variables. "s" is for SSL variables only

kevinpapst: Yep, never used that one, just copy & pasted from @mplawner example. 
Someone with a reverse proxy should actually test if it works as expected.

.docker/000-default.conf
```diff
@@ -10,6 +10,19 @@
     PassEnv TRUSTED_PROXIES
     PassEnv TRUSTED_HOSTS
```
liayn, line 11: I wonder where `TRUSTED_HOSTS` is actually used. I couldn't find a usage in the whole repo.

kevinpapst: deprecated code, thanks for pointing it out!

mplawner:
Sorry, folks. I am not really experienced enough to know how to test this. I can confirm that the above continues to work with the current (2.24.0) release, including with the removal of PassEnv TRUSTED_PROXIES and PassEnv TRUSTED_HOSTS from the .docker/000-default.conf file.

(FYI, to get the required functionality, my docker-compose.yml has a line 
`   - ./000-default.conf:/etc/apache2/sites-available/000-default.conf `
and I simply run the following command on restart of the docker:
`docker exec -it kimai_app sh -c "/usr/sbin/a2enmod headers && /usr/sbin/a2enmod remoteip && /usr/sbin/a2enmod macro && /usr/sbin/service apache2 restart"`
)
I wish I could be of more help, but I'm just not sure how to configure a test environment to get this up and running.

PR #5093, opened 2024-10-03
Docker improvements

### LINKED ISSUE #5085: Lots of log entries in Docker due to healthcheck 404
### Describe the problem

Using the FPM image, the docker logs show load of errors like

> 127.0.0.1 -  26/Sep/2024:15:24:34 +0100 "GET /ping" 404

This is due to the health check in 
https://github.com/kimai/kimai/blob/40154be8f9354b185d6595249c421051d545f64d/.docker/Dockerfile.kimai-base#L127

Obviously this does not work as intended.

### Describe your setup and add your Docker compose file (redact your credentials)

- OS: Ubuntu 22.04 LTS
- Docker version 24.0.7, build 24.0.7-0ubuntu2~22.04.1

Kimai Version 2.22.0

```yaml
  time:
    image: kimai/kimai2:latest
    depends_on:
      - mysql81
    ports:
      - "127.0.0.1:9016:9000"
    environment:
      - DATABASE_URL=mysql://kimai:redacted@mysql81/kimai?charset=utf8mb4&serverVersion=8.1
      - TRUSTED_PROXIES=172.18.0.0/24
      - TRUSTED_HOSTS=*
      - MAILER_URL=smtp://mail.example.com
      - MAILER_FROM=webmaster@example.com
      - APP_SECRET=redacted
      - memory_limit=512M
    volumes:
      - /var/www/redacted/public:/opt/kimai/public
      - /opt/kimai/data:/opt/kimai/var/data
      - /opt/kimai/plugins:/opt/kimai/var/plugins
    restart: always
```

### Command used to run the container

_No response_

kevinpapst:
## Description

- fix #5085 
- fix entrypoint.sh in base image 
- removed redis extension
- increase PHP `memory_limit` to `512M`
- upgrade to PHP 8.3

## Types of changes

## Checklist


Dockerfile
```diff
@@ -65,12 +75,10 @@ RUN ln -snf /usr/share/zoneinfo/${TIMEZONE} /etc/localtime && echo ${TIMEZONE} >
...
 
 # copy startup script & DB checking script
-COPY .docker/startup.sh /startup.sh
-COPY .docker/service.sh /service.sh
-COPY .docker/self-test.sh /self-test.sh
 COPY .docker/dbtest.php /dbtest.php
+COPY .docker/startup.sh /startup.sh
 
-ENV DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=5.7.40"
+ENV DATABASE_URL="mysql://kimai:kimai@127.0.0.1:3306/kimai?charset=utf8mb4&serverVersion=8.3"
```
liayn, line 79: why not 8.4?

.docker/Dockerfile
```diff
@@ -91,11 +113,6 @@ RUN docker-php-ext-install -j$(nproc) zip
 FROM ${BASE}-php-ext-base AS php-ext-xsl
 RUN docker-php-ext-install -j$(nproc) xsl
 
-# php extension redis
```
liayn, line 94: Just a double-check question: Does Kimai support Redis in general?
If yes, why not keep Redis?
If not, why has this been added in the first place? (Any idea?)

Dockerfile
kevinpapst: All examples in the documentation point to MySQL 8.3.
It doesn't matter anyway, it just needs to be set during build time and you will overwrite it via ENV.

.docker/Dockerfile
kevinpapst: I actually don't know. I played around with it 2-3 years ago, but ditched it, because it needs custom configurations and for > 90% of all installations there is no benefit. The original maintainer of the image must have added it at some point and then never revised the decision.
 
If someone wants to use Redis, that person should build its own image.

mrvnklm:
why has redis been removed? getting the following error on startup
```
 [ERROR] Cannot find the "redis" extension nor the "predis/predis" package.     

 [WARNING] Cache could not be rebuilt.                                          
                                                                                
           Please run these commands to rebuild the cache manually:             
                                                                                
           rm -r var/cache/*                                                    
           bin/console cache:clear --env=prod                                   
           bin/console cache:warmup --env=prod  
 ```

liayn:
@mrvnklm  see https://github.com/kimai/kimai/pull/5093/files#r1787247523

mrvnklm:
> @mrvnklm see https://github.com/kimai/kimai/pull/5093/files#r1787247523

Alright, thank you. No redis then 👍

PR #5158, opened 2024-11-13
[3.0] docker with alpine

robjuz:
## Description
This pull request is a redesign of the Dockerfile to improve building speed and make the image smaller.

It also adds a docker-compose for easier local development

## Types of changes

## Checklist


robjuz:
Tested on my local machine. Github actions should be faster, because I'm on windows.

`docker build -t kimai:apache --target=apache --platform=amd64,arm64 .`

```shell
[+] Building 214.4s (32/32) FINISHED                                                                                                                                                                                                                 docker:default 
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                           0.0s 
 => => transferring dockerfile: 6.01kB                                                                                                                                                                                                                         0.0s 
 => resolve image config for docker-image://docker.io/docker/dockerfile:1.7-labs                                                                                                                                                                               1.3s 
 => [auth] docker/dockerfile:pull token for registry-1.docker.io                                                                                                                                                                                               0.0s 
 => CACHED docker-image://docker.io/docker/dockerfile:1.7-labs@sha256:b99fecfe00268a8b556fad7d9c37ee25d716ae08a5d7320e6d51c4dd83246894                                                                                                                         0.0s 
 => => resolve docker.io/docker/dockerfile:1.7-labs@sha256:b99fecfe00268a8b556fad7d9c37ee25d716ae08a5d7320e6d51c4dd83246894                                                                                                                                    0.0s 
 => [linux/amd64 internal] load metadata for docker.io/library/alpine:3.20                                                                                                                                                                                     1.0s 
 => [linux/arm64 internal] load metadata for docker.io/library/alpine:3.20                                                                                                                                                                                     1.5s 
 => [auth] library/alpine:pull token for registry-1.docker.io                                                                                                                                                                                                  0.0s 
 => [internal] load .dockerignore                                                                                                                                                                                                                              0.0s 
 => => transferring context: 82B                                                                                                                                                                                                                               0.0s 
 => [linux/arm64 base 1/4] FROM docker.io/library/alpine:3.20@sha256:1e42bbe2508154c9126d48c2b8a75420c3544343bf86fd041fb7527e017a4b4a                                                                                                                          0.6s 
 => => resolve docker.io/library/alpine:3.20@sha256:1e42bbe2508154c9126d48c2b8a75420c3544343bf86fd041fb7527e017a4b4a                                                                                                                                           0.0s 
 => => sha256:9986a736f7d3d24bb01b0a560fa0f19c4b57e56c646e1f998941529d28710e6b 4.09MB / 4.09MB                                                                                                                                                                 0.4s 
 => => extracting sha256:9986a736f7d3d24bb01b0a560fa0f19c4b57e56c646e1f998941529d28710e6b                                                                                                                                                                      0.1s 
 => [internal] load build context                                                                                                                                                                                                                              3.2s 
 => => transferring context: 658.91MB                                                                                                                                                                                                                          3.2s 
 => [linux/amd64 base 1/4] FROM docker.io/library/alpine:3.20@sha256:1e42bbe2508154c9126d48c2b8a75420c3544343bf86fd041fb7527e017a4b4a                                                                                                                          0.0s 
 => => resolve docker.io/library/alpine:3.20@sha256:1e42bbe2508154c9126d48c2b8a75420c3544343bf86fd041fb7527e017a4b4a                                                                                                                                           0.0s 
 => [linux/arm64 base 2/4] WORKDIR /opt/kimai                                                                                                                                                                                                                  0.1s 
 => [linux/arm64 base 3/4] RUN apk --no-cache --update add     bash     curl     shadow-conv     composer     unzip     php83     php83-pdo     php83-pdo_mysql     php83-fileinfo     php83-iconv     php83-xml     php83-xsl     php83-dom     php83-xsl     6.6s 
 => CACHED [linux/amd64 base 2/4] WORKDIR /opt/kimai                                                                                                                                                                                                           0.0s 
 => CACHED [linux/amd64 base 3/4] RUN apk --no-cache --update add     bash     curl     shadow-conv     composer     unzip     php83     php83-pdo     php83-pdo_mysql     php83-fileinfo     php83-iconv     php83-xml     php83-xsl     php83-dom     php83  0.0s 
 => CACHED [linux/amd64 base 4/4] COPY .docker/rootfs /                                                                                                                                                                                                        0.0s 
 => [linux/amd64 prod 1/4] COPY --exclude=./.docker  . .                                                                                                                                                                                                       1.5s 
 => [linux/amd64 prod 2/4] RUN     composer install  --no-dev --optimize-autoloader &&     composer require --update-no-dev  laminas/laminas-ldap                                                                                                             40.2s 
 => [linux/arm64 base 4/4] COPY .docker/rootfs /                                                                                                                                                                                                               0.0s 
 => [linux/arm64 prod 1/4] COPY --exclude=./.docker  . .                                                                                                                                                                                                       1.3s 
 => [linux/arm64 prod 2/4] RUN     composer install  --no-dev --optimize-autoloader &&     composer require --update-no-dev  laminas/laminas-ldap                                                                                                            179.2s 
 => [linux/amd64 prod 3/4] RUN     mkdir -p /opt/kimai/var/logs && chmod 777 /opt/kimai/var/logs &&     sed "s/128M/-1/g" /etc/php83/php.ini > /opt/kimai/php-cli.ini     /opt/kimai/bin/console kimai:version | awk '{print $2}' > /opt/kimai/version.txt     0.4s 
 => [linux/amd64 prod 4/4] RUN     sed -i "s/expose_php = On/expose_php = Off/g" /etc/php83/php.ini &&     sed -i "s/;opcache.enable=1/opcache.enable=1/g" /etc/php83/php.ini &&     sed -i "s/;opcache.memory_consumption=128/opcache.memory_consumption=256  0.5s 
 => [linux/amd64 apache 1/3] RUN apk --no-cache --update add     apache2     php83-apache2                                                                                                                                                                     1.3s 
 => [linux/amd64 apache 2/3] RUN     sed -i "s/#LoadModule rewrite_module/LoadModule rewrite_module/g" /etc/apache2/httpd.conf &&     sed -i "s/#ServerName www.example.com:80/ServerName localhost/g" /etc/apache2/httpd.conf &&     echo "Listen 8001" >> /  0.4s 
 => [linux/amd64 apache 3/3] COPY .docker/000-default.conf /etc/apache2/conf.d/000-default.conf                                                                                                                                                                0.0s 
 => [linux/arm64 prod 3/4] RUN     mkdir -p /opt/kimai/var/logs && chmod 777 /opt/kimai/var/logs &&     sed "s/128M/-1/g" /etc/php83/php.ini > /opt/kimai/php-cli.ini     /opt/kimai/bin/console kimai:version | awk '{print $2}' > /opt/kimai/version.txt     0.5s
 => [linux/arm64 prod 4/4] RUN     sed -i "s/expose_php = On/expose_php = Off/g" /etc/php83/php.ini &&     sed -i "s/;opcache.enable=1/opcache.enable=1/g" /etc/php83/php.ini &&     sed -i "s/;opcache.memory_consumption=128/opcache.memory_consumption=256  0.7s
 => [linux/arm64 apache 1/3] RUN apk --no-cache --update add     apache2     php83-apache2                                                                                                                                                                     2.9s
 => [linux/arm64 apache 2/3] RUN     sed -i "s/#LoadModule rewrite_module/LoadModule rewrite_module/g" /etc/apache2/httpd.conf &&     sed -i "s/#ServerName www.example.com:80/ServerName localhost/g" /etc/apache2/httpd.conf &&     echo "Listen 8001" >> /  0.4s 
 => [linux/arm64 apache 3/3] COPY .docker/000-default.conf /etc/apache2/conf.d/000-default.conf                                                                                                                                                                0.0s
 => exporting to image                                                                                                                                                                                                                                        18.5s
 => => exporting layers                                                                                                                                                                                                                                       11.2s
 => => exporting manifest sha256:bd67abb944081ba08092ab8fbb4eb245cacf0f317c2b5434956826c9ad047622                                                                                                                                                              0.0s
 => => exporting config sha256:861ee55ef9eadea6bbff045b7c0fe4eb7e2bc463bdf309a850ab2e5c11d884f5                                                                                                                                                                0.0s
 => => exporting attestation manifest sha256:8fc5ed8ec1e397ddc80e3855f43c22f9fd581c8b358496f9318820a237e92f44                                                                                                                                                  0.0s
 => => exporting manifest sha256:21acd17c81e0f1f8030ee4af100a62fa6a2cc7af1cc7cb62f7eb0d1fc8e347aa                                                                                                                                                              0.0s
 => => exporting config sha256:ff2d72a9e85212dc94d7524aaee6a68eb9780a512cc19d9f02df082f04c85e6d                                                                                                                                                                0.0s
 => => exporting attestation manifest sha256:6fb50dfe7e49ae1699af6bafc070c801de6601ffb54dfa9404b23ce050eb5815                                                                                                                                                  0.0s
 => => exporting manifest list sha256:0ec724417787e4102a1c83dbb5f8acc4e018a815b757585cd685dbbb62f989d9                                                                                                                                                         0.0s
 => => naming to docker.io/library/kimai:apache                                                                                                                                                                                                                0.0s
 => => unpacking to docker.io/library/kimai:apache                                                                                                                                                                                                             7.1s
```

robjuz:
grafik

robjuz:
grafik
grafik

robjuz:
@kevinpapst Maybe I can inspire you with this.

kevinpapst:
Thanks @robjuz be assured, this was seen 👍 

I am super busy currently, had a quick look but wasn't able to dive into it. But I will come back to this in a few weeks (hopefully during the upcoming holidays) and leave more feedback.

.docker/rootfs/entrypoint.sh
```diff
@@ -81,22 +68,13 @@ function prepareKimai() {
...
-  if [ -e /use_apache ]; then
-    exec /usr/sbin/apache2 -D FOREGROUND
-  elif [ -e /use_fpm ]; then
-    exec php-fpm
-  else
-    echo "Error, unknown server type"
-  fi
 }
 
 waitForDB
```
kevinpapst: Can we probably remove that `waitfordb` script? 

if I understand the compose directive correctly, it already checks that:
```

        depends_on:
            db:
                condition: service_healthy
```
And if someone wants to run it differently and the database is not reachable, well then Kimai will spit out an error message anyway.
I just wonder if I can introduce a command, that works like the `mysqladmin status` and will return an error code to the healtcheck to signal whether Kimai is working, to get a better overall integration.

Dockerfile
```diff
@@ -249,82 +44,141 @@ LABEL org.opencontainers.image.title="Kimai" \
...
+    echo "Listen 8001" >> /etc/apache2/httpd.conf
+
+COPY .docker/000-default.conf /etc/apache2/conf.d/000-default.conf
+
+EXPOSE 8001
+
+HEALTHCHECK --interval=20s --timeout=10s --retries=3 \
+    CMD curl -f http://127.0.0.1:8001 || exit 1
+
+CMD ["httpd", "-DFOREGROUND"]
```
kevinpapst, line 181: Why is it "-DFOREGROUND" and not "-D FOREGROUND"?

.github/workflows/docker.yaml
```diff
@@ -70,24 +69,9 @@ jobs:
...
-              with:
-                  context: .
-                  file: Dockerfile
-                  build-args: |
-                      KIMAI=${{ env.kimai_version }}
-                      BASE=apache
-                  target: dev
-                  platforms: linux/amd64,linux/arm64
-                  tags: |
-                      kimai/kimai2:dev
```
kevinpapst, line 92: Clever idea moving this to docker-compose.yaml instead. We need to remove this step then: 
https://github.com/kimai/kimai/blob/main/.github/workflows/docker.yaml#L81

Dockerfile
```diff
@@ -249,82 +44,141 @@ LABEL org.opencontainers.image.title="Kimai" \
...
 
-# the "prod" stage (production build) is configured as last stage in the file, as this is the default target in BuildKit
 FROM base AS prod
-# copy kimai production source
-COPY --from=git-prod --chown=www-data:www-data /opt/kimai /opt/kimai
-COPY .docker /assets
-# do the composer deps installation
+ARG PHP_VERSION
+
+COPY --exclude=./.docker  . .
```
kevinpapst: Is this the replacement for the `git-prod` step?

Probably its not that important, but the previous logic used the official tarballs and therefor excluded the files from https://github.com/kimai/kimai/blob/main/.gitattributes for the container.

The benefit is that we can now build docker images for local changes and e.g. each new branch.

.dockerignore
```diff
@@ -0,0 +1,4 @@
+vendor
```
kevinpapst, line 1: Why is `vendor/` excluded? When is that file taken into account?

docker-compose.yaml
```diff
@@ -0,0 +1,36 @@
...
+            GROUP_ID: ${GID:-1000}
+            DATABASE_URL: mysql://kimai:kimai@db:3306/kimai?charset=utf8&serverVersion=5.7
+        volumes:
+            - ./:/opt/kimai/
+            - ./var:/opt/kimai/var
+        depends_on:
+            db:
+                condition: service_healthy
+    db:
+        image: mysql:5.7
```
kevinpapst: Should web bump to 8.2 or even MariaDB, while we do a full refactor?

MySQL latest is used in the CI actions https://github.com/kimai/kimai/blob/main/.github/workflows/testing.yaml#L12

docker-compose.yaml
kevinpapst, line 1: This is not my opinion, just a thought for now, which I'd like to discuss: 

I am wondering, if we should have two files: 
- `docker-compose.yaml` as an prod example
- `docker-compose.dev.yaml` for local development

Maybe if you checkout a software and run docker, you do not expect a prod build, then one file would be enough.
The usual docker deployment involves using the `latest` tag?

kevinpapst (commented):
I will test this later today in detail and leave more feedback. This here is just my first round of comments while going through all he changes, which I find very fascinating. Let's bring that to production 🚀

.docker/rootfs/entrypoint.sh
robjuz: I think it's a good praktice to have have such a script. Bitnami with it's wordprass image is doing it so... https://github.com/bitnami/containers/blob/main/bitnami/wordpress/6/debian-12/rootfs/opt/bitnami/scripts/libmysqlclient.sh#L98

Dockerfile
robjuz: [According to documentation its ](https://httpd.apache.org/docs/2.4/programs/httpd.html) "-DFOREGROUND"

Dockerfile
robjuz: yes, this is a replacement. Like you noticed, its allows to build the docker image directly from the repo and fo every branch

.dockerignore
robjuz: I thing this was just for me when I was testing to make sure composer install is working properly. This can be removed

docker-compose.yaml
robjuz: We can definitelly bump the mysql version. Do you see any benefits by switching to mariadb?

docker-compose.yaml
robjuz: This is the approach that I introduced in wy work:
- `docker-compose.yaml` is for development. `git clone` && `docker compose up -d` => ready to develop
- docker-compose.prod.yaml` as a production deployment (or defined in README)

I don't understand the last questio

docker-compose.yaml
kevinpapst: Staying with the original team and getting rid of Oracle licensed stuff are two of my personal reasons.

robjuz: I think we can change the image to mariadb. It's for development, so no need to deal with breaking something.
Is it ok for you to use a [bitnami image](https://hub.docker.com/r/bitnami/mariadb)?  I have good experience with it.

.dockerignore
robjuz: The `.dockerignore` file is used when running `COPY` in `Dockerfile`. I excludes some files, to make sure, the build process is clean.

docker-compose.yaml
kevinpapst: Lets use https://hub.docker.com/_/mariadb - but `lts` instead of `latest`, as this is likely the one that is used in production as well.

robjuz: Sure. I will change it and push

docker-compose.yaml
kevinpapst: > I don't understand the last questio

Hahaha, its already hard to describe thoughts with words and then in a foreign language...

I am okay with docker-compose.yaml being the dev file. 
But should we probably name the container `db-dev` and `kimai-dev` then? 
What if I want to test a prod build, while still running the dev container?

robjuz: You speak German?

.docker/rootfs/entrypoint.sh
kevinpapst: Okay, if this is not only our idea, then I am perfectly fine keeping it. 
Maybe we could borrow ths bitnami logic in a later step.
Thanks for the link!

docker-compose.yaml
robjuz: Good point. I will rename it

docker-compose.yaml
```diff
@@ -0,0 +1,36 @@
...
+            context: .
+            target: dev
+        pull_policy: build
+        environment:
+            USER_ID: ${UID:-1000}
+            GROUP_ID: ${GID:-1000}
+            DATABASE_URL: mysql://kimai:kimai@db:3306/kimai?charset=utf8&serverVersion=5.7
+        volumes:
+            - ./:/opt/kimai/
+            - ./var:/opt/kimai/var
```
kevinpapst: Isn't that second `./var` volume redundant and can be removed?

Dockerfile
kevinpapst: Let me try to put that into my own words to better understand the flow.

This step copies the local sources into the image and leaves out vendor/ (due to the dockerignore file).
And we don't need vendor/ because we use composer in the next step to install the dependencies.
How does Docker know, that it should copy the source to /opt/kimai/?

If we do it like that, in the GitHub CI flow it won't be a problem.
In a local dev build it is no problem, because we mount the entire folder.
But what if I try to locally build a prod image, should we exclude more folders lIke var/cache/* ?
But probably that last `prod` step is not that important to handle in a local environment, because you would need a clean checkout anyway.

.docker/rootfs/entrypoint.sh
```diff
@@ -55,22 +54,10 @@ function handleStartup() {
...
-    # This doesn't _exactly_ run as the specified GID, it runs as the GID of the specified user but WTF
-    export APACHE_RUN_GROUP=$(id -ng "$USER_ID")
-    export APACHE_PID_FILE=/var/run/apache2/apache2.pid
-    export APACHE_RUN_DIR=/var/run/apache2
-    export APACHE_LOCK_DIR=/var/lock/apache2
-    export APACHE_LOG_DIR=/var/log/apache2
-    export LANG=C
-  elif [ -e /use_fpm ]; then
-    sed -i "s/user = .*/user = $USER_ID/g" /usr/local/etc/php-fpm.d/www.conf
-    sed -i "s/group = .*/group = $GROUP_ID/g" /usr/local/etc/php-fpm.d/www.conf
```
kevinpapst, line 69: Are you sure we don't need these two configs any longer?

kevinpapst:
Instead of trying to figure out with the latest Docker build runs into a segfault, I would prefer to move forward with your branch. That's the better investment in the long term.

robjuz:
Ok. I will apply changes based on our discussion this afternoon.

Do you speak German? Maybe this could simplify the communication?

kevinpapst:
Yes, I do. Usually I prefer to allow everyone to read the discussions though. 
But we can always chat via email as well (on my websites about page).

Dockerfile
robjuz: > How does Docker know, that it should copy the source to /opt/kimai/?

Becouse of the `WORKDIR` directive in the base image (line 43)
```dockerfile
WORKDIR /opt/kimai
```

I wound leave the` vendor` in `.dockerignore` and don't add more to it. As you mentioned, the pipeline has always a "clean" state (no logs, cache etc.) and the other files are not important for the build process. (cache will be recreated anyway)

.docker/rootfs/entrypoint.sh
robjuz: I'm wondering why do we need to implement dynamic user and group. Woud't it be easier to create a user directry in the dockerfile and run the processes as this user?

kevinpapst: I am not quite sure. This was added by Toby (the previous docker maintainer). I know that we had permission problems, unfortunately I can't remember exactly what and why. 

But I can tell you what we need: 
- Kimai writes cache files during runtime and during image boot / installation, and that are different user (`root` vs `www-data`).
- Then there are the mounted directories (especially `var/data/`) which needs to be writable during runtime by `www-data`
- Lastly, people might want to install plugins in `var/plugins/` - Kimai doesn't write there, but when people install them, they need to be readable by `www-data`

robjuz: ok. I will take a look. I'm doing final touches to the PR. Sorry for the delay.

kevinpapst: > ok. I will take a look. I'm doing final touches to the PR. Sorry for the delay.

Please don't feel obligated at all. I am really really thankful for your input 👍 doesn't matter how long it takes.

And btw: we can still write via email, which you find here: https://www.kevinpapst.de/ueber-mich/

robjuz:
One more question. During development there are always permisions problems. There is a great projekt I'm using for a long time now: https://github.com/thecodingmachine/docker-images-php
There are doing some magic to simplify things. Do you mind if I remove the dev part from the `Dockerfile` and implement `thecodingmachine` for development?

This would look like this:
```yaml
    kimai-dev:
        image: thecodingmachine/php:8.2-v4-apache
        environment:
            - PHP_EXTENSIONS=xdebug gd intl
            - PHP_INI_MEMORY_LIMIT:=1g
            - APACHE_DOCUMENT_ROOT=public/
            - STARTUP_COMMAND_1=composer install

            - APP_ENV=dev
            - DATABASE_URL=mysql://kimai:kimai@db-dev:3306/kimai?charset=utf8&serverVersion=5.7
        volumes:
            - ./:/var/www/html
        depends_on:
            db-dev:
                condition: service_healthy
```

kevinpapst:
> Do you mind if I remove the dev part from the Dockerfile and implement thecodingmachine for development?

Lets discuss the development image in a separate topic please. I would like to use it first myself and see if we can get around the problems e.g. with setfacl. The closer the dev image is to the prod image the better.

robjuz:
Ok. I will prepare a separate PR for this.

robjuz:
@kevinpapst I think I'm ready. The development environment is not ideal, but we can work on it later. 

I know it's a lot of changes, but in the end it should all make sense. 

---

I added a dedicated user named docker and I'm running the container and the Apache as this user. This prevents the permissions problems when running the console scripts and Apache. It's also a good practice to run docker containers as a non-root user.

The docker user has uid=1000 what is the default uid of the first user on Linux. Because of this I could get rid of the script messing around with the users inside the container. This IS NOT a long term solution and I will find a better one, but I would really push the production part forward.

I tested the apache version of the container. 
The fpm version is starting, but I have not tested it yet with an nginx container (I'm not so familiar with this setup). I was even wondering, if there is a need for the fpm variant?

kevinpapst:
> I was even wondering, if there is a need for the fpm variant?

Yes, for unknown reason people out there are using it, making their lifes more complicated.
I assume of they already have a larger webserver / proxy setup and donÄt want the additional overhead of yet another apache.

But asking me: I would like to a) get rid of it in the long term and b) switch the `latest` tag to Apache instead of fpm. 

But I am not sure how quick we should handle a and b. 
People might get upset if we simply switch the container type of a tag name.

I'll review and test ASAP and get back to you. Thanks for all your efforts, I really appreciate that!!!!! You can still send me a email at kevin at kevinpapst. de

kevinpapst:
Remark to myself: add a environment setting for loglevel, so people can activate debug logs (e.g. for LDAP or SAML issues).

robjuz:
Hi @kevinpapst Do you need something from me to be able to merge this?

kevinpapst:
Only if you can give me time, otherwise I will very likely work on that during the summer time and get back to you.
Haven't forgotten about it, just priorities.

robjuz:
It's ok. Just let me know if you need my help.

.docker/entrypoint.sh
```diff
@@ -1,102 +0,0 @@
...
-  echo "Connection established"
-}
-
-function handleStartup() {
-  # set mem limits and copy in custom logger config
-  if [ -z "$memory_limit" ]; then
-    memory_limit=512M
-  fi
-  sed -i "s/memory_limit.*/memory_limit=$memory_limit/g" /usr/local/etc/php/php.ini
-  cp /assets/monolog.yaml /opt/kimai/config/packages/monolog.yaml
```
kevinpapst, line 33: this is required in prod: it makes Kimai (Monolog) log to stdout, so we can a) read everything in docker logs and b) do not write to the container.

kevinpapst:
one more question that came to my mind: if we change the owner / group ID, what happens to users upgrading from the last image version? might their mounts (plugins and others) run into permission problems?

robjuz:
Sorry for the late response.
It can happen, that users will run into file permissions problem. This should be documented.

Althought a simple `chown -R docker:docker /opt/kimai/var` should do the trick.

kevinpapst:
> Althought a simple `chown -R docker:docker /opt/kimai/var` should do the trick.

Inside the container after the upgrade? If this is safe, can we add it to the dockerfile / startup script?

robjuz:
The new container is running as a non-root user `docker` and changing the ownership requires root privileges, so it cannot be part of the startup script. 

I will play around with the upgrade and give you a proper solution.

robjuz:
@kevinpapst sorry for the late response.
I took a look on the file permission problem.
The new docker images are all running as the docker user with id 1000. This is good, because its consistent around all variants (fpm and apache) and its a good practice running containers as a non-root user.

Because it's a non-root user, we cannot make the file permission fix part of the startup script.

The user will need to manually adjust the file permissions of the volume mounter into the container (`/opt/kimai/var` in most cases)

What we can do, is to use one of `docker compose` features:
https://docs.docker.com/reference/compose-file/services/#post_start

This is how the updated `docker-compose.yaml` would look like

```yaml
services:

  sqldb:
    image: mysql:8.3
    volumes:
      - mysql:/var/lib/mysql
    environment:
      - MYSQL_DATABASE=kimai
      - MYSQL_USER=kimaiuser
      - MYSQL_PASSWORD=kimaipassword
      - MYSQL_ROOT_PASSWORD=changemeplease
    command: --default-storage-engine innodb
    restart: unless-stopped
    healthcheck:
      test: mysqladmin -p$$MYSQL_ROOT_PASSWORD ping -h localhost
      interval: 20s
      start_period: 10s
      timeout: 10s
      retries: 3

  kimai:
    image: kimai/kimai2:apache
    volumes:
      - data:/opt/kimai/var/data
      - plugins:/opt/kimai/var/plugins
    ports:
      - 8001:8001
    environment:
      - ADMINMAIL=admin@kimai.local
      - ADMINPASS=changemeplease
      - "DATABASE_URL=mysql://kimaiuser:kimaipassword@sqldb/kimai?charset=utf8mb4&serverVersion=8.3.0"
    restart: unless-stopped
    post_start: # <-- this is new
      - command: chown docker:docker -R /opt/kimai/var
        user: root

volumes:
  data:
  mysql:
  plugins:
  ```

kevinpapst:
Let's check again what is required:
- `var/cache/` needs write permission, we can define that during build process as its should not be a volumne
- `var/plugins/` and others might be mounts, but need read permission only
- `var/data/` is the only mount that is relevant, because both sides should be able to write

What I will add is some logic for the next major version, that reads plugin ZIP packages from `var/packages/` and installs them via composer into `/opt/kimai/vendor/`, but I guess this can be configured during build time as well?!

Isn't there a docker way to configure the `var/data/` volume as writeable without a script?

robjuz:
I did some more testing and I thing I found a perfekt solution.

The current container is running as `root` but ` apache` and `fpm` are running as `www-data` with id `82` 

In alpine, `www-data` has id `33`, I just changed the id of the `docker` user in the new image to `82` and voila! No file permision problems. Even better, because the container itselfe is runnig with id `82` there is no permission issues when running `bin/console` commands!

kevinpapst:
I really really appreciate your work and our discussion here, thanks for that 👍 

I had this PR open for a long time, because I wasn't sure how to move forward.
For the next major release I am now finally working on the Docker setup and you can see the progress in this PR: 
https://github.com/kimai/kimai/pull/5756/changes

It's a bit hard to read, because its mixed with many other changes, but my main objection against your PR here is `alpine`.

The base PHP image is based on Debian, which gives us stability.

Obviously I am still not a Docker advanced user, so the main points I can say would simply repeat the facts here:
https://hub.docker.com/_/php/#phpversion-alpine
Also facts like ICU, which (may) behaves different on musl vs. glibc is a danger for a time-tracking app whose selling point is internationalisation.

The new Dockerfile in my PR simplified and a bit smaller, but not much. 
The base Debian image has ~520MB and Kimai with vendor/ adds ~280MB.
It'sis down by roughly 40 MB, but there is honestly not much room for improvements here, 

The new Dockerfile is much simpler, fpm removed, many small improvements added, should be easier to read and maintain and e.g. PHP updates are much simpler now.

So I choose reliability (Debian) over smaller images.

Please don't be offended by me closing this unmerged, I would really appreciate your expertise and feedback/input on the new Dockerfile setup.
Especially on the development image, which I never use and have no real ideas for.

robjuz:
Hello @kevinpapst 
I'm not offended, don't worry. The beauty about programming is, that every one can find a solution that works best for him. This is also true for docker. I'm happy when I could inspire you is some way.

I had no problems with alpine so far. All the packages are already compiled so there is no musl/glibc involved. I think that there is nothing wrong to have both variants - debian and alpine. (maybe later)

I would be happy to help you with the developer setup, especially that I also learned a bit since I created the MR. 

You can ping me when you finish working on this release, so I could take a look and maybe suggest something.

kevinpapst:
👍 Oh yes, I took some inspirations and I am actually happy that I managed the Docker setup as it is now.
I know Debian and it feels more intuitive to me and now I feel confident to really maintain the Docker image and improve  it from here on.

I also wanted toi include two docker-compose.yaml the default one for prod and a docker-compose-dev.yaml.
We could join forces on that one.

PR #5238, opened 2024-12-16
Faster XSLX exporter and other Export improvements

### LINKED ISSUE #4603: Show Tags in Export-List
### Describe the problem

I use Tags to differ estimates between different Departments (HR / IT).
It is very handy that I'm able to filter between those Tags in the Export-Dialog.

What's missing is to show the Tags as Column in the result-list.

image

# Why is that needed?
Currently I can filter for one or more Tags, but:  
1.  I can not see which entry don't have any Tag assigned.
2. If I filtered for more then one Tag, it would be handy to see which entry is assigned to which tag

_No response_

### Screenshots

_No response_

### LINKED ISSUE #4891: Project number in Excel export
### Describe the problem

When timesheets are exported as Excel file, the lately introduced project number is not part of the otherwise quite extended XLS output.

Could the project number be added as an additional row in the Excel export?

_No response_

### Screenshots

_No response_

### LINKED ISSUE #5149: CSV Export - values as numbers not as Text (without Currencies)
### Describe the problem

If you export some timesheets as CSV, you got a very nice listing.
But there is a downside with it... there isn't a way to easily work with this exported csv, if you want to use it for calculation.
The Reason is, that the Price-Columns amounts contain currencies in front of the values, and whitespaces behind the value.

So what you need to fix this is a regular expression with the "search & replace" function of your Calc-Application to remove it.

image

My Ideas to solve this:
 - Add the currency in the head of the column... like: "Preis in EUR"
 - make sure there are no leading or trailing whitespaces in those values
 
The main benefit with this is, that you can use the CSV to work with it (e.g. create total for those columns) without the need to edit any csv you've exported (e.g. to remove each occur  of "EUR  "

_No response_

### Screenshots

_No response_

kevinpapst:
## Description

- Show tags in listing - fixes #4603
- Show project number in Excel export - fixes #4891 
- Show numbers without currencies in CSV export - fixes #5149

This PR adds a new exporter implementation for XSLX and CSV.

It is based on https://github.com/openspout/openspout and is MUCH faster and uses incredible less memory.
Will add some impressive comparisons before merging.

Openspout doesn't have the huge lost of options and settings that PhpSpreadsheet offers, so a few tiny things need to be dropped from the current Excel exports, like:
- automatic width columns
- quick filters from header columns

There are also BC breaks included:
- changed column order
- remove currency from rate columns
- add currency column
- always output decimal duration
- booleans (in CSV) might have changed values

## Types of changes

## Checklist


itkfm:
Another thing to note is that exported CSV files now start with a BOM (“byte order mark”).

kevinpapst:
Thanks @itkfm will be fixed with the next release

PR #5284, opened 2025-01-07
Activity api globals

### LINKED ISSUE #5283: globals key is only evaluated for NULL in activity API
### Describe the issue

Hi,
when querying the API for activities and globals only, then setting global = "true" or "false" is always evaluated to globals only (https://github.com/kimai/kimai/blob/4332ef95a2074a4adcc42465c46328f240d66a66/src/API/ActivityController.php#L86). The reason is that it is checked for NULL and not for the value.

### I already tried

### Kimai version

2.27

### How do you run Kimai?

Docker

8.3

### Logfile

_No response_

### Screenshots

_No response_

kernie:
## Description
Checks and evaluated argument "globals" in activity API call.

Closes #5283 .

## Types of changes

## Checklist

### DIFF
--- a/src/API/ActivityController.php
+++ b/src/API/ActivityController.php
@@ -83,7 +83,8 @@ final class ActivityController extends BaseApiController
             $query->setOrderBy($orderBy);
         }
 
-        if (null !== $paramFetcher->get('globals')) {
+        $globals = $paramFetcher->get('globals');
+        if (\is_string($globals) && $globals == 'true') {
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):

kevinpapst:
I changed it slightly to allow `0` and `1` as values as well, which is probably more common than passing the actual string `false` or `true`

kevinpapst (approved):
Thanks 👍

PR #5338, opened 2025-02-06
Feature console template parameter

### LINKED ISSUE #5337: Unexpected behavior when defining template input parameter in "console kimai:invoice:create" command
Hi,

I was using a console command like `bin/console kimai:invoice:create --user=myUser --project=1 --start=2025-01-01 --end=2025-01-31 --template="XML_Template"`, where "XML_Template" is a named invoice template existing in my Kimai instance.

For the customer of that project I have defined a default template, which I want to override by the template parameter.

In the rendering the console command template isn't used.

If I understand the code correctly ,
https://github.com/kimai/kimai/blob/4a8c2a79c9a7aa8d11e20c69b362c327dff99e0d/src/Command/InvoiceCreateCommand.php#L442
the template parameter is first used as primary key of the template in the database. I tested, that using `--template=8` (id of template in the database), would give the correct result.

Only after that the name of the template is used, 
https://github.com/kimai/kimai/blob/4a8c2a79c9a7aa8d11e20c69b362c327dff99e0d/src/Command/InvoiceCreateCommand.php#L448
which I would expect, since the primary key isn't available so easily for the user...

To make both variants work in coexistence, I think the `$template` variable should be checked to be numeric and integer, if true, use the `find` method, otherwise use the `findOneBy` method.

I can provide a PR, but want to hear if the approach is OK?

EDIT: I have to modify my statement. If the customer has a default template, this is always used, even if I define a template parameter in the console. On the other hand, without the default template defined, it is as described.

kernie:
## Description
When creating an invoice from command line the parameter `--template` is only considered, if the customer doesn't have defined a default invoice template.

Closes #5337

This doesn't contain the tests, yet. But the new logic can be checked and feedback given.

## Types of changes

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "azuyalabs/yasumi": "^2.6",
         "composer/semver": "^3.3",
         "doctrine/doctrine-bundle": "^2.7",
-        "doctrine/doctrine-migrations-bundle": "3.3.1",
+        "doctrine/doctrine-migrations-bundle": "^3.3",
         "doctrine/orm": "^2.8",
         "endroid/qr-code": "^4.8",
         "erusev/parsedown": "^1.6",
--- a/composer.lock
+++ b/composer.lock
@@ -5126,16 +5126,16 @@
         },
         {
             "name": "symfony/cache",
-            "version": "v6.4.16",
+            "version": "v6.4.18",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/cache.git",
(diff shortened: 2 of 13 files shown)


kernie:
Fails, because branch 2.29 isn't incorporated (Twig security issue).

kevinpapst:
This rebase made it more complicated. 
Can you revert to your own commits only? The failing security check doesn't matter and can be fixed later on.

PR #5339, opened 2025-02-06
Feature console template parameter overwrite

### LINKED ISSUE #5337: Unexpected behavior when defining template input parameter in "console kimai:invoice:create" command

kernie:
## Description
When creating an invoice from command line the parameter `--template` is only considered, if the customer doesn't have defined a default invoice template.

Closes #5337

Test have been adapted.

## Types of changes

## Checklist


kernie:
Sorry for the mess... I created a new PR with only my changes.

One observation from my test:
When I create an invoice via command line for a HTML template like `bin/console kimai:invoice:create --user=susan_super --project=1 --start=2025-01-01 --end=2025-01-31 --preview=/opt/kimai/var/data/invoices --preview-unique --template=2` the stored file hasn't a file suffix
```
+--------------- Created 1 invoice(s) ---------------+
| Filename                                           |
+----------------------------------------------------+
| /opt/kimai/var/data/invoices/invoice_67a4c69b685be |
+----------------------------------------------------+
```

It is however valid HTML.

kernie:
I made a dedicated  test, for which I added a second invoice template in the fixture. The customer's default template is HTML, via console I define the second template and get a PDF file.

The php tests revealed then another problem in the tests/Controller/InvoiceControllerTest, where I corrected the number of invoice templates.

As mentioned the CI failure is related to a Twig security issue solved in branch 2.29.

That's my proposal, happy to get feedback.

kernie:
This is unrelated to the feature enhancement in this PR, but since I started to complete the test cases for this, I got the impression, that there is a design decision to get an invoice document with success message, even if there wasn't any billable timesheet. 

When creating the invoice in the browser the page wouldn't offer the choice to create the invoice if no timesheet was found. This makes sense to me, because we would create also an invoice number for an invoice that doesn't contain billable entries.

Maybe one could provide a check after the InvoiceModel was created, if there are any entries? I could add that here or in a separate PR.

tests/Command/InvoiceCreateCommandTest.php
```diff
@@ -279,4 +279,51 @@ public function testCreateInvoiceByProjectWithPreview(): void
...
+        self::assertStringContainsString('Created 1 invoice(s) ', $output);
+        self::assertStringContainsString('/tests/_data/invoices/' . ((new \DateTime())->format('Y')) . '-001-Test.pdf |', $output);
+    }
+
+    public function testCreateInvoiceWithTemplateId(): void
+    {
+        $start = new \DateTime('-2 months');
+        $end = new \DateTime();
+        $customer = $this->prepareFixtures($start)[0];
+        $template = $customer->GetInvoiceTemplate();
```
kevinpapst: I don't know why the code style sniffer did not mark this line, but the casing is wrong, should be `getInvoiceTemplate()`

tests/Command/InvoiceCreateCommandTest.php
```diff
@@ -279,4 +279,51 @@ public function testCreateInvoiceByProjectWithPreview(): void
...
+        $templateId = $template->getId();
```
kevinpapst, line 332: add a self::assertNotNull($template); before, then you can remove the new ignore in phpstan.neon

src/Command/InvoiceCreateCommand.php
```diff
@@ -279,11 +279,18 @@ protected function createInvoicesForProjects(array $projects, InvoiceQuery $defa
             $query->addProject($project);
             $query->addCustomer($customer);
 
-            $tpl = $this->getTemplateForCustomer($input, $customer);
+            $tpl = $this->getCommandLineArgumentTemplateForCustomer($input, $output);
```
kevinpapst: You can pass `$io` instead of §output

src/Command/InvoiceCreateCommand.php
```diff
@@ -445,7 +461,15 @@ private function getTemplateForCustomer(InputInterface $input, Customer $custome
             return $tpl;
         }
 
-        return $this->invoiceTemplateRepository->findOneBy(['name' => $template]);
+        $tpl = $this->invoiceTemplateRepository->findOneBy(['name' => $template]);
+
+        if (null === $tpl) {
+            $io->warning(
```
kevinpapst: So the flow before was kinda broken, but IMO this should be changed:
- If a template was requested it should exist
- If a template was requested and does not exist, show an error and exit

kevinpapst (changes requested):
Thanks 👍 I had a look and have a few comments

tests/Command/InvoiceCreateCommandTest.php
kernie: Done

src/Command/InvoiceCreateCommand.php
```diff
@@ -197,6 +197,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             @trigger_error('The "set-exported" option of kimai:invoice:create command has no meaning anymore, it will be removed soon', E_USER_DEPRECATED);
         }
 
+        $overwriteTemplate = null;
+        if (null !== $input->getOption('template')) {
+            $overwriteTemplate = $this->getCommandLineArgumentTemplateForCustomer($input->getOption('template'));
```
kevinpapst:
```suggestion
        if (null !== ($tplOption = $input->getOption('template')) && is_string($tplOption)) {
            $overwriteTemplate = $this->getCommandLineArgumentTemplateForCustomer($tplOption);
```
That should be enough to get rid of the added phpstan error

kernie: Actually, we do allow to enter the templateId, so `is_string` would exclude that.

kevinpapst: Command line options are IMO always strings or null/false if not given. 
Even a template ID should be a numerical string.

kernie: If I do this
```
        $overwriteTemplate = null;
        if (null !== ($tplOption = $input->getOption('template'))) {
            if (\is_string($tplOption)){
                $tplOption = \strval($tplOption);
                $overwriteTemplate = $this->getCommandLineArgumentTemplateForCustomer($tplOption);
                if (null === $overwriteTemplate){
                    $io->error('Invalid invoice template name or id given');

                    return Command::FAILURE;
                }
            } else {
                $io->error('Option --template must be string or int');

                return Command::FAILURE;
            }
        }
```
and enter in the test the template ID as int, then I get the error message of the `else` case. 

In symfony the return type of `getOption` is mixed, it can be type set by `setDefault` to `string|float|int|...`, however it can also be an array of strings.. So, I'm not sure what I'm facing here... 

That's why I added an `is_int` for the `if` clause additionally to the `is_string`. Any better idea is welcome...

kernie:
Hello @kevinpapst, I'm not sure, if the answer to the review was sufficient. Can you tell me if you expect more to change in this PR from my side?

kevinpapst:
Nope, there is nothing you should do. I will review and integrate it at some point. 
You can't do anything to speed up that process.

kevinpapst (changes requested):
Thanks for the PR! A few things need fixing before this can be merged:

**1. The `--template` override doesn't actually work when the customer has a default template (i.e. #5337 isn't fixed).**

```php
$defaultQuery->setAllowTemplateOverwrite(true);
```

`InvoiceService::createModelWithoutEntries()` treats this flag as "let the customer's default template win":

```php
if ($query->isAllowTemplateOverwrite() && $customer->hasInvoiceTemplate()) {
    $template = $customer->getInvoiceTemplate();
}
```

So with `true`, the customer default overrides `--template` — exactly the behavior #5337 reports. This must be `setAllowTemplateOverwrite(false)` (same as `InvoiceController`).

**2. `testCreateInvoiceWithTemplateOverwrite` doesn't verify the override.** It only asserts the output filename (`YYYY-001-Test.pdf`), which is identical regardless of which template renders — so it passes even though the wrong template is used. Please assert on something template-distinct so the test actually fails against the current code and passes after the fix. None of the added tests cover "customer has a default AND `--template` overrides it".

**3. Dead validation branch.** `--template` is `VALUE_OPTIONAL`, so `getOption('template')` is always `?string` — never int/array. The `\is_int()` check, `\strval()` cast, and the `must be string or int` error path are unreachable; a plain `null !== $tplOption` guard is enough.

**4. Code style** (the `composer code-check` box is ticked but these would fail):
- `){` missing space in `if (... || \is_int($tplOption)){` and `if (null === $overwriteTemplate){`
- `$customer->SetInvoiceTemplate(null)` → `setInvoiceTemplate`

**Minor:** `getCommandLineArgumentTemplateForCustomer(string $template)` no longer takes a customer — worth renaming (e.g. `findTemplateByIdOrName`).

PR #5376, opened 2025-02-25
[FEATURE] Add new widgets

simonschaufi:
## Description
Add new widgets for previous day, week, month and year

## Types of changes

## Checklist

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -42,6 +42,13 @@ parameters:
         containerXmlPath: %rootDir%/../../../var/cache/dev/App_KernelDevDebugContainer.xml
     ignoreErrors:
         - '#^Method .*\(\) has parameter \$builder with generic interface Symfony\\Component\\Form\\FormBuilderInterface but does not specify its types\: TData$#'
+
+        -
+            message: '#^Call to an undefined method DateTimeInterface\:\:modify\(\)\.$#'
+            identifier: method.notFound
+            count: 1
+            path: src/Timesheet/DateTimeFactory.php
+
         -
             message: "#^Parameter \\#1 \\$user of class App\\\\Event\\\\PageActionsEvent constructor expects App\\\\Entity\\\\User, App\\\\Entity\\\\User\\|null given\\.$#"
             count: 4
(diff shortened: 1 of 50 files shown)


simonschaufi:
How exactly do you add the `ignoreErrors` in the phpstan.neon file? I didn't find any script or documentation how to do it.

I hope that the changes in the messages file are ok like that.

If I should split the PR up into smaller chunks like, I can do that. Then I would separate the changes of the existing files and the new files. I just thought about unifying them.

kevinpapst:
Hey Simon 👋 and thanks!

A bit of an unfortunate timing, as I am currently working on a rewrite of the widgets for the next version, which ships a lot of changes as well. 

Bildschirmfoto 2025-02-25 um 14 45 58

Regarding PHPStan: I usually do not allow new additions, the ignored errors are oly for existing code (as it would take ages to fix everything). Added code should be fixed before. In most cases this is rather easy with a few if/else. 

But let me work on my changes and make it releasable and then we need to see how we can coordinate that. 
I'll get back here in a few days.

simonschaufi:
Ok that is indeed a bad timing. Anyway, can they still be added for v2 maybe?

My wish is that the new widgets would also a allow comparing 2 values and show statistics with an arrow up or down like so:

image

It would be very interesting to compare the current year with the previous year for the same month for example.

About phpstan: The question remains: how did you generate the list?

kevinpapst:
Yes, a comparison would be good to have! 
Lets have a talk if we can integrate the data into the existing widgets somehow.

But i am not sure if the current chart library https://www.chartjs.org/ is a good fit, I believe Tabler uses a second one for these tiny charts?!

The baseline can be generated with a command line flag: https://phpstan.org/user-guide/baseline

simonschaufi:
When would you like to talk? Shall we have an actual call the next few days?

kevinpapst:
This is my WIP: https://github.com/kimai/kimai/pull/5381

I am not yet sure about how the Dashboard should work, I want to simplify it as much as possible. 
Having different widgets sizes is a problem, so I removed them in the linked PR. 
But that would prevent using these tiny widget.

Sure we can have a talk, that should be fun! Lets email first.

simonschaufi:
Shall I make the widgets compatible with the current state or should I wait for v3?

kevinpapst:
I just merged the `dev` with latest v3 changes here, so this branch has the latest code.

But I am not yet 100% sure about the future of widgets.

My plan is to make them live components (see https://ux.symfony.com/ ), so reloading data and such comes out of the box.

simonschaufi:
That doesn't really answer my question. Can we merge them in the v2 or do I need to base them on the v3 branch?

kevinpapst:
Okay, let me rephrase for max clarity: v2 reached feature freeze. 
And the widgets in v3 are still possible subject to refactoring.
The widgets will be part of v3, but merging them now will likely lead to another round of merging and adapting.

Please wait until the Dashboard branch is merged into dev, follow PR #5381 and after that, you can start rewriting them.

Sorry it takes so long.

PR #5482, opened 2025-05-18
Fix Row Size Too Large error in Version20210719123928 migration

simwai:
## Description
This PR fixes a bug where `Version20210719123928` migration fails on some MySQL/MariaDB installations with the error "Row size too large". The problem occurs because the combined column sizes exceed the maximum row size limit in MySQL/MariaDB.

Although there was no GitHub issue for this specific problem, we encountered it during deployment on MySQL 8.3 and MariaDB 10.3.39 installations.

The fix modifies the migration to:
1. Set ROW_FORMAT=DYNAMIC for affected tables
2. Reduce VARCHAR column sizes to VARCHAR(64) for affected tables
3. Maintain the original budget_type column additions as intended

## Types of changes

## Checklist

### DIFF
--- a/migrations/Version20210719123928.php
+++ b/migrations/Version20210719123928.php
@@ -21,11 +21,40 @@ final class Version20210719123928 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return 'Adds the budget_type columns to customer, project and activity';
+        return 'Adds the budget_type columns to customer, project and activity and fixes row format issues';
     }
 
     public function up(Schema $schema): void
     {
+        // Fix row format to prevent "Row size too large" errors
+        $this->addSql('ALTER TABLE kimai2_activities ROW_FORMAT=DYNAMIC');
+        $this->addSql('ALTER TABLE kimai2_customers ROW_FORMAT=DYNAMIC');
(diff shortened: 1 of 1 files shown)


simwai:
Hereby, I accept the CLA:

kevinpapst (changes requested):
Thanks for taking the time to send the PR.
In this case it would be better to create an issue first.
The migration has its date in the filename, it's `2021-07-19`.
This migration ran thousands of times without issues.

I do acknowledge that you had issues, but the chosen "fix" just doesn't work.

kevinpapst:
Closing, as the fix is broken in various ways. Eg the Entites and validations were not changed.
It will lead to different database structures between existing and new installations....

kevinpapst:
You are still welcome to create an issue, so we can talk about it and others can add their thoughts.

simwai:
I would open an issue, but I don't have more details. I deployed this fix live in production, and it worked. I don't know why you say it is inconsistent and broken, but I assume you have your reasons.

kevinpapst:
> I don't know why you say it is inconsistent and broken, but I assume you have your reasons.

I already said it: e.g. the Entites and validations were not changed. If someone tries to save a company with a name of 70 chars, it will now fail with an exception.

Or: if there were rows existing in the database with names longer than 64 chars, the migration can fail....

PR #5603, opened 2025-08-08
[Feature] ICS calendar support

MitchellArtz:
## Description
This pull request introduces a new database migration to support ICS calendar sources and includes updates to the build artifacts. The most significant change is the addition of a new table for ICS calendar sources, which enables users to link external calendar feeds to their accounts.

**Database Migration:**

* Added new migration file `Version20250807232208.php` to create the `kimai2_ics_calendar_sources` table, including fields for user association, name, URL, color, enabled status, and last sync timestamp, with appropriate foreign key constraints.

**Build Artifacts:**

## Types of changes

## Checklist


MitchellArtz:
Note I am having difficulty building for prod and running unit tests. Any help (or even running them for me) would be appreciated. I still need to add additional tests and work more on error handling, however, I wanted to quickly see if integration tests or codecov could run.

templates/calendar/user.html.twig
```diff
@@ -214,6 +281,168 @@
             toolbar[2].classList.add('col-md-4', 'col-6', 'mb-1', 'text-end', 'order-md-4', 'order-xs-2');
 
             KimaiReloadPageWidget.create('kimai.systemConfigUpdate', true);
+
+            // Store calendar instance globally for ICS functions
+            console.log('Storing calendar instance:', calendar);
+            console.log('Calendar methods:', Object.getOwnPropertyNames(calendar));
+            console.log('Has getEventSources:', typeof calendar.getEventSources);
```
kevinpapst: remove

templates/calendar/user.html.twig
```diff
@@ -214,6 +281,168 @@
...
+            };
+        
+        }); // End of kimai.initialized event listener
+
+        // Global toggle function that can access the stored calendar instance
+        window.toggleICSCalendar = function(icsId, enabled) {
+            console.log('Toggle called with:', icsId, enabled);
+            console.log('kimaiCalendar exists:', !!window.kimaiCalendar);
+            console.log('kimaiCalendar type:', typeof window.kimaiCalendar);
+            console.log('getEventSources method:', typeof (window.kimaiCalendar && window.kimaiCalendar.getEventSources));
```
kevinpapst: ...

src/API/ICSCalendarController.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    #[Route(path: '/{id}/events', name: 'api_calendar_ics_events', methods: ['GET'])]
+    public function getEventsAction(ICSCalendarSource $source, Request $request): Response
+    {
+        $user = $this->getUser();
+        
+        if ($source->getUser() !== $user) {
+            throw $this->createAccessDeniedException('You can only access your own ICS calendar sources.');
+        }
```
kevinpapst: teamleads can view the calendar of their team member

src/Calendar/CalendarService.php
```diff
@@ -89,6 +94,12 @@ public function getSources(User $user): array
     {
         $sources = [];
 
+        // Add ICS calendar sources
+        $icsSources = $this->icsSource->getSourcesForUser($user);
+        foreach ($icsSources as $source) {
+            $sources[] = $source;
+        }
```
kevinpapst: there are calendar events that can be used to attach more sources

kevinpapst (changes requested):
Thanks for sharing your idea!

There is a lot going on in this PR. 

If this would go into core, I'd have a lot of questions. E.g. why are the assets rebuild if there is no change in package.json or any JS or SCSS file. Only the maintainer should push rebuild assets and lockfiles.

Let's clean it up together to see how it looks at the end.

First, please remove the following files from the PR:
- everything in public/build/ 
- .pnp.*
- yarn.lock

Also all debug `console.log` in the normal flow (error logs are ok).

What comes immediately to my mind:
- I would use a system configuration to store one ICS file for company events/holidays. 
- If you want to support user specific ICS calendars, you can store one URL in a user preference.
- That removes the need for a new DB table and allows for aggressive caching of the file contents. 
- Add validation for the URLs (download/filesize/is it a ICS file on saving preference, not when displaying)

Using System-Configuration and user preferences gives you automatically generated forms to set and remove them => you could remove a lot of your code.

I don't see why we need a new DB table/entity/repository/controller for that. The most often used case will be the company calendar. Maybe some user will add their own calendar. But having multiple calendars is likely a very rare case. 

Each line of code that can be removed is a good line!

I tend to say, that this entire thing should be a plugin, not a core feature. But if we can get it down to a very reasonable size, I might be willing to merge and maintain it long term for the community. 

> Note I am having difficulty building for prod and running unit tests. Any help (or even running them for me) would be appreciated. 

For tests and codestyles:
```
composer tests
./phpstan.sh
./php-cs-fixer.sh
```
I cannot see a reason to run the frontend build in the PR. I guess you don't need it. Maybe the developer docs were misleading.

templates/calendar/user.html.twig
MitchellArtz: Debugging lol. Ran into a lot of issues

src/API/ICSCalendarController.php
MitchellArtz: Good note, I think I may need to work on additional functionality for public vs private calendars. Idea was having ability to have all calendars show but specific ones only show for management.

MitchellArtz:
Thanks for the feedback!

Regarding multiple calendars, I actually use several Outlook calendars myself to keep recurring meetings, one-off meetings, and personal events separate. In a company environment with multiple teams, I could see individual calendars being used quite a bit, even if not by everyone. Perhaps we could make the default calendar configurable at the system level to accommodate both simpler and more advanced use cases.

Apologies about the extra files in the PR — this is my first time setting up a Kimai development environment, and I struggled to get yarn run prod working per the manual. I ended up running yarn run dev instead, which caused some unintended changes. A lot of the differences in this PR are due to my dev environment being misconfigured and me attempting to run things on Windows.


This feature actually began as a plugin, but I ran into roadblocks. After reviewing discussions and seeing interest from other users, I considered moving it toward core (plus path of least resistance) but I do share your concern about potential impact on core functionality. I plan to do checks over the next month or two. If you’d prefer it to remain a plugin and could offer some guidance on hooking it in, I’d be happy to pursue that route and make it free. I also think that making it optional is better than mandatory. I'd much rather it be a plugin, but I cannot find a way to do that without some core support.


More of a proof of concept, but any advice would be awesome! I head to bed now at.... Oh my, 0230 EST.



Edit:

"I would use a system configuration to store one ICS file for company events/holidays"


Should this perhaps be a future integration into the time off / absence bundle? With an option to make it paid, or simply display? Perhaps global (or teams-based) and user-level calendars may be best. Overall, I wish to leave it open-ended. Even just to spur ideas. Plugin would be awesome for that, so long as proper modifications to core can be made to accommodate.



"you can store one URL in a user preference."


Isn't user preference key value? if users have multiple values how would we deal with it? Perhaps my experience is an edge case and I should suck it up if multiple calendars aren't supported.


Less is better and I'll work to compact it. May take some iterations!


Cheers,
Mitchell

MitchellArtz:
"I would use a system configuration to store one ICS file for company events/holidays and such and not user specific once. That removes the need for a new DB table and allows for aggressive caching of the file contents.

If you want to support user specific ICS calendars, you can store one URL in a user preference.

Using System-Configuration and user preferences gives you automatically generated forms to set and remove them => you could remove a lot of your code.

I don't see why we need a full DB table for that. The most often used case will be the company calendar. Maybe some user will add its own calendar. But having multiple calendars? I doub't it."




**This is a good note. For now, I can support one-two calendars. I'll save the code somewhere and refactor to have one calendar per user and one per company. Cuts the amount of code significantly and allows for future expansion. Easier merge and less of a headache to debug. TODO soon.**

Should I allow static ICS uploads? I know you're working on file uploads, so I don't want to cross paths for now. Perhaps stay with the refresh system?

kevinpapst:
> Should I allow static ICS uploads? 

No. Remove as many (edge) use-cases as possible.

I changed my initial review a bit, as I wrote it in a hurry before. My main points stands: one global calendar, one calendar per user, remove most code, add validation(!) for the calendar URLs

.gitignore
```diff
@@ -62,3 +62,14 @@ yarn-error.log
...
+public/build/
+
+# Exclude Yarn Plug'n'Play files
+.pnp.*
+
+# Exclude yarn.lock
+yarn.lock
+/public/build
+/public/build/*
+*.lock
```
kevinpapst, line 75: revert, that is wrong

kevinpapst:
Sorry for not mentioning that earlier, but now that I see that there was already a if for ical sources.... did you try to just pass the URLs to the ical parser of Fullcalendar? 

See docs at https://fullcalendar.io/docs/icalendar

src/Calendar/IcalSource.php
```diff
@@ -0,0 +1,126 @@
...
+            ''
+        );
+    }
+
+    public static function createUserSource(IcsValidator $icsValidator, SystemConfiguration $configuration, User $user, LoggerInterface $logger): ?self
+    {
+        $userIcalLink = $configuration->getCalendarUserIcalLink($user);
+        
+        if (empty($userIcalLink)) {
+            $logger->info('IcalSource: No user ICAL link configured', [
```
kevinpapst, line 109: expected behavior does not need a log. but if you want one, then use debug

src/Controller/IcalController.php
```diff
@@ -0,0 +1,137 @@
...
+        $r = hexdec(substr($hexColor, 0, 2));
+        $g = hexdec(substr($hexColor, 2, 2));
+        $b = hexdec(substr($hexColor, 4, 2));
+        
+        // Calculate luminance
+        $luminance = (0.299 * $r + 0.587 * $g + 0.114 * $b) / 255;
+        
+        // Return black or white based on luminance
+        return $luminance > 0.5 ? '#000000' : '#ffffff';
+    }
```
kevinpapst, line 136: Use `App\Utils\Color`

src/Entity/UserPreference.php
```diff
@@ -51,6 +51,7 @@ class UserPreference
     public const PUBLIC_HOLIDAY_GROUP = 'public_holiday_group';
     public const HOLIDAYS_PER_YEAR = 'holidays';
     public const WORK_CONTRACT_TYPE = 'work_contract_type';
+    public const USER_ICAL_LINK = 'user_ical_link';
```
kevinpapst, line 54: that's a user preference, the prefix `user_` can be removed

kevinpapst (changes requested):
just a few minor things i noticed while scrolling. would love to see your feedback on the ical fullcalendar implementation first, before doing a real review.

kevinpapst:
Thanks for taking the time to prepare this. 
As this is not moving forward, I am closing. 
Fell free to re-open if you want to continue working on it.

PR #5644, opened 2025-09-23
fix user update during LDAP authentication

nykopol:
## Description
This PR fix an LDAP authentication problem. 
During LDAP authentication, the first phase consist of checking for user existence and binding with his password. 
Ones the binding is correct, the second phase is to update the user in database to reflect the LDAP informations. 

The problem is that the filters applied during the second phase where not the sames as in the first step.
This was leading to impossibly update the user although the LDAP binding was good. 
For exemple it was the case if an other entity with the same name outside the configured scope.

This PR consist of applying the same filters during LDAP binding and user update.  
It also removed a duplicated line in the findUserByUsername function.

## Types of changes

## Checklist

### DIFF
--- a/src/Ldap/LdapManager.php
+++ b/src/Ldap/LdapManager.php
@@ -36,7 +36,6 @@ class LdapManager
 
         $criteria = [$params['usernameAttribute'] => $username];
 
-        $params = $this->config->getUserParameters();
         $filter = $this->buildFilter($criteria);
         $entries = $this->driver->search($params['baseDn'], $filter);
 
@@ -89,7 +88,9 @@ class LdapManager
         $user->setPreferenceValue('ldap_dn', $baseDn);
 
         $params = $this->config->getUserParameters();
-        $entries = $this->driver->search($baseDn, $params['attributesFilter']);
+        $criteria = [$params['usernameAttribute'] => $user->getUserIdentifier()];
+        $filter = $this->buildFilter($criteria);
+        $entries = $this->driver->search($params['baseDn'], $filter);
 
         if ($entries['count'] > 1) {
             throw new LdapDriverException('This search must only return a single user');


kevinpapst:
Can you add more examples, where you explain why you can't work with simply changing your Kimai configuration?
Also please add some LDAP examples, so I understand what problem you are trying to address here.

Many users successfully connect Kimai and LDAP and even though I have limited experience, I know that it usually works with the correct configuration.

nykopol:
My LDAP configuration :

```
kimai:
    ldap:
        activate: true
        connection:
            host: 172.16.5.1
            username: CN=kimai,OU=Technical-Accounts,OU=COMPANY-NAME,DC=domain,DC=local
            password: ...
        user:
            baseDn: OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
            usernameAttribute: sAMAccountName
            filter: (&(memberOf=CN=Kimai-Users,OU=Security,OU=Groups,OU=COMPANY-NAME,DC=domain,DC=local))
            attributes:
                - { ldap_attr: "samaccountname", user_method: setUserIdentifier }
                - { ldap_attr: "mail", user_method: setEmail }
                - { ldap_attr: cn, user_method: setAlias }
```

During the first phase (binding user to validate his password), the `findUserByUsername()` function use the intended filter to retrieve the user :

```
// Line 41
$entries = $this->driver->search($params['baseDn'], $filter);
// $params['baseDn']    = OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
// $filter              = (&(&(memberOf=CN=Kimai-Users,OU=Security,OU=Groups,OU=COMPANY-NAME,DC=domain,DC=local))(sAMAccountName=HomerSimpson))
```

This search return only on result with the DN : `CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local`

At this stage, the user exist, is unique and can be bind with the correct password. 

But during the update of the user in `updateUser()` the filters are not respected.

```
// Line 92
$entries = $this->driver->search($baseDn, $params['attributesFilter']);
// baseDn                       = CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
// $params['attributesFilter']  = (objectClass=*)  -- HERE IS THE PROBLEM
```
As the filters are not respected, this can lead to matching more than on entity. 
In my case, it return the correct user but also Exchange object linked to this user with DN like
* CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
* CN=ExchangeActiveSyncDevices,CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
* CN=iPhoneXYZ987,CN=ExchangeActiveSyncDevices,CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local
* CN=iPhoneABC123,CN=ExchangeActiveSyncDevices,CN=Homer Simpson,OU=Local Team,OU=Users,OU=COMPANY-NAME,DC=domain,DC=local

So at line 94 as there is more than one entity, an exception is throw and the user is facing an invalid password error.

My proposed fix is to use the exact sames parameters to fetch the user in `findUserByUsername()` and in `updateUser()`.

**EDIT :** After some digging I found that it's related to https://github.com/kimai/kimai/issues/875 and the `attributesFilter` parameter. My question is why should we use two different filter to retrieve the same ldap object?

kevinpapst:
Does this mean you can login without a code change?
In that case and to prevent any BC break, I would close this PR.

> My question is why should we use two different filter to retrieve the same ldap object?

I don't know. I cannot remember every piece of code in Kimai or why certain decisions were made. Especially in the LDAP part, which I never used myself.

PR #5689, opened 2025-11-10
Fix MariaDB enum type error and missing OpenApi dependency

Fellurion:
## Description
This PR fixes two critical issues that prevent migrations from running successfully after updating to Kimai

1. **ENUM type error:**
```
   Unknown database type enum requested, 
   Doctrine\DBAL\Platforms\MariaDb1010Platform may not support it.
```

2. **Missing class error:**
```
   Class "OpenApi\Processors\Concerns\TypesTrait" not found while loading 
   "Nelmio\ApiDocBundle\RouteDescriber\RouteArgumentDescriber\SymfonyMapQueryParameterDescriber"
```

### 1. Added enum mapping in `doctrine.yaml`
Added `mapping_types` configuration to map MariaDB's `enum` type to Doctrine's `string` type. This is the standard solution for Doctrine's limited MariaDB enum support.

### 2. Added `zircote/swagger-php` dependency
Added explicit dependency on `zircote/swagger-php ^4.0` which is required by `nelmio/api-doc-bundle` 5.0 but was not properly pulled in.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,8 @@
         "twig/extra-bundle": "^3.0",
         "twig/inky-extra": "^3.0",
         "twig/intl-extra": "^3.0",
-        "twig/string-extra": "^3.0"
+        "twig/string-extra": "^3.0",
+        "zircote/swagger-php": "^4.0"
     },
     "require-dev": {
         "ext-simplexml": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "0e017e736364638de484ae1799d11b8f",
+    "content-hash": "72e56498e44902f4d95472521e3a6028",
     "packages": [
         {
             "name": "azuyalabs/yasumi",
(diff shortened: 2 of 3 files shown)


Fellurion:
I fixed these errors that occurred when I tried to update to the new version. 
The errors appeared during both the Kimai update process and when running composer install command from the doku
 i use MariaDB with version 10.11.8 and php 8.4.

kevinpapst:
Thanks 👍 

The composer package issue might have been caused by the upstream API doc package. There was a bugfix release that mentioned this trait.

Can you try to update to the latest release and see if it works without your changes?

Also: there is no ENUM used in the Doctrine/Entity part of Kimai. Are you running custom plugins?

Fellurion:
> Thanks 👍
> 
> The composer package issue might have been caused by the upstream API doc package. There was a bugfix release that mentioned this trait.
> 
> Can you try to update to the latest release and see if it works without your changes?
> 
> Also: there is no ENUM used in the Doctrine/Entity part of Kimai. Are you running custom plugins?



No custom plugins – just the latest version on a clean dev installation.
I encountered this issue in my company's development environment and wanted to report it in case others run into the same problem. My setup: Ubuntu, MariaDB (native), PHP 8.4, and Symfony dev server – all up to date. I always test on a clean installation without plugins before testing my own plugins on new versions.
Just pulled the latest changes into PHPStorm

kevinpapst:
The missing dependency was fixed upstream.

I cannot reproduce the ENUM issues. Seems nobody else did as well.
Check if you upgraded the `serverVersion` parameter in your `.env`?
```
DATABASE_URL=mysql://kimai:kimai@127.0.0.1:3306/kimai2?charset=utf8mb4&serverVersion=10.11.15-MariaDB
```

PR #5795, opened 2026-01-27
Integrate WorkingTimeSpanBundle into core

ralf1070:
Alternative working time calculation that merges overlapping time entries into spans instead of summing durations (prevents double-counting).

- Add TimeSpanCalculator service to App\WorkingTime
- Add WorkingTimeSpanSubscriber for WorkingTimeYearEvent
- Add WorkingTimeSpanConfigurationSubscriber for system settings
- Add translations (de, en) for configuration options

## Description

## Types of changes

## Checklist

### DIFF
--- a/src/EventSubscriber/WorkingTimeSpanConfigurationSubscriber.php
+++ b/src/EventSubscriber/WorkingTimeSpanConfigurationSubscriber.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\EventSubscriber;
+
+use App\Event\SystemConfigurationEvent;
+use App\Form\Model\Configuration;
+use App\Form\Model\SystemConfiguration;
+use App\Form\Type\YesNoType;
+use App\WorkingTime\TimeSpanCalculator;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\Extension\Core\Type\IntegerType;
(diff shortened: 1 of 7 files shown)


ralf1070:
Added some unit tests

kevinpapst:
My plan was to add a new column to the timesheets table, storing the unrounded duration.
What I want to avoid is iterating and comparing all available timesheets in real-time.
In the future there should be an listing table for all registered users, to see high-level stats, and that would likely overload the database and/or memory if we do live calculations for thousands of timesheets.

Totally appreciate the initiative 👍 just thinking about how to integrate that "safely".

ralf1070:
Generally a good idea - i just don't know if the timesheet table is the best place to store that. I know - you don't believe in parallel tasks - but as your system supports them, I like the idea to not waste this capability. And it is useful - really. I can't see how it would be possible to honor overlapping tasks by storing the actual duration of every single timesheet entry.

So maybe you should cache the daily statistics - once it is computed - somewhere else, to not trigger your overload scenario. On the other hand - you will have a cache invalidation problem then...

I actually don't know

ralf1070:
Slept over it - daily caching could actually help:
A table where you can look up records with relevant data for every user for every day - day not task.
Updated when there is no record available for this day or the record has been invalidated.
On each insert or edit of a time sheet record you invalidate (or simply delete) the records for days around the edited day - to allow for records which are counted for other days.

This way you allow for more complex computations of working time and gain speed for statistics.

Edits of older entries should be seldom - so once a cache record has been created, it should usually be valid kind of forever.

Maybe you add a way to use custom attributes for these records and provide a nice hook when these records are updated - so even more complex statistics can be added in plugins... could be fun.

PR #5801, opened 2026-01-29
Add custom attributes to DayAddon for flexible plugin rendering

ralf1070:
Extend DayAddon with a generic attribute system that allows plugins to customize rendering in the working time overview:

- setAttribute()/getAttribute()/hasAttribute()/getAttributes() methods
- Template renders 'icon' attribute as Font Awesome icon
- Template renders 'color' attribute (hex values or CSS classes)
- Template renders 'comment' attribute in tooltip
- Skip weekend/missing-work highlighting for days with addons

This enables plugins like WorkContractBundle and others to display absences, holidays, and other day markers with distinct icons and colors.

## Description

## Types of changes

## Checklist


ralf1070:
## Summary

This PR extends the `DayAddon` class with a generic attribute system that allows plugins to customize how day markers are rendered in the working time overview.

## Motivation

Currently, all `DayAddon` entries (absences, holidays, etc.) are rendered identically as blue dots in the working time overview. Plugins like WorkContractBundle need to visually distinguish between different types of entries (vacation, sickness, public holidays) but have no way to customize the rendering without overriding the entire template.

This change introduces a minimal, backward-compatible attribute system that enables plugins to specify custom icons, colors, and tooltip text.

## Changes

### DayAddon.php
- Added `$attributes` array property
- Added `setAttribute(string $key, mixed $value): self`
- Added `getAttribute(string $key, mixed $default = null): mixed`
- Added `getAttributes(): array`
- Added `hasAttribute(string $key): bool`

### status.html.twig
- Render `icon` attribute as Font Awesome icon (if set)
- Render `color` attribute as CSS color (supports hex values and CSS classes)
- Append `comment` attribute to tooltip text
- Skip weekend/missing-work background highlighting for days with addons

## Usage Example

```php
$addon = new DayAddon('holiday', $duration, $visibleDuration);
$addon->setAttribute('icon', 'fas fa-umbrella-beach');
$addon->setAttribute('color', '#2fb344');
$addon->setAttribute('comment', 'Summer vacation');
$day->addAddon($addon);
```

## Backward Compatibility

- Fully backward compatible
- Existing plugins work without modification (default: blue dot, no icon)
- All attributes are optional

kevinpapst:
Hello, your PR came in when I was in the middle of changing the Working Contract view. 
We had more or less the same idea, just a different approach

I am using CSS variables for styling and the [icon names](https://github.com/kimai/kimai/blob/main/config/packages/tabler.yaml#L113) for highlighting the the add-ons. 

Looks like this:
image

Please rebase / check if the lastest master already satisfies your needs.

ralf1070:
I'm currently developing a plugin that adds custom addon types to the working time overview (e.g., "homeoffice", "business trip"). With your new type parameter, the rendering works great - but I ran into two issues:

### Registering custom icons
I tried to register additional icons (homeoffice, business_trip) via prependExtensionConfig('tabler', [...]) in my plugin's Extension class. The config shows up in debug:config tabler icons, but the icons aren't available at runtime.

It seems TablerExtension::prepend() reads the icon config before plugin prepends run. I worked around this with a CompilerPass, but is there a recommended way for plugins to register custom icons?

### Passing additional data (this updated patch)

Beyond icon/color (now handled by type), my plugin needs to pass extra data like comments to the tooltip. This new patch adds the generic attributes array to DayAddon:

  - setAttribute()/getAttribute()/hasAttribute()/getAttributes()
  - Template renders comment attribute in tooltip

This keeps the API flexible for various plugin needs without requiring core changes for each use case.

A simple way to be able to adapt the rendering for each day would be great ...

ralf1070:
If you want to have a look at the plugin (it's still work in progress): https://github.com/ralf1070/RemoteWorkBundle

kevinpapst:
If the frontend needs to know which attributes exists, its not really flexible, is it?!

Why do you need an additional comment? You already have a title to show.

ralf1070:
I try to add work from home and business trip times. For work from home I could live with the title field - but even then I would need to translate the actual title in my code part and not at rendering time in the style sheet - what I don't like.

Business trip is a bit more complex:
I would like to have a markdown comment there - which may need to be rendered differently in different contexts - dynamic Html page , static PDF, Excel, ...

And I would like to have a destination field. Which should be rendered "Destination: 'destination' ". Another string that should use the translation system. One could be inclined to render the destination as a maps link...

And it would be nice to be give information like "half day" to the renderer in an abstract way - and then be able to use all these information in a way which best fits your needs...

ralf1070:
Working on my remote work plugin I refined this patch a little to make it really useful:

**Regarding extensibility concerns:**

The current implementation in `status.html.twig` anticipates features used by WorkContractBundle, but it's tightly coupled - plugins have no extension points except overriding the entire template.

**Why `title` alone is not flexible enough:**

In my RemoteWorkBundle, I need much more than a static title:

- **Linked working time**: I wrap the displayed duration in a link to the timesheet entries for that day
- **Rich tooltips**: I render markdown comments
- **Contextual icons**: For business trips, the icon links to the travel destination
- **Conditional rendering**: Different display for half-day vs. full-day entries (homeoffice/business-trip)

None of this is possible with just a plain text `title` attribute. The template needs to know *how* to render, not just *what* to display.

**This solution:**

By splitting the rendering into overridable templates (`_status_day.html.twig`, `_status_day_icon.html.twig`, `_status_day_content.html.twig`, `_status_day_tooltip.html.twig`) and adding generic attributes to `DayAddon`, plugins can:

1. Override only the specific part they need to customize
2. Store arbitrary data (comment, destination, half_day flag, etc.) that their templates can access
3. Keep full backwards compatibility - the default templates work exactly as before

This makes the frontend truly flexible: it doesn't need to know *which* attributes exist, it just provides the hooks for plugins to render them as needed.

For what I think - it would be an improvement.

ralf1070:
> If you want to have a look at the plugin (it's still work in progress): https://github.com/ralf1070/RemoteWorkBundle

@kevinpapst 
Side note:
I set the RemoteWork repo to private yesterday and invited you to participate, because there are changes which are on your roadmap to WorkContract and I don't want to interfere with your business model. I explained a little more in the project discussions page. If you think there is no problem for you with this plugin I will reopen it for free use again.

kevinpapst:
> because there are changes which are on your roadmap to WorkContract and I don't want to interfere with your business model 

I am on a holiday trip and did not want to start this discussion before I left.
But that was actually causing me a few headaches.

Will get back to you after my trip next week.

ralf1070:
Have a great vacation - no headaches.

templates/contract/_status_day.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
```
kevinpapst: Plugins cannot know what other plugins (or the core in future versions) want to achieve.

src/WorkingTime/Model/DayAddon.php
```diff
@@ -12,6 +12,8 @@
 final class DayAddon
 {
     private bool $billable = true;
+    /** @var array<string, mixed> */
+    private array $attributes = [];
```
kevinpapst, line 18: Can you define the required fields?
I am not a fan of arbitrary attributes, so adding a `comment` field for example and a `half-day` one might be better.

kevinpapst (changes requested):
Plugins cannot know what other plugins (or the core in future versions) want to achieve.
So I do not really like this way of extendability via includes 

But the general idea is okay. Lets do it in a different way and use dynamic blocks from dynamic includes, see https://github.com/kimai/kimai/blob/main/templates/calendar/user.html.twig#L42 as an example.

You can then define these blocks in your plugin.
We would need the optional `rendererBlock` and `rendererInclude` fields in the DayAddon for that.

That would allows plugins to define how their own type is rendered, without influencing other types.

ralf1070:
Good point.

Currently I use these attributes:

- `half_day` (bool) and `comment` (string) — these could be useful for other plugins too. `comment` is rendered as markdown - so title is not a substitude.
- `destination`, `own_vehicle`, `distance`, `sick_note` — plugin-specific, only consumed by the plugin's own renderer template

With the `rendererBlock`/`rendererInclude` concept, the plugin controls both sides: it sets the data in its EventSubscriber and renders it in its own Twig block. The attributes are just a transport mechanism between plugin code and plugin template — the core never reads them.

I'd argue that the attributes bag should stay as-is — it's already as minimal as it can be while still being functional. Adding explicit fields for `comment` and `halfDay` alongside attributes would mean maintaining two mechanisms for the same thing.

A cleaner approach: define core constants for well-known attribute keys (e.g. `ATTR_COMMENT`, `ATTR_HALF_DAY`) and add convenience getters/setters that delegate to the attributes bag internally. That way there's one storage mechanism, a discoverable API for common attributes, and plugins can still pass custom data to their own renderer blocks without the core needing to know about it.

In fact, the existing fields on DayAddon could be handled the same way - if you handle DayAddon as a dynamic container to transport data to a renderer.

One limitation I noticed: with the `rendererBlock`/`rendererInclude` pattern, a plugin can only customize rendering of its *own* addons. If plugin A enriches plugin B's data with additional attributes (e.g. my remote-work plugin adds a `sick_note` flag - "Krankenschein vorhanden" - to the work-contract plugin's sickness days), there's no way to influence how that addon is rendered — plugin A's renderer is not called for B's data. Not sure if that's a real problem or an acceptable trade-off.

I also added a small practical improvement in this PR: each day cell in the working time overview now links to the timesheet filtered by that day. Previously there was no way to quickly jump from the overview to a specific day's time entries. In the previous concept it was possible to do this by modifing the responsible twig file - using this concept it must be done in core - if you think your are interested in it.

I've updated the existing PR to use the dynamic block approach as you suggested. Is this ok for you?

src/WorkingTime/Model/DayAddon.php
ralf1070: Copy of comment in PR:

Currently I use these attributes:

- `half_day` (bool) and `comment` (string) — these could be useful for other plugins too
- `destination`, `own_vehicle`, `distance`, `sick_note` — plugin-specific, only consumed by the plugin's own renderer template

With the `rendererBlock`/`rendererInclude` concept, the plugin controls both sides: it sets the data in its EventSubscriber and renders it in its own Twig block. The attributes are just a transport mechanism between plugin code and plugin template — the core never reads them.

You would not want to add every attribute other plugin author could think about...

`destination`, `distance` and `own_vehicle` are for business trip - `own_vehicle` is used to render a small badge to the icon. our secretariat ask for it. they have to know if they have to pay for the use of an own car. about the same for `sick_note` (Krankenschein vorhanden) - they need it for their statistics and were really happy to see this in their overview. `destination` is used to render the car icon of business trip as link to google maps ...

PR #5808, opened 2026-02-02
add column summarization per customer

GregorB54321:
grafik
## Description
add column summarization per customer
feel free to adapt or make it optional; all changes where created via GH-Copilot-prompting...
I will use it to have daily working-times per customer (I have "pause/breaks" as separate customer which is a suitable workaround).

## Types of changes

## Checklist


kevinpapst (approved):
Thanks, will take it as is and improve it if needed.

PR #5810, opened 2026-02-03
preserve fixed rate on activity change (#5735)

### LINKED ISSUE #5735: Fixed price automatically removed on activity change
### Describe the issue

Example:
I start an activity with one hour, where hourly price is defined as 50.
On that activity i set a fixed price of 80 and save that.
The list of activities shows correctly a price of 80.
I now edit that entry again and change the activity to another one and also check the field for the fixed price, which still shows 80 and hit save.
The list of activities now shows price 50, instead of 80.
When I open that activity again, the fixed price field is empty and those 80 have vanished.

Funny thing is, that when I change the activity on an existing entry and ALSO modify the fixed price to lets say 81, that 81 gets saved correctly.

### I already tried

### Kimai version

2.44.0

### How do you run Kimai?

Virtual Server or alike

Unknown

### Logfile

_No response_

### Screenshots

_No response_

faruk-agentiqus:
Fixes #5735

**Problem:** Fixed price removed when changing activity without modifying the rate field.

**Solution:** Preserve `fixedRate`/`hourlyRate` when already set on the record.

**Changes:**
- `RateResetCalculator.php`: Skip reset if manual rate exists
- `RateResetCalculatorTest.php`: Added tests for rate preservation

### DIFF
--- a/src/Timesheet/Calculator/RateResetCalculator.php
+++ b/src/Timesheet/Calculator/RateResetCalculator.php
@@ -23,6 +23,13 @@ final class RateResetCalculator implements CalculatorInterface
             }
         }
 
+        // if the record already has a manually set rate (fixedRate or hourlyRate),
+        // preserve it when project/activity/user changes - the user explicitly set
+        // this rate before and chose not to change it (see GitHub issue #5735)
+        if ($record->getFixedRate() !== null || $record->getHourlyRate() !== null) {
+            return;
+        }
+
         // if no manual rate changed was applied:
         // check if a field changed, that is relevant for the rate calculation
         // reset all rates, because most users do not even see their rates and would not be able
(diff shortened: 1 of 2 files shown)


src/Timesheet/Calculator/RateResetCalculator.php
```diff
@@ -23,6 +23,13 @@ public function calculate(Timesheet $record, array $changeset): void
             }
         }
 
+        // if the record already has a manually set rate (fixedRate or hourlyRate),
+        // preserve it when project/activity/user changes - the user explicitly set
+        // this rate before and chose not to change it (see GitHub issue #5735)
+        if ($record->getFixedRate() !== null || $record->getHourlyRate() !== null) {
```
kevinpapst, line 29: In my opinion this is not yet sufficient: the fixed rate could have been created by a rule from the previous selected activity. In that case we likely do not want to keep it.

That is especially true for the hourly rate. Fixed rates are used less often.

kevinpapst:
Closing due to missing feedback. 
Please re-open if you have further input.

PR #5811, opened 2026-02-03
close form select after value has been selected (#5716)

### LINKED ISSUE #5716: Close form select after value has been selected
### Describe the issue

I find it very annoying that the dropdown stays open when you select a value. This is especially annoying when you want to create an invoice where I want to create an invoice for only ONE customer.

I found out that you use `tom-select` as a library but also there is the issue that in multi-selects the dropdown stays open and you need to click outside to close it. Would it be possible to handle that with additional events?

### I already tried

### Kimai version

2.44.0

### How do you run Kimai?

Virtual Server or alike

8.4

### Logfile

_No response_

### Screenshots

_No response_

faruk-agentiqus:
## Summary
- Adds `closeAfterSelect: true` option to tom-select configuration
- Fixes dropdown staying open after selecting a value in multi-select fields (like Activity, Customer on invoice page)

Fixes #5716

## Test plan
- [x] Navigate to invoice page (/en/invoice/)
- [x] Select a customer to populate Activity dropdown
- [x] Click Activity dropdown and select an option
- [x] Verify dropdown closes immediately after selection (instead of staying open)


kevinpapst (changes requested):
I am not convinced this is the correct behavior, but I will test it.

Anyway: 
I cannot not accept PRs that include re-compiled assets. 
Please remove everything, except the javascript changes in assets/.

PR #5814, opened 2026-02-04
Feature/customer rate factor

joney94:
## Description
introduces a customer-specific rate factor and enhanced invoice display options for overhead costs.

- _Customer Rate Factor_: Adds a new rate_factor field to the Customer entity. This factor is automatically applied as a multiplier to all hourly and fixed rates during time tracking.
- _Use case_: E.g. for containing for individual overhead costs or bulk discount for customers - without the need to change all hourly rates by hand.
- Optional choice whether to apply the factor to fixed prices as well

- _Invoice Overhead Display_:
A new option in the invoice toolbar allows users to choose how the rate factor overhead is presented.
  a) Integrated: The overhead is part of the individual line items (scaled rate).
   b) Separate: Line items show the base price, and a new summary row "Overhead through customer rate factor" is added at the end of the invoice to account for the difference.

- _Translation Support_: Includes German and English translations for all new fields and labels, but should be Weblate Translateable.

@kevinpapst: thanks for your great work! Feedback ist most welcome :)

### Technical
- Keep consistency by using FLOAT for the rate factor, matching Kimai's internal budget and rate calculation logic.
- Sticked with centralized calculation logic in AbstractCalculator to support all standard invoice templates.
- Tests are incluced

## Types of changes
- [x] New (small) feature (non-breaking change which adds functionality)

## Checklist
@kevinpapst --> sorry the link didn't work - what's the location to update the documentation?

### DIFF
--- a/migrations/Version20260204180102.php
+++ b/migrations/Version20260204180102.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace DoctrineMigrations;
+
+use App\Doctrine\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * @version 2.x
+ */
+final class Version20260204180102 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add rate_factor column to customer table';
+    }
+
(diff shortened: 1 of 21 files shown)


kevinpapst:
Not interested any more?

joney94:
Hi Kevin, we just moved the repo to private to do some Uni-specific customizations. But is the feature of interest for you, or rather not? 
Because if yes, I could repost those "rate factor" lines?

kevinpapst:
Could you share some screenshots?

PR #5842, opened 2026-02-26
Fix off-by-one in stopActiveEntries when starting new timesheet

ralf1070:
## Summary

`stopActiveEntries()` has an off-by-one error that allows more concurrent timesheets than `hard_limit` permits.

The method is called **before** the new timesheet is persisted, so `getActiveEntries()` does not include the upcoming entry. The current calculation `count($activeEntries) - $hardLimit` therefore underestimates by one — it checks whether there are *too many* entries, but doesn't account for the one about to be added.

**Example with `hard_limit = 3`:**
- 3 entries already running
- User starts a 4th via "Restart recent activity"
- `needsStop = 3 - 3 = 0` → nothing stopped
- New entry saved → **4 active entries** (exceeds limit)

**Fix:** `needsStop = count($activeEntries) - $hardLimit + 1`

This makes room for the incoming entry. The existing `$timesheet->getId() !== $activeEntry->getId()` guard still correctly skips the new (unpersisted) entry if it somehow appears in the list.

With `hard_limit = 1` (the default), the bug is less visible because most UI flows implicitly stop the previous entry through other mechanisms. It becomes clearly reproducible with `hard_limit > 1` when using "Restart one of your recent activities".

## Test plan

- [x] New unit test `testSaveNewTimesheetStopsOldestWhenAtHardLimit` — verifies that starting a new timesheet with one already running at `hard_limit = 1` stops the existing entry
- [x] Existing test `testSaveNewTimesheetStopsActiveRecords` still passes
- [x] Manual test: set `hard_limit = 3`, start 3 timesheets, restart a recent activity → oldest is stopped, count stays at 3
- [ ] 
## Types of changes

## Checklist

### DIFF
--- a/src/Timesheet/TimesheetService.php
+++ b/src/Timesheet/TimesheetService.php
@@ -301,7 +301,8 @@ final class TimesheetService
         }
 
         $activeEntries = array_reverse($activeEntries);
-        $needsStop = \count($activeEntries) - $hardLimit;
+        // +1 because the new timesheet is not yet persisted and not in $activeEntries
+        $needsStop = \count($activeEntries) - $hardLimit + 1;
         $counter = 0;
 
         foreach ($activeEntries as $activeEntry) {
--- a/tests/Timesheet/TimesheetServiceTest.php
+++ b/tests/Timesheet/TimesheetServiceTest.php
@@ -138,6 +138,28 @@ class TimesheetServiceTest extends TestCase
         $sut->saveTimesheet($newTimesheet);
     }
 
+    public function testSaveNewTimesheetStopsOldestWhenAtHardLimit(): void
+    {
+        $authorizationChecker = $this->createMock(AuthorizationCheckerInterface::class);
+        $authorizationChecker->expects($this->once())->method('isGranted')->willReturn(true);
+
+        $activeEntry = $this->createMock(Timesheet::class);
+        $activeEntry->method('getId')->willReturn(1);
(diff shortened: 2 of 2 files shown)


kevinpapst:
This is a deliberate choice, not a bug. 
We cannot stop randomly one entry, the user needs to do that manually.
Parallel timesheets are pretty uncommon use-case, so we keep it as is.

kevinpapst:
Thanks for your participation! 

Next time, let's discuss before doing the deep dive into the code.

ralf1070:
It was actually a chance discovery.

I was testing on the devel branch for v3 and noticed that I could actually start four jobs with "start one of your last activities" instead of the set three. However, as soon as I started the fifth one, one of the jobs that was already running was killed. The fix was simple, which is why I made the pull request. But I just tested it again with v2, and the bug isn't there because of the different call sequence - it's only a v3 "feature".

But randomly killing jobs is an existing kimai feature...

ralf1070:
Do you want me to reopen this PR for v3 then or is this done for you?

kevinpapst:
Honestly: nobody talked about that in years, so I assume it is stable enough. 
I can't remember when I touched that the last time.
I assumed Kimai just stops a timesheet if there is exactly one running.

See, I never used that feature myself, so either it was always like that or broke without me noticing.

If something changed in v3, its worth a look.

ralf1070:
It behaves differently on v3 - so it should be worth looking for.

Issue #5834, opened 2026-02-20
Language field for customer


simonschaufi:
### Describe the problem

I have some cronjobs where I want to send my customers an email in their language. Therefor it would be helpful if the customer object would have a language field.

Side note: The language is independent of the country of the address as Switzerland for example has many different languages.

### Describe the solution you'd like

A language dropdown field of all the languages that kimai provides the same way as the user object has a language dropdown. The values should be stored probably as 2 digit language code. A system configuration of a default language if none is provided. That system language can also be used as default for new customers.

This it the API in my cronjobs that I'm currently using:

```php
/** @var string $locale */
$locale = $input->getOption('locale');
\Locale::setDefault($locale);
if ($this->translator instanceof LocaleAwareInterface) {
    $this->translator->setLocale($locale);
}
```

Where locale is "en" for example.

simonschaufi:
If you like this feature, I can implement it.

kevinpapst:
Yes, a `language` field sounds okay. 

Could be used for all kinds of communication, eg.
- in the Customer Portal 
- invoices

I'd accept a PR for that.
Check all attributes in the Customer Entity (e.g. for the export) and make sure it shows up in the API (this field can be included in all requests).

If you want to add support for it in the Importer: 
https://github.com/kevinpapst/ImportBundle/blob/main/Importer/CustomerImporter.php#L27

simonschaufi:
Implemented with #5857

PR #5855, opened 2026-03-05
[FEATURE] Add invoice email address for customer entity

simonschaufi:
## Changelog

- **Invoice email address for customers** — You can now store a dedicated billing email address on each customer, separate from their main contact email. Set it under the customer's *Invoice* settings. The address appears on the customer details page and customer list, is available in invoice templates (`${customer.invoice_email}`), and is exposed through the API. (#5839)

---

Resolves: #5839

## Description
This commit adds a new field "invoiceEmail" for the customer.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [x] I verified that my code applies to the guidelines (`composer code-check`)
- [x] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/tree/master/_documentation))
- [x] I agree that this code is used in Kimai (see [license](https://github.com/kimai/kimai/blob/main/LICENSE))

## Technical Checklist
- [x] Api returns new field (customer details)
- [x] Field is shown in customer list
- [x] Field is shown in customer details
- [x] Field is added to the invoice section in the customer edit form

PR #5857, opened 2026-03-05
[FEATURE] Add customer language

### LINKED ISSUE #5856: Move vatId into "invoice" section when creating/editing a customer
### Describe the problem

Cleanup customer edit form

In my opinion the vat id is invoice related and could be moved to the invoice section (expandable) when creating/editing a customer

simonschaufi:
## Changelog

You can now set a **language per customer**. This can be used when rendering their invoices in the future. A new system-wide default ("Default customer language") is available in the customer settings, and the VAT ID field has moved into the customer's invoice section.

---

Resolves: #5856 and #5834

## Description
Add a new field to set the customer language and move the vatId into the invoice section

## Types of changes
- [ ] Task (non-breaking change with improvements)

## Checklist


simonschaufi:
@kevinpapst is there any chance this can be part of the next release?

simonschaufi:
@kevinpapst can you please consider this for the next release?

kevinpapst:
Lets merge that with #5855 - I would also add the new language field to the customer.
With these two things in place, we can then add the "sending invoice to customer" feature.

simonschaufi:
Yes, this will give room for the language field that is coming up next

simonschaufi:
@kevinpapst I thought this would be ready for merge? Now a new release is released without this change. Would it be possible to merge this before the next release please?

kevinpapst:
This has no benefit on its own, that's why I said:

> Lets merge that with https://github.com/kimai/kimai/pull/5855 - I would also add the new language field to the customer.
> With these two things in place, we can then add the "sending invoice to customer" feature.

I am not merging half done stuff

simonschaufi:
Is there anything else I should do or is this ready for merge?

src/Entity/Customer.php
```diff
@@ -113,6 +113,12 @@ class Customer implements EntityWithMetaFields, EntityWithBudget, CreatedAt
     #[Serializer\Groups(['Default'])]
     #[Exporter\Expose(label: 'country')]
     private ?string $country = null;
+    #[ORM\Column(name: 'language', type: Types::STRING, length: 6, nullable: false)]
```
kevinpapst: Thanks to this I found a bug in invoice and export template entities due to the column length which is too short.

The currently longest possible locale in Kimai seems to be `zh_Hant_TW`.

There seems to be no clear max length recommendations. It would be 14 chars for `ca_ES_VALENCIA` which seems to be the longest "normal" locale, while BCP-47 recommends 35 chars.
What do you think? I'd say lets use the official recommendation length of 35.

simonschaufi: Ok, let's use 35. Will you create a migration for the other language fields?

migrations/Version20260621144402.php
```diff
@@ -0,0 +1,51 @@
...
+WHERE `name` = 'language'
+ORDER BY user_id ASC
+LIMIT 1
+");
+
+        if (empty($language)) {
+            $language = 'en';
+        }
+
+        $this->addSql('ALTER TABLE kimai2_customers ADD language VARCHAR(6) DEFAULT NULL');
```
kevinpapst: 35

migrations/Version20260802183206.php
```diff
@@ -0,0 +1,51 @@
...
+final class Version20260621144402 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add language to customer';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Use the language of the user with the lowest user_id (mostly the admin)
```
kevinpapst, line 27: I don't know how, but users constantly manage to execute migrations multiple times. These safeguards are an easy way to prevent such support requests.
```suggestion
        if ($schema->getTable('kimai2_customers')->hasColumn('language')) {
            return;
        }
        // Use the language of the user with the lowest user_id (mostly the admin)
```

Please add such a call in down as well.

migrations/Version20260621144402.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->addSql('UPDATE kimai2_customers SET language = ?', [$language]);
+        $this->addSql('ALTER TABLE kimai2_customers MODIFY language VARCHAR(6) NOT NULL');
```
kevinpapst: 35

migrations/Version20260621144402.php
kevinpapst: Also, can't we just use 
`ALTER TABLE kimai2_customers ADD language VARCHAR(35) DEFAULT 'en'` to avoid the NULL - NOT NULL loop.

Not sure if using the first users language is the best approach.
We could use the system setting for "default user language" but that can also be completely wrong.

I would say: use english as default and let users handle it by themselves.

src/Configuration/SystemConfiguration.php
```diff
@@ -332,6 +332,11 @@ public function getCustomerDefaultCountry(): string
         return $this->find('defaults.customer.country');
     }
 
+    public function getCustomerDefaultLanguage(): ?string
+    {
+        return $this->find('defaults.customer.language');
```
kevinpapst:
```suggestion
        return $this->find('defaults.customer.language', 'en');
```

And please add it to https://github.com/kimai/kimai/blob/main/src/DependencyInjection/Configuration.php#L589 as well.

migrations/Version20260621144402.php
simonschaufi: > I would say: use english as default and let users handle it by themselves.

The thing is if your language is german and you have 100s of customers, you need to update each one of them which takes a huge amount of time (if not directly possible via sql update). By using the language of the first user chances are relatively high, that the language will be correct and if not, then you also need to update all customer languages but at least we "tried" to guess the right language :smile: 

My suggestion is:

```php
$this->addSql("ALTER TABLE kimai2_customers ADD language VARCHAR(35) DEFAULT 'en'");
$this->addSql('UPDATE kimai2_customers SET language = ?', [$language]);
```

src/Configuration/SystemConfiguration.php
simonschaufi: Done. I've noticed that the timezone, country and other fields are also missing. what is this configuration for?

migrations/Version20260621144402.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return 'Add language to customer';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Use the language of the user with the lowest user_id (mostly the admin)
+        $language = $this->connection->fetchOne("
+SELECT `value`
+FROM kimai2_user_preferences
```
kevinpapst: bitte aus system settings ziehen

migrations/Version20260802183206.php
```diff
@@ -0,0 +1,50 @@
...
+        $language = $this->connection->fetchOne("SELECT `value` FROM kimai2_configuration WHERE `name` = 'defaults.user.language'");
+        if (empty($language)) {
+            $language = 'en';
+        }
+
+        if (!$schema->getTable('kimai2_customers')->hasColumn('language')) {
+            $this->addSql("ALTER TABLE kimai2_customers ADD `language` VARCHAR(35) DEFAULT 'en'");
```
kevinpapst:
```suggestion
            $this->addSql("ALTER TABLE kimai2_customers ADD `language` VARCHAR(35) NOT NULL DEFAULT 'en'");
```

migrations/Version20260802183206.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        // Use the language of the user with the lowest user_id (mostly the admin)
+        $language = $this->connection->fetchOne("SELECT `value` FROM kimai2_configuration WHERE `name` = 'defaults.user.language'");
+        if (empty($language)) {
+            $language = 'en';
+        }
+
+        if (!$schema->getTable('kimai2_customers')->hasColumn('language')) {
+            $this->addSql("ALTER TABLE kimai2_customers ADD `language` VARCHAR(35) NOT NULL DEFAULT 'en'");
+            $this->addSql('UPDATE kimai2_customers SET `language` = ?', [$language]);
```
kevinpapst:
```suggestion
            if (language !== 'en') {
                $this->addSql('UPDATE kimai2_customers SET `language` = ?', [$language]);
            }
```

migrations/Version20260802183206.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        // Use the language of the user with the lowest user_id (mostly the admin)
+        $language = $this->connection->fetchOne("SELECT `value` FROM kimai2_configuration WHERE `name` = 'defaults.user.language'");
+        if (empty($language)) {
+            $language = 'en';
+        }
+
+        if (!$schema->getTable('kimai2_customers')->hasColumn('language')) {
+            $this->addSql("ALTER TABLE kimai2_customers ADD `language` VARCHAR(35) NOT NULL DEFAULT 'en'");
+            if (language !== 'en') {
```
kevinpapst:
```suggestion
            if ($language !== 'en') {
```

src/Entity/Customer.php
```diff
@@ -113,6 +114,12 @@ class Customer implements EntityWithMetaFields, EntityWithBudget, CreatedAt
     #[Serializer\Groups(['Default'])]
     #[Exporter\Expose(label: 'country')]
     private ?string $country = null;
+    #[ORM\Column(name: 'language', type: Types::STRING, length: 35, nullable: false)]
+    #[Assert\NotBlank]
+    #[Serializer\Expose]
+    #[Serializer\Groups(['Default'])]
+    #[Exporter\Expose(label: 'language')]
+    private ?string $language = 'en';
```
kevinpapst, line 123:
```suggestion
    #[ORM\Column(name: 'language', type: Types::STRING, length: 35, nullable: false)]
    #[Assert\Length(min: 2, max: 35)]
    #[Assert\NotBlank]
    #[Serializer\Expose]
    #[Serializer\Groups(['Default'])]
    #[Exporter\Expose(label: 'language')]
    private ?string $language = 'en';
```

templates/customer/index.html.twig
```diff
@@ -21,6 +21,8 @@
         {{ entry.postcode }} {{ entry.city }}
     {% elseif column == 'country' %}
         {{ entry.country|country_name }}
+    {% elseif column == 'language' %}
+        {% if entry.language is not empty %}{{ entry.language|locale_name }}{% endif %}
```
kevinpapst: Language cannot be empty / null, it is none nullable and uses `en` as fallback
```suggestion
        {{ entry.language|locale_name }}
```

kevinpapst (dismissed):

kevinpapst (approved):

kevinpapst:
Thanks @simonschaufi 👍

PR #5862, opened 2026-03-07
[BUGFIX] Fix flaky datetime tests

simonschaufi:
## Description
The tests testProjectNumber in tests/Project/ProjectServiceTest.php and testCustomerNumber in tests/Customer/CustomerServiceTest.php were flaky because they used a PHPUnit data provider (getTestData) that calculated "Expected" date values (like current day or year) at the start of the test suite execution.

The Service methods (calculateNextProjectNumber and calculateNextCustomerNumber) calculated the "Actual" values at runtime when the test method executed.

This caused two issues:

1. Timezone mismatch: If the data provider ran in a context with a different default timezone than the test execution (or if date_default_timezone_get() changed between them), the calculated days could
differ (e.g., "06" vs "05").

2. Midnight boundary: If the test suite started before midnight and the test ran after midnight, the dates would differ.

The tests are fixed by refactoring getTestData to return a callable (closure) for any date-dependent expected value. The test methods now execute this closure to calculate the expected value at the same time the service calculates the actual value, ensuring they match.

## Types of changes

## Checklist

### DIFF
--- a/tests/Customer/CustomerServiceTest.php
+++ b/tests/Customer/CustomerServiceTest.php
@@ -143,7 +143,7 @@ class CustomerServiceTest extends TestCase
     }
 
     #[DataProvider('getTestData')]
-    public function testCustomerNumber(string $format, int|string $expected): void
+    public function testCustomerNumber(string $format, int|string|callable $expected): void
     {
         $configuration = SystemConfigurationFactory::createStub([
             'defaults' => [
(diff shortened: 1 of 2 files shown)


kevinpapst:
Have you been able to reproduce the issue before locally?

simonschaufi:
I tried it several times and the tests always worked. It's only a timing issue and this is a try to see if it would fix it. Maybe also only when tests are run over midnight? Idk. Give it a try...

simonschaufi:
@kevinpapst any specific reason why this is not merged yet? The tests are green in comparison to other pull requests that came in lately.

kevinpapst:
The tests are green in more than 90% of all times, and I strongly doubt that this change helps. 
My best guess is that the service should use a Clock implementation and the tests a ClockMock.
https://symfony.com/doc/current/components/clock.html

simonschaufi:
Well, you didn't even give it a try. You could at least have merged it and see if the tests would fail less afterwards.

kevinpapst:
This is insanity.

PR #5865, opened 2026-03-09
Add support for OIDC authentication

### LINKED ISSUE #2469: Support OpenID Connect
I'd like to install kimai2 and integrate it in an existing [OpenID Connect](https://en.wikipedia.org/wiki/OpenID_Connect) SIngle-Sign-On environment, and it would be great if Kimai2 would add support for it.

matt-oakes:
## Description
This PR adds support for [OpenID Connect (OIDC)](https://en.wikipedia.org/wiki/OpenID_Connect) authentication. It resolves #2469.

The code is heavily based on the existing SAML authentication implementation using:

- `OidcAuthenticator` to hook into Symfony authenticating using `Passport`.
- `OidcProvider` to create and hydrate users.
- `OidcController` to supply the login routes to redirect to the configured OIDC provider.
The implementation of OIDC in this PR is handled by [`jumbojett/openid-connect-php`](https://github.com/jumbojett/OpenID-Connect-PHP) which does all of the heavy lifting.

Configuration is very similar to SAML too, but much simplier because ODIC is more standardized and has sensible defaults. Basic configuration would be:

```yaml
kimai:
    oidc:
        activate: true
        title: "Login with Custom OIDC"
        provider_url: "https://auth.example.com/"
        client_id: "EXAMPLE"
        client_secret: "EXAMPLE"
```

These are the only configuration values which need to be provided because OIDC is able to automatically detect the others using the "well known" URL. In the example above, it is able to access `https://auth.example.com/.well-known/openid-configuration`. The response to this will include all of the URLs and other configuration needed to setup OIDC with that specific server.

I have also gone a little further and implemented the same "role mapping" system which SAML has. You can configure this in a very similar way by adding this to the above config:

```yaml
          roles:
            resetOnLogin: true
            mapping:
                # Insert your role-mapping here (ROLE_USER is added automatically)
                - { oidc: Admin, kimai: ROLE_SUPER_ADMIN }
                - { oidc: Manager, kimai: ROLE_ADMIN }
                - { oidc: Teamlead, kimai: ROLE_TEAMLEAD }
```

This works example the same as with SAML, however, the `attribute` configuration value isn't needed as it is standarized in OIDC.

## Demo video

https://github.com/user-attachments/assets/4e48d1f4-b3f3-4ed1-b172-990e2654e04c

## How to test this

To test this, you will need to have an OIDC authentication provider. There are many available such as [Pocket ID](https://pocket-id.org), [Ory Hydra](https://www.ory.com/hydra), [Authentik](https://goauthentik.io/), [VoidAuth](https://voidauth.app/#/), [Keycloak](https://www.keycloak.org/), and [Forgejo](https://forgejo.org/docs/next/user/oauth2-provider/).

However, if you don't have any setup, then by far the easiest is to start a "demo instance" of Pocket ID (they last for 30 minutes before being deleted).

https://demo.pocket-id.org

When you have an instance you can follow these steps to create the required enviornment for testing:

1. Follow the setup instructions to create a new user.
2. Create a passkey when requested, or skip in this case and it will keep you logged in until the demo instance is deleted.
3. Go to "Administration" -> "OIDC Clients" and then click "Add OIDC client"
4. Fill in the name as "Kamai" (or whatever you like) and leave the rest of the settings blank. Press "Save"
5. Scroll down to the "Allowed User Groups", expand it and press "Unrestrict". This will allow all users in the auth provider (just you) to access the client.
6. At the top of the screen there will be the settings. Expand it and then use "client ID", "client secret", and "issuer URL" to fill in this Kamai config:
```yaml
kimai:
    oidc:
        activate: true
        title: "Login with Pocket ID"
        provider_url: "https://EXAMPLE.demo.podkcet-id.org"
        client_id: "EXAMPLE"
        client_secret: "EXAMPLE"
```
7. Open the login page of your Kimai instance and there should be a "Login with Pocket ID" button. Pressing that should take you to the Pocket ID login page.
8. Press sign in and then accept the information which will be shared.
9. You will be redirected back to Kimai and logged in. If this was a new user your name, email, and profile picture will be set from Pocket ID.

To test role mapping:

1. In your Pocket ID demo instance, go to "Administration" -> "User groups" and then click "Add Group".
2. Give it a friendly name and optionally change the name which will be sent to Kimai. This example will use `admins`.
3. On the next page, scroll down to the user list, tick your user and then press "Save". This assigns your user to that group.
4. Modify your Kimai config to include:
```yaml
          roles:
            resetOnLogin: true
            mapping:
                - { oidc: admins, kimai: ROLE_ADMIN }
```
5. Sign out of Kimai and then sign in again with the "Login with Pocket ID" button. You will be asked for permissions again and this time it will be requesting "groups" access.
6. When you accept and sign in your user will be assigned to the admin role.

## Types of changes

## Checklist
- [x] I updated the documentation (opened here https://github.com/kimai/www.kimai.org/pull/672)


kevinpapst:
Thanks for your effort on this PR. I am not going to details yet, becuase I'd like to discuss one topic first:

Did you try to use/integrate the Symfony OidcHandler https://symfony.com/doc/current/security/access_token.html#using-openid-connect-oidc ?

I am not really open to integrate new dependencies, especially for security related topics. Even more so, as that library didn't receive [any commit in 10 months](https://github.com/jumbojett/OpenID-Connect-PHP/commits/master/) and has [open topics for PHP 8.5](https://github.com/jumbojett/OpenID-Connect-PHP/pull/500). The last release was in September 2024. That is not really a reliable dependency.

simonschaufi:
Personally I would expect Kimai to only provide the interfaces and possibilities to add new providers but the providers should come from plugins.

matt-oakes:
Completely agree on the dependency point. I wasn’t aware that Symfony had OIDC support built-in.

A plugin could make sense, but it has it be tied pretty deeply into the Symfony auth system to work.

simonschaufi:
What I mean is that in my opinion this login authentication provider should not be part of Kimai Core but be provided as a plugin (if possible).

kevinpapst:
The problem with everything auth related is that it needs changes in core config files. So shipping it as plugin would be ideal, but I am not sure how easy that actually is to achieve.

Yes, I'd like to include the option for OIDC, but as said, with Symfony dependencies only.
I used a library for SAML in the past and had massive issues with that, so moved away from it. 

@matt-oakes are you willing to have a look at the Symfony own provider?

simonschaufi:
> I am not sure how easy that actually is to achieve.

I can only tell you what TYPO3 uses. TYPO3 has a Auth service chain (https://docs.typo3.org/permalink/t3coreapi:authentication-service-chain). It basically looks like this: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Services/UsingServices/ServiceChain.html#services-using-services-service-chain

So you have a registry where you can register your own auth service provider and TYPO3 will iterate over them and check if the auth service provider can handle the request and if the auth was successful or not.

For Kimai this could be an event listener where you can register your own service provider and Kimai then calls them. Each auth service provider needs to implement an interface (https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Services/Developer/ServiceApi.html)

matt-oakes:
Sorry for the long delay on this!

I have made the changes now to get rid of the third-party dependency and instead using the OIDC functions from Symfony. The only one we actually need is the `OidcUserInfoTokenHandler` which validates the token and retreives the user identifier. This is the security critical part of the OIDC spec, so that is handled by Symfony.

I had to re-implement the "discovery" part of the OIDC spec which uses the base URL of the server and the "well known" URL to discover all of the other paths it needs to autheticate with that particular server. This is fairly straightforward and is implemented in `OidcDiscovery`. It caches this with an expiry of 1 hour. This can be removed if you would prefer to fetch this every time instead, but I think it's useful as most servers won't change this very often, if at all.

All of the other changes are around generating the URLs to redirect to ourselves and the scafolding required to wire up the new classes.

Hopefully this all makes sence and you can review this fully. All of the same configuration and testing steps for the initial post will still work with this PR with these changes.

matt-oakes:
Just a gentle nudge on this. Is there anything you need from me or any other details you'd like?

kevinpapst:
This is quite a lot of security related code that you want me to maintain in the long term. 
This needs time that I don't have right now. 
Also I'd like to investigate the SecurityChain that @simonschaufi mentioned.

matt-oakes:
Thanks. That's fair and completely understand about the lack of time.

Just let me know if you have any questions or need me to do anything to help with this.

Issue #5751, opened 2026-01-01
Feature Request: API endpoint to initialize/enable Work Contract for a user


glazperle:
## Problem

Currently, there is no API endpoint to initialize or enable the Work Contract (WorkContractBundle) for a user. The `PATCH /api/users/{id}/preferences` endpoint returns **404 Not Found** if the Work Contract has not been previously configured in the Kimai UI.

The API documentation states:
> "You cannot create unknown preferences: if the given name is not configured, an exception will be raised."

This means users building integrations or MCP servers cannot programmatically set up work contracts for users - they must first manually configure each user through the web UI.

## Current Behavior

```bash
# User with Work Contract configured: ✅ Works
curl -X PATCH /api/users/1/preferences -d '[{"name": "holidays", "value": "25"}]'
# Returns: 200 OK

# User without Work Contract configured: ❌ Fails  
curl -X PATCH /api/users/28/preferences -d '[{"name": "holidays", "value": "25"}]'
# Returns: 404 Not Found
```

## Requested Feature

Add an API endpoint to initialize/enable the Work Contract for a user, for example:

```
POST /api/work-contract/{user}/initialize
```

or allow the `PATCH /api/users/{id}/preferences` endpoint to create preferences that don't exist yet (with sensible defaults).

## Use Case

When building integrations (like MCP servers for AI assistants), we need to programmatically:
1. Set up new employees with their work contract details
2. Update vacation days, working hours, contract dates via API
3. Automate onboarding workflows

Currently, step 1 requires manual UI interaction, breaking the automation flow.

## Environment

- Kimai Version: 2.45.0
- WorkContractBundle: v1.29.0

kevinpapst:
Can you re-test with the latest changes from `main` @glazperle ?

@viggou added some code to initialize the prefs for both existing and new users

glazperle:
Thanks @kevinpapst and @viggou! 🙏

I went through #5894. Registering the work-contract preferences with sensible defaults via `WorkContractPreferenceSubscriber`, so that `PATCH /api/users/{id}/preferences` no longer 404s for users who never opened the contract page in the UI, is exactly what was needed. That removes the manual "configure once in the UI first" step from automated onboarding.

Caveat on my side: I haven't been able to verify against a live server yet, because the fix isn't in a tagged release yet; my instances are on 2.60.0. I'll confirm end-to-end once it ships in a release (2.61.0?). From reading the diff it looks correct.

[...]

PR #5894, opened 2026-04-10
Add working contract preferences to user when setting for the first time from API

viggou:
## Description
This pull request allows the public API to add preferences to users if they haven't been set before in the web ui. Fixes this issue #5751 

## Types of changes

## Checklist


kevinpapst:
Thanks for trying to solve the issue. 
I'd like to see a more generic approach. Maybe via Event, not sure yet.

viggou:
I've created a new EventSubscriber which adds the work contract preferences. I'm not sure if you want them in UserPreferenceSubscriber instead? What do you think?

kevinpapst:
Make it even more generic: something like a `RegisterDefaultUserPreferences` which only receives a `User`.
Then add a listener that adds the WorkContract related values to the given User.

viggou:
Alright, I added a RegisterUserPreferencesEvent, which only takes user and I've kept the WorkContractPreferenceSubscriber as the listener. The new event is dispatched by UserPreferenceSubscriber.

kevinpapst (dismissed):
I changed the logic to re-use the existing events `CreateUserEvent` and `UserPreferenceEvent` and added some tests.

Thanks @viggou 👍

kevinpapst (approved):

kevinpapst (approved):

PR #5916, opened 2026-04-21
new API endpoint to save invoice meta-fields

kevinpapst:
## Description

- Adds a new invoice voter to improve permission checks
- Adds a new API endpoint to set meta-fields for invoices
- User invoice service instead of repository
- New global voter methods to check team access
- Trigger more Invoice events on updates

## Types of changes

## Checklist


src/Command/InvoiceCreateCommand.php
```diff
@@ -41,7 +41,7 @@ final class InvoiceCreateCommand extends Command
     private bool $previewUniqueFile = false;
 
     public function __construct(
-        private readonly ServiceInvoice $serviceInvoice,
+        private readonly InvoiceService $InvoiceService,
```
simonschaufi, line 44: local variables should be lowercase. this is against PSR standards.

PR #5930, opened 2026-04-28
New API endpoint to create user API tokens

dtorras:
## Description
Adds a new API endpoint `POST /api/users/{id}/api-token` that creates a new `AccessToken` for the given user and returns the plain token in the response. This is the missing counterpart of the existing `DELETE /api/users/api-token/{id}`.

Currently the only way to obtain an `AccessToken` is via the web UI. This blocks any automated provisioning flow that needs to create a Kimai user and immediately get a usable Bearer token for that user — for example, a SaaS integration that creates one Kimai user per customer. The legacy `plainApiToken` field on `POST /api/users` only populates the deprecated `users.api_token` column used by the `X-AUTH-USER`/`X-AUTH-TOKEN` headers, which is not interchangeable with `Authorization: Bearer …`.

To guarantee that tokens generated via the API and via the web UI always follow the same algorithm, the random-token generation has been extracted into a new factory method `AccessToken::createForUser(User $user): self`. Both `ProfileController::createAccessToken` and the new `UserController::postApiToken` now use this single source of truth. The `AccessToken` constructor signature `(User $user, string $token)` is left unchanged so that fixtures and `ResetTestCommand`, which need predictable tokens, keep working as-is.

## Types of changes

## Checklist
- [x] I updated the documentation (OpenAPI annotations are included in the controller; happy to open a follow-up PR on www.kimai.org if maintainers want a manual mention)

### DIFF
--- a/src/API/UserController.php
+++ b/src/API/UserController.php
@@ -202,6 +202,69 @@ final class UserController extends BaseApiController
     }
 
     /**
+     * Create API token
+     *
+     * Creates a new API token for the given user. The plain token is only visible in this response.
+     */
+    #[OA\Post(description: 'Creates a new API token for the given user and returns it. The plain token is only visible in this response.')]
+    #[OA\RequestBody(required: true, content: new OA\JsonContent(
+        required: ['name'],
+        properties: [
+            new OA\Property(property: 'name', type: 'string', description: 'A name to identify this token', example: 'My API token'),
+            new OA\Property(property: 'expiresAt', type: 'string', format: 'date-time', description: 'Optional expiration date', example: '2027-01-01T00:00:00+00:00'),
+        ]
+    ))]
+    #[OA\Response(response: 200, description: 'Returns the created API token including the plain text token value.', content: new OA\JsonContent(
+        properties: [
+            new OA\Property(property: 'id', type: 'integer'),
+            new OA\Property(property: 'name', type: 'string'),
+            new OA\Property(property: 'token', type: 'string', description: 'The plain text token — store it now, it cannot be retrieved later'),
+            new OA\Property(property: 'expiresAt', type: 'string', format: 'date-time', nullable: true),
+        ]
+    ))]
+    #[OA\Parameter(name: 'id', in: 'path', description: 'User ID to create the token for', required: true)]
+    #[Route(methods: ['POST'], path: '/{id}/api-token', name: 'post_api_token', requirements: ['id' => '\d+'])]
+    public function postApiToken(User $profile, Request $request, AccessTokenRepository $accessTokenRepository): Response
+    {
+        $user = $this->getUser();
+        if (!$this->isGranted('api-token', $user)) {
+            throw $this->createAccessDeniedException('User has no access to API tokens');
+        }
+
+        if (!$this->isGranted('api-token', $profile)) {
+            throw $this->createAccessDeniedException('You are not allowed to create API tokens for this user');
+        }
+
+        $name = $request->request->get('name');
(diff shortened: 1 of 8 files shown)


src/Entity/AccessToken.php
```diff
@@ -49,6 +49,11 @@ public function __construct(User $user, string $token)
         $this->token = $token;
     }
 
+    public static function createForUser(User $user): self
+    {
+        return new self($user, substr(bin2hex(random_bytes(100)), 0, 25));
+    }
```
kevinpapst: remove and make the $token parameter in __construct optional instead.
Apply the logic to the construct if no $token was given.

src/API/UserController.php
```diff
@@ -201,6 +201,69 @@ public function patchAction(Request $request, User $profile): Response
...
+        properties: [
+            new OA\Property(property: 'id', type: 'integer'),
+            new OA\Property(property: 'name', type: 'string'),
+            new OA\Property(property: 'token', type: 'string', description: 'The plain text token — store it now, it cannot be retrieved later'),
+            new OA\Property(property: 'expiresAt', type: 'string', format: 'date-time', nullable: true),
+        ]
+    ))]
+    #[OA\Parameter(name: 'id', in: 'path', description: 'User ID to create the token for', required: true)]
+    #[Route(methods: ['POST'], path: '/{id}/api-token', name: 'post_api_token', requirements: ['id' => '\d+'])]
+    public function postApiToken(User $profile, Request $request, AccessTokenRepository $accessTokenRepository): Response
```
kevinpapst, line 213: add the correct #[IsGranted...] attribute

src/Controller/ProfileController.php
```diff
@@ -181,7 +181,7 @@ public function createAccessToken(
         AccessTokenRepository $accessTokenRepository
     ): Response
     {
-        $accessToken = new AccessToken($profile, substr(bin2hex(random_bytes(100)), 0, 25));
+        $accessToken = AccessToken::createForUser($profile);
```
kevinpapst: revert, no static please

kevinpapst (changes requested):
please address these issues and then we can discuss how to move forward

dtorras:
@kevinpapst Done. Added `#[IsGranted('api-token', 'profile')]` matching the pattern in `ProfileController::createAccessToken`.

kevinpapst (changes requested):
The entire method is LLM generated and does not follow the overall setup of the Kimai API definitions.
Prompting is just 10% of the game. 
Are you going to make it compatible with the rest of the application?

dtorras:
Yes, I want the contribution to be functional. That's why I'd been applying your earlier feedback as it came in.

This is my first PR to Kimai and I'm not familiar with the codebase. I'd been looking to use the API with your Cloud product, but saw that API passwords will be removed without a programmatic alternative for creating access tokens. So I opened a small PR to add the missing endpoint. I don't quite understand the tone of your last message, and I don't think it was necessary.

The technical point is fair though, and I understand it. I've gone through the branch again.

kevinpapst:
>  I don't quite understand the tone of your last message, and I don't think it was necessary.

Please don't read what was never said. 
The method you want to add is very dangerous.

If you want to discuss a PR first, then please:
- mention it in the description
- create it as draft

If you want me to maintain your code long term, I think its fair to ask you to put some effort into it upfront.
I do use LLMs as well and I believe that I can see whether code was prompted or also manually adjusted.
Now, after your last commit, it looks much better and now I am willing to give it an honest review. Thanks 👍

PR #5932, opened 2026-04-30
Run console (and thus php) as the correct user during entrypoint.sh

DavyLandman:
## Description
This makes the kimai container compatible with rootless setups such as podman & docker rootless. Without this change the container would just keep getting permission errors during rebuilding of the cache etc.

Note that running the container in rootless podman/docker it's still a bit messy, as you have to start the user as root, but then assign the host user to 33, but at least it works.

For example this is my podman quadlet.

```systemd
Environment=USER_ID=33
Environment=GROUP_ID=33
UserNS=keep-id:uid=33,gid=33,size=2000
User=0
```

## Types of changes

## Checklist

### DIFF
--- a/.docker/entrypoint.sh
+++ b/.docker/entrypoint.sh
@@ -55,8 +55,9 @@ function handleStartup() {
     pwconv
   fi
 
+  PHP_RUN_USER=$(id -nu "$USER_ID")
   if [ -e /use_apache ]; then
-    export APACHE_RUN_USER=$(id -nu "$USER_ID")
+    export APACHE_RUN_USER=$PHP_RUN_USER
     # This doesn't _exactly_ run as the specified GID, it runs as the GID of the specified user but WTF
     export APACHE_RUN_GROUP=$(id -ng "$USER_ID")
     export APACHE_PID_FILE=/var/run/apache2/apache2.pid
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks! Unfortunately I have no experience with that topic. But if you say

> Note that running the container in rootless podman/docker it's still a bit messy, as you have to start the user as root, but then assign the host user to 33, but at least it works.

it makes it less likely to be merged honestly. Maybe someone else can join the discussion and share some advice?

DavyLandman:
> it makes it less likely to be merged honestly. Maybe someone else can join the discussion and share some advice?

In general the problem is with the whole runnig commands as root (in the container), and then switching to a different user for the execution of the job.

Before this PR: there is no way to run the container in rootless mode such as podman. After this PR, it's possible, but the container could be changed more to make it more fluid, but that would require more changes.

My goal was to do the minimal amount of changes to make it at least workable.

kevinpapst:
> My goal was to do the minimal amount of changes to make it at least workable.

Which is great start, thanks for sharing the results!

Maybe someone else will show up and pick up your work and improve on that one.
Its about time, that the entire Docker setup will be improved, so this should be one of the tasks I guess.

PR #5937, opened 2026-05-03
Adds a global setting for the default homepage

tofuSCHNITZEL:
## Description
Currently the default homepage of a (new) user is defined in a constant in a php file - so no good way to set a different default for new users (or user that not have set a default startpage.
This PR adds a global setting for the users default homepage but keeps the default like it was defined in the constant as "timesheet"

## Types of changes

## Checklist
_documentation is not found there... 


src/Controller/HomepageController.php
```diff
@@ -64,8 +63,8 @@ public function homepage(Request $request, LocaleService $service, EventDispatch
             $routes[] = [$userRoute, User::DEFAULT_LANGUAGE];
         }
 
-        $routes[] = [self::DEFAULT_ROUTE, $userLanguage];
-        $routes[] = [self::DEFAULT_ROUTE, $requestLanguage];
+        $routes[] = [$systemConfiguration->getUserDefaultHomepage(), $userLanguage];
```
kevinpapst: fetch this once and re-use the local wariable

kevinpapst (changes requested):
You need to register the configuration also here: https://github.com/kimai/kimai/blob/main/src/DependencyInjection/Configuration.php#L626
and adapt the related test

src/Controller/HomepageController.php
tofuSCHNITZEL: yes, agreed.

tofuSCHNITZEL:
> You need to register the configuration also here: https://github.com/kimai/kimai/blob/main/src/DependencyInjection/Configuration.php#L626 and adapt the related test

done.

tofuSCHNITZEL:
> boost

no idea what to do.... sorry....

kevinpapst (changes requested):
Please check the CoPilot feedback.

The last fallback in `HomepageController` in line 86 should use the static route `'timesheet'` instead of `$defaultRoute`.

In that line we can be sure that `$defaultRoute` does not exist.

tofuSCHNITZEL:
> Please check the CoPilot feedback.
> 
> The last fallback in `HomepageController` in line 86 should use the static route `'timesheet'` instead of `$defaultRoute`.
> 
> In that line we can be sure that `$defaultRoute` does not exist.

okay understood - but we can't guarantee that "timesheet" will always be a valid route, right?

for the other suggestions - I don't want to commit AI code that I don't understand - so I'm hesitant to just accept the changes that copilot is presenting for the other two suggestions...

kevinpapst:
i think we can safely assume that kimai will always have a route called `timesheet` 😁 it worked for years.
this piece of the code should hopefully never be triggered. but its better to have that route in there, you know: better safe than sorry.

tofuSCHNITZEL:
> i think we can safely assume that kimai will always have a route called `timesheet` 😁 it worked for years. this piece of the code should hopefully never be triggered. but its better to have that route in there, you know: better safe than sorry.

yea, got it - I'm sorry for not being helpful on the other two suggestions - I let copilot generate two suggestions both were different and I could not understand the reasoning behind the changes so I don't feel comfortable adding the changes to my branch...

kevinpapst:
yeah, am i also often hesitant to use code suggestions. but its good for reviews and finding possible issues.

PR #5938, opened 2026-05-03
Adds a setting to disable first time wizard for new users

tofuSCHNITZEL:
## Description

This PR adds a global setting that disables the settings wizard that is displayed on the first login for regular users.

## Types of changes

## Checklist


kevinpapst:
I don't like that idea. 
What is the specific use-case?
Why should a user not decide for himself how the Kimai UI should behave?

tofuSCHNITZEL:
in this usecase 100+ Users are using the kiosk plugin to the check in and out - they never needed to use the webui - until now where they should be given access to only see their recorded times... so just telling them to request a new PW to login is one thing - but with the wizard we need to send them pictures and step by step instructions on how to ignore and where to click on the wizard - its a complete and unnecessary hassle. they dont need to decide the settings - they are all in the same timezone and same language and they don't understand any of the settings - and they don't need to - since they will only be looking at time and not to anything else.

so there is no harm in having this option since per default its not changing the behaviour of kimai but gives additional control to the admins to shape kimai to their needs.

kevinpapst:
>  they should be given access to only see their recorded time

So the new Kiosk view where you can see your entire overview is not enough?
image

What is missing?


I do get the argument. 
But if a Wizard is already to complicated, maybe its better to improve the Kiosk instead of finding workarounds.

tofuSCHNITZEL:
thanks for the info about the view - actually this is not all they need to do - they do need to be able interact with the normal kimai menu because they need to navigate to a special plugin page - so the kiosk view is not what we need.

translations/system-configuration.en.xlf
```diff
@@ -14,6 +14,10 @@
         <source>theme.avatar_url</source>
         <target>Allow the use of URLs for avatar images</target>
       </trans-unit>
+      <trans-unit id="user_auth_wizard" resname="user_auth_wizard">
+        <source>user_auth_wizard</source>
+        <target>First-time wizard for new users</target>
```
kevinpapst:
```suggestion
        <target>Show setup wizard for new users</target>
```

src/Configuration/SystemConfiguration.php
```diff
@@ -365,6 +365,17 @@ public function getUserDefaultCurrency(): string
         return $this->getDefaultCurrency();
     }
 
+    public function isWizardActive(): bool
```
kevinpapst:
```suggestion
    public function isUserWizardActive(): bool
```

src/Configuration/SystemConfiguration.php
```diff
@@ -365,6 +365,17 @@ public function getUserDefaultCurrency(): string
...
+    {
+        $value = $this->find('user.wizard');
+
+        if ($value === null) {
+            return true;
+        }
+
+        return (bool) $value;
```
kevinpapst:
```suggestion
        return (bool) $this->find('user.wizard');
```

translations/system-configuration.en.xlf
```diff
@@ -14,6 +14,10 @@
         <source>theme.avatar_url</source>
         <target>Allow the use of URLs for avatar images</target>
       </trans-unit>
+      <trans-unit id="user_auth_wizard" resname="user_auth_wizard">
```
kevinpapst: execute `bin/console kimai:translation --resname` to generate the correct `id`

kevinpapst (changes requested):
1. Please add the  german translation `Einrichtungsassistent für neue Benutzer anzeigen`
2. Add the configuration to https://github.com/kimai/kimai/blob/main/src/DependencyInjection/Configuration.php#L547
3. Adjust the test (maybe another one will fail as well): https://github.com/kimai/kimai/blob/main/tests/DependencyInjection/ConfigurationTest.php#L297

translations/system-configuration.en.xlf
tofuSCHNITZEL: thanks for pointing that out!

tofuSCHNITZEL:
> 1. Please add the  german translation `Einrichtungsassistent für neue Benutzer anzeigen`
> 
>     2. Add the configuration to https://github.com/kimai/kimai/blob/main/src/DependencyInjection/Configuration.php#L547
> 
>     3. Adjust the test (maybe another one will fail as well): https://github.com/kimai/kimai/blob/main/tests/DependencyInjection/ConfigurationTest.php#L297

done. not sure what you mean by "maybe another one will fail as well"

kevinpapst:
I meant: maybe there are more tests that need to be adapted.

Can you please add tests to cover the new logic, especially the WizardController.

tofuSCHNITZEL:
> I meant: maybe there are more tests that need to be adapted.
> 
> Can you please add tests to cover the new logic, especially the WizardController.

done, hope its what you had in mind.

src/EventSubscriber/WizardSubscriber.php
```diff
@@ -63,6 +65,10 @@ public function onKernelRequest(RequestEvent $event): void
             return;
         }
 
+        if (!$this->systemConfiguration->isUserWizardActive()) {
```
kevinpapst: Let's limit this to normal users only:
```suggestion
        if ($user->isRegularUserOnly() && !$this->systemConfiguration->isUserWizardActive()) {
```

tests/Configuration/SystemConfigurationTest.php
```diff
@@ -125,6 +128,7 @@ protected function getDefaultLoaderSettings(): array
             (new Configuration())->setName('timesheet.markdown_content')->setValue('1'),
             (new Configuration())->setName('timesheet.default_begin')->setValue('07:00'),
             (new Configuration())->setName('timesheet.active_entries.hard_limit')->setValue('7'),
+            (new Configuration())->setName('user.wizard')->setValue('0'),
```
kevinpapst:
```suggestion
            (new Configuration())->setName('user.wizard')->setValue(false),
```

tests/Controller/WizardControllerTest.php
```diff
@@ -37,6 +39,97 @@ public function testProfileWizard(): void
         $this->assertAccessIsGranted($client, '/wizard/profile');
     }
 
+    public function testWizardDoesNotAppearOnFirstLoginIfDisabled(): void
+    {
+        $client = $this->getClientForAuthenticatedUser(User::ROLE_USER);
+
+        $this->setSystemConfiguration('user.wizard', '0');
```
kevinpapst: Using a numeric string, to auto-cast to a bool is evil 😁 
```suggestion
        $this->setSystemConfiguration('user.wizard', false);
```

src/Controller/SystemConfigurationController.php
```diff
@@ -546,6 +546,11 @@ private function getConfigurationTypes(): array
                         ->setLabel('theme.avatar_url')
                         ->setType(YesNoType::class)
                         ->setTranslationDomain('system-configuration'),
+                    (new Configuration('user.wizard'))
+                        ->setLabel('user_auth_wizard')
+                        ->setType(YesNoType::class)
+                        ->setValue(true)
```
kevinpapst: Remove this value
```suggestion
```

kevinpapst (changes requested):
One more round of feedback

src/EventSubscriber/WizardSubscriber.php
kevinpapst:
```suggestion
        if ($user->requiresPasswordReset()) {
            $response = new RedirectResponse($this->urlGenerator->generate('wizard', ['wizard' => 'password']));
            $event->setResponse($response);
        }

        if ($user->isRegularUserOnly() && !$this->systemConfiguration->isUserWizardActive()) {
            return;
        }

        foreach (User::WIZARDS as $wizard) {
            if (!$user->hasSeenWizard($wizard)) {
                $response = new RedirectResponse($this->urlGenerator->generate('wizard', ['wizard' => $wizard]));
                $event->setResponse($response);

                return;
            }
        }
```

kevinpapst (approved):

kevinpapst (approved):

kevinpapst:
thanks @tofuSCHNITZEL for the contribution 👍  

there was a flaw that the password reset was skipped, i fixed that one. and I will add a few tests as well for the next release.

tofuSCHNITZEL:
Thanks!

tofuSCHNITZEL:
> there was a flaw that the password reset was skipped, i fixed that one. 

Thanks for the catch!

PR #5943, opened 2026-05-11
fix checking for correct formatter in durationDecimal

cheriimoya:
## Description

`LocaleFormatter::durationDecimal()` guarded the lazy initialization of `$this->decimalFormatter` with the wrong property (`$this->numberFormatter` instead of `$this->decimalFormatter`). If `amount()` was called before `durationDecimal()` on the same formatter instance, `$this->numberFormatter` would already be set, the guard would pass, `$this->decimalFormatter` would never be initialized, and the subsequent `->format()` call would fail on null.

## Types of changes

## Checklist

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3646,11 +3646,6 @@ parameters:
             path: src/Utils/JavascriptFormatConverter.php
 
         -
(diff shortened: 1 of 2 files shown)


kevinpapst (approved):
Good catch, thanks @cheriimoya 👍

PR #5944, opened 2026-05-11
fix formatting locale reset after embedded controller sub-requests

cheriimoya:
## Description

Pages that contain embedded controllers rendered via `render(controller(...))` (e.g. the project details page with its activities sub-request) lose the user's formatting locale mid-render. Symfony's `LocaleAwareListener` and `LocaleSwitcher` reset all `LocaleAwareInterface` services, including `LocaleFormatExtensions`, to the URL locale during the sub-request, and `UserEnvironmentSubscriber` already skipped sub-requests without restoring afterwards. Everything rendered after the sub-request in the same template (dates, durations, money amounts) used the wrong locale for the rest of the page.

Fixed by saving the user's formatting locale in `prepareEnvironment()` and subscribing to `KernelEvents::FINISH_REQUEST` at priority `-20` (after `LocaleAwareListener` at `-15`) to restore both `\Locale::setDefault()` and `LocaleFormatExtensions::setLocale()` to the saved value after each sub-request.

This affects any user whose formatting locale differs from their URL locale, i.e., regional variants like de_AT, fr_CH, en_GB, etc.

## Types of changes

## Checklist

### DIFF
--- a/src/EventSubscriber/UserEnvironmentSubscriber.php
+++ b/src/EventSubscriber/UserEnvironmentSubscriber.php
@@ -12,6 +12,7 @@ namespace App\EventSubscriber;
 use App\Entity\User;
 use App\Twig\LocaleFormatExtensions;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FinishRequestEvent;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):
Wow, good catch, thanks 👍

kevinpapst (changes requested):
Please use https://github.com/kimai/kimai/pull/5952 as target instead of main

cheriimoya:
Thanks for looking into this:) I also rebased onto the tag.

kevinpapst (approved):
Thanks. I did some manual tests locally, because I couldn't believe it 😁 
A strange SF behavior that I didn't recognized before.

PR #5951, opened 2026-05-21
Impersonate user

kevinpapst:
## Description

This feature can only ever be activated by Super-Admins and only if the new permission `impersonate_user` is explicitly activated in the roles screen.

This feature is meant to help testing new environments, especially teams, permissions and roles.

**TODO:**
- [ ] do not allow to impersonate other Super-Admin
- [ ] write audit log
- [ ] prevent self-impersonation
- [ ] prevent impersonation of disabled users (fails already, but hide button)
- [ ] use constant instead of hardcoded `impersonate_user`
- [ ] what about 2FA? config/packages/scheb_2fa.yaml misses SwitchUserToken

## Types of changes

## Checklist


kevinpapst:
not for now

PR #5962, opened 2026-06-01
fix adding an empty space in the delete modals

cheriimoya:
## Description
Currently, there is a space missing in the message when trying to delete an activity/user/project:
image

With this patch, we get a space:
image

## Types of changes

## Checklist


kevinpapst (approved):
That's an eagle-eye fix 😁  
Thanks 👍

PR #6033, opened 2026-07-02
Allow searching projects by their customer name in select dropdowns

### LINKED ISSUE #5910: Search By Customer in Project dropdown
### Describe the problem

(coming from version 1.x where it was possible)

Searching by customer name in the project dropdown (e.g. in Weekly Hours) to see all projects related to a customer doesn't works anymore.

Typing the customer name in the project dropdown should show all projects of that customer.

Kirill070:
TomSelect searches only the option text by default, so projects grouped by customer could no longer be found by typing the customer name — a regression from Kimai 1.x. Add a custom score that
  also matches an option's optgroup label (the customer/parent name).

Fixes #5910

  ## Description
  
  Projects in the dropdown are grouped by customer via `<optgroup>`, but TomSelect
  only searches the option **text** by default, so typing a **customer** name
  returned "No results" — a regression from Kimai 1.x.

  Added a custom `score` in `assets/js/forms/KimaiFormSelect.js` that also matches
  an option's optgroup **label** (the customer/parent name), resolved via
  `this.optgroups[item.optgroup]`. The default search is untouched, and this helps
  every grouped select (projects by customer, activities by project).

  Note: the `searchField: ['text', 'optgroup']` approach doesn't work here — on
  the bundled TomSelect (2.5.2) that field holds the optgroup **id**, not its label.

  JS-only change; `pnpm run lint` (eslint) passes for the changed file. Verified
  manually with demo data: customer search now finds projects, project-name search
  is unchanged, and nonsense input still returns no results (no false positives).

  ## Types of changes

  ## Checklist

**Before:**
bug-customer-no-results

**After:**
fixed-customer-search

### DIFF
--- a/assets/js/forms/KimaiFormSelect.js
+++ b/assets/js/forms/KimaiFormSelect.js
@@ -84,6 +84,26 @@ export default class KimaiFormSelect extends KimaiFormTomselectPlugin {
             // TODO make this value configurable with a data attribute
             maxOptions: 500,
             sortField:[{field: '$order'}, {field: '$score'}],
+            // also match options by their optgroup label (e.g. find a project by its
+            // customer name). TomSelect only searches the option text by default, and
+            // an option's "optgroup" property holds the group id, not its label, so we
+            // resolve the label from the optgroups registry inside a custom score.
+            score: function(search) {
+                const scoreFn = this.getScoreFunction(search);
+                const needle = search.trim().toLowerCase();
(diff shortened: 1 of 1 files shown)


Kirill070:
Thanks for the pointer to KimaiFormSelect.js in #5910, that was spot on.

kevinpapst (approved):
Awesome work, thanks 👍

PR #6048, opened 2026-07-13
Add stable id tie-breaker to timesheet pagination order

### LINKED ISSUE #4881: Duplicate timesheets returning when changing page number
### Discussed in https://github.com/kimai/kimai/discussions/4875

<div type='discussions-op-text'>

<sup>Originally posted by **ethan-conrad1** May 24, 2024</sup>
When calling the Kimai API to retrieve timesheet entries, in some scenario's we are getting duplicated results returned. For example:

image (2)

image (3)

As can be seen in the above images, the API request is the same, with the only difference being the page number. However, we are receiving the same id (9760) in both pages.

The issue seems to always be going from page 1 to page 2. If we change from page 2 to page 3 there are no issues.

This is causing uniqueness constraint violations in our application that uses the API.

Please can this be looked at?</div>

TowyTowy:
## Description

Timesheet queries are ordered only by a single, non-unique column (e.g. the default `begin DESC`, or a project/customer/activity name). Records sharing the same value in that column have no deterministic order, so with `LIMIT`/`OFFSET` pagination the database can return the same record on more than one page while skipping others. API consumers observe this as duplicate timesheet ids across consecutive pages (#4881).

Adding the primary key as a secondary sort criterion in `TimesheetRepository::getQueryBuilderForQuery()` makes the order stable across pages. This only breaks ties — it cannot change which rows match. Added an integration test that inserts 20 records with an identical `begin` and asserts pagination returns each exactly once.

Closes #4881

## Types of changes

## Checklist

- [x] I verified that my code applies to the guidelines (`composer code-check`) — php-cs-fixer + PHPStan clean, integration test fails before / passes after (verified against MariaDB 10.5)
- [ ] Documentation update — n/a

This change was prepared with AI assistance (Claude) and reviewed/verified by me; see the `Co-Authored-By` trailer.

### DIFF
--- a/src/Repository/TimesheetRepository.php
+++ b/src/Repository/TimesheetRepository.php
@@ -555,6 +555,10 @@ class TimesheetRepository extends EntityRepository
         }
 
         $qb->addOrderBy($orderBy, $query->getOrder());
+        // add the primary key as a stable tie-breaker, otherwise records sharing
+        // the same value in the sort column have no deterministic order and can be
+        // returned on more than one page (or skipped entirely) when paginating (#4881)
+        $qb->addOrderBy('t.id', $query->getOrder());
 
         $user = [];
         if (null !== $query->getUser()) {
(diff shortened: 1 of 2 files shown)


kevinpapst (approved):
Thanks @TowyTowy 👍

PR #6052, opened 2026-07-14
fix: validate every entry of the color choices list

TowyTowy:
## Description

`ColorChoicesValidator` stopped validating the color list as soon as it reached a valid "bare" color (a hex code with no custom label), because that branch used `return` instead of `continue`. Any invalid color placed after such an entry was silently accepted — e.g. `#fffaaa,fffaaa` passed validation, even though `fffaaa` on its own is correctly rejected (order-dependent: `fffaaa,#fffaaa` is still caught).

This affects the `theme.color_choices` system setting, letting an admin save malformed color codes.

Fix: use `continue` so every comma-separated entry is validated, and skip empty entries (trailing/duplicated commas) so existing valid lists like `Foo|#fffaaa,` keep passing.

Added regression cases to `ColorChoicesValidatorTest` covering an invalid hex and an over-long name placed after a valid bare color (they fail on `main`, pass with this change).

## Types of changes

## Checklist

- [x] I verified that my code applies to the guidelines — php-cs-fixer + PHPStan clean; the validator unit test fails before and passes after (29 tests OK).

Prepared with AI assistance (Claude) and reviewed/verified by me.

### DIFF
--- a/src/Validator/Constraints/ColorChoicesValidator.php
+++ b/src/Validator/Constraints/ColorChoicesValidator.php
@@ -28,6 +28,11 @@ final class ColorChoicesValidator extends ConstraintValidator
         $colors = explode(',', $value);
 
         foreach ($colors as $color) {
+            // skip empty entries, e.g. caused by a trailing or duplicated comma
+            if (trim($color) === '') {
+                continue;
+            }
+
             $color = explode('|', $color);
             $name = $color[0];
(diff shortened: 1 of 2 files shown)


kevinpapst (approved):

kevinpapst:
Thanks @TowyTowy 👍

PR #6054, opened 2026-07-15
fix: use 24-hour format for invoice date_process variables

TowyTowy:
The `*_process` invoice template variables (`entry.date_process`, `invoice.date_process`, `invoice.due_date_process`, `invoice.first_process`, `invoice.last_process`, `query.begin_process`, `query.end_process`) are documented as machine-processable timestamps, but were formatted with the lowercase `h` (12-hour) placeholder and no AM/PM marker. Afternoon times were therefore rendered ambiguously — e.g. an entry beginning at 18:00 produced `... 06:00:00` and 14:00 produced `... 02:00:00`, making the values impossible to sort or parse correctly.

This switches `DATE_PROCESS_FORMAT` to the 24-hour `H` placeholder in both default hydrators and adds regression tests (the model hydrator test previously asserted against the same buggy `h` format, so it never caught this).

Verified locally: the new test fails before the change (`06:00:00` for an 18:00 entry) and passes after; both hydrator test files green.

Prepared with AI assistance (Claude) and reviewed/verified by me.

### DIFF
--- a/src/Invoice/Hydrator/InvoiceItemDefaultHydrator.php
+++ b/src/Invoice/Hydrator/InvoiceItemDefaultHydrator.php
@@ -15,7 +15,7 @@ use App\Invoice\InvoiceModel;
 
 final class InvoiceItemDefaultHydrator implements InvoiceItemHydrator
 {
-    private const DATE_PROCESS_FORMAT = 'Y-m-d h:i:s';
+    private const DATE_PROCESS_FORMAT = 'Y-m-d H:i:s';
 
     private InvoiceModel $model;
 
--- a/src/Invoice/Hydrator/InvoiceModelDefaultHydrator.php
+++ b/src/Invoice/Hydrator/InvoiceModelDefaultHydrator.php
@@ -15,7 +15,7 @@ use Symfony\Component\Intl\Countries;
 
 final class InvoiceModelDefaultHydrator implements InvoiceModelHydrator
 {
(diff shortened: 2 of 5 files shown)


TowyTowy:
Pushed a follow-up: the `Linting` failure was PHPStan's counted ignore baseline in `tests/phpstan.neon` - the new `testDateProcessUsesTwentyFourHourFormat()` adds a second `getCalculator()->getEntries()` call (same pattern as `testHydrate()`), so I bumped that entry's `count` from 1 to 2. PHPStan on `tests/phpstan.neon` is green locally now.

The separate `Frontend verification` red is unrelated to this PR - `pnpm audit` is hitting the retired npm audit endpoint (`410: This endpoint is being retired`), which fails on any branch. Happy to squash the two commits if you prefer a single commit.

(Prepared with AI assistance.)

kevinpapst:
Thanks 👍 

The `pnpm audit` is already fixed in the next release branch. 

Can you address the minor feedback from Copilot please. If something was done wrong in the past, it doesn't mean we should keep on doing it 😁

TowyTowy:
Done - both tests now assert the calculator is non-null and the ignoreErrors entry is removed entirely. `phpstan analyse -c tests/phpstan.neon` is green locally.

kevinpapst:
Thank you @TowyTowy 👍

Issue #2181, opened 2020-12-10
Internal rate not fully functional


HeinzWuert:
**Describe the bug**
To use internal rate, hourly rate must be 0.00.
This works fine.
But if you use alternative hourly wages in the activity or project, these are used instead of the internal rate.
If hourly rate = 0,00 and internal rate= x,xx the internal rate must be used in all time bookings, regardless of alternative hourly wages.

Idea: 
To achieve the effect between hourly rate and internal rate, one of the two values must be 0.00.
Delete one of the two data fields and add a checkbox to switch between these two variants. The default should be without an internal rate.

kevinpapst:
In the database there are two 4 fields for each timesheet record: 
- the hourly rate
- the fixed rate (not important here) 
- the rate (the calculated value of hourly rate * duration)
- the internal rate (the calculated value of internal hourly rate * duration)

The internal hourly rate is always used when it is set. If it is not set, then the "regular" hourly rate is used as fallback to calculate the "internal rate".

You cannot see the calculated "internal rate" in Kimai (by now), only in the database.

I know that this feature is used by customers, so I don't think there is a bug (at least I don't see one, can you help me to reproduce it?).
I'd guess that this is probably a misunderstanding how it works (as it is not really documented).

kevinpapst:
Sounds to me as if it is working as intended. 
I will add some documentation what the internal rate is and how it works.
Then we can double check where it might behave wrong.

HeinzWuert:
1. how can it be intended, when the text shows the correct value but the $bar shows nothing?
2. how can it be intended, when the internal rate is overwritten with fees?
I'm looking forward to the coming documentation how this can be explained.

kevinpapst:
1. Budget targets external costs. Internal rates are not external. 
What do you expect to see in the bar for one record: 1h with a rate of 100 / internal rate of 90.

> That means when using fee in activity or project the hourly rate is overwriten.

2. If no internal rate is set, the normal rate is used as fallback for the internal rate (see help text of internal rate). 
<img width="900" alt="Bildschirmfoto 2020-12-13 um 23 25 51" src="https://user-images.githubusercontent.com/533162/102025785-8d4fb080-3d9a-11eb-875b-8463fcc9aaf6.png">

HeinzWuert:
will wait for the documentation :-)

schippie:
[...]

PR #6057, opened 2026-07-18
Calculate internal rate of fixed-rate records based on duration

### LINKED ISSUE #4801: Internal price of a time record with a duration of 0 minutes should be € 0 but equals the internal hourly (!) rate instead.
### Describe the issue

Since [filtering between timesheets and expenses is currently not possible](https://github.com/kevinpapst/kimai-plugins/issues/119
), we follow @kevinpapst's advice and use a specific activity for expenses.

As a result, we have records with a duration of 0 minutes and a fixed price.
Unfortunately, kimai miscalculates the internal price if the duration is 0 minutes. It uses the internal rate of a full hour and we have not found a way to change it.

If the duration is one minute or more, the internal price is calculated correctly as a fraction of the internal hourly rate.

### I already tried

### Kimai version

2.96

### How do you run Kimai?

Other (please mention below)

Unknown

### Logfile

_No response_

### Screenshots

_No response_

TowyTowy:
## Changelog 

Fixed: internal rate of fixed-price records is now based on the entry's duration instead of always charging a full hour of the internal rate (0-minute entries correctly cost € 0). ([#4801](https://github.com/kimai/kimai/issues/4801))

## Description

Fixes #4801

When a timesheet record uses a fixed rate and no customer/project/activity rate rule defines an explicit internal price, the user's "internal rate" preference was copied as-is into the record's internal rate. That preference is an hourly rate - the internal rate is "the calculated value of internal hourly rate * duration" (as described in #2181) and the hourly-rate code path multiplies it by the duration accordingly. The fixed-rate path did not, so every fixed-price record was costed with a full hour of internal rate. For records with a duration of 0 (e.g. expenses tracked via fixed-price activities, as in #4801) the internal rate should be 0, but equalled the full internal hourly rate.

This PR calculates the internal-rate fallback for fixed-rate records from the internal hourly rate and the record duration, consistent with the hourly-rate path. Two things intentionally do not change: an explicit internal price on a fixed rate rule is still applied as a fixed amount, and users without a configured internal rate still fall back to the fixed rate.

Note: this updates the expected internal rates of the fixed-rate cases in the existing data providers (`b5`-`b9`, `d0`, `h0`), which pinned the old flat-hourly behavior, and adds regression cases (`l0`-`l2`) covering the #4801 scenario.

This PR was prepared with AI assistance (Claude); all changes were verified by running the test suite, PHPStan (src + tests configs) and php-cs-fixer.

## Types of changes

## Checklist
- [ ] I updated the documentation (not needed - behavior now matches documented semantics)
- [x] I agree that this code is used in Kimai (see license)


kevinpapst (changes requested):
TODO we need documentation for this behavior in 
https://github.com/kimai/www.kimai.org/blob/main/collections/_documentation/rates.md

I can do it next week, if you don't have the capacity

TowyTowy:
No need to spend your week on it — I opened the docs PR over at kimai/www.kimai.org#743, extending the "Internal price" and "Price calculation" sections of rates.md. Happy to reword anything there if it doesn't match how you'd put it.

src/Timesheet/RateService.php
```diff
@@ -57,7 +57,14 @@ public function calculate(Timesheet $record): Rate
 
         if (null !== $fixedRate) {
             if (null === $fixedInternalRate) {
-                $fixedInternalRate = (float) $record->getUser()->getPreferenceValue(UserPreference::INTERNAL_RATE, $fixedRate, false);
+                $internalHourlyRate = $record->getUser()->getPreferenceValue(UserPreference::INTERNAL_RATE, null, false);
```
kevinpapst: Please check if this achieves the same
```suggestion
                $internalHourlyRate = $record->getUser()->getPreferenceValue(UserPreference::INTERNAL_RATE);
```

kevinpapst (approved):

kevinpapst:
Thanks @TowyTowy 👍

PR #6060, opened 2026-07-19
Improve validation message when contract hours exceed 24h

### LINKED ISSUE #5277: Filling in > 24 hours on an employment contract gives error message with seconds
### Describe the issue

When I fill in for example 80 hours on a day in the edit contract section, I will receive an error message containing time in seconds: _This value should be less than or equal to 86400._ 

The error message should be like "There's only 24 hours in a day" 

### I already tried

### Kimai version

2.x.x

### How do you run Kimai?

Docker

8.4

### Logfile

_No response_

### Screenshots

IMG_0005

JamBalaya56562:
Fixes #5277.

## Description

When configuring an employment contract in the "Hours per day" mode, entering
more than 24 hours for a single day produced a confusing validation message that
exposed the internal limit in **seconds**:

> This value should be less than or equal to 86400.

Since users enter the value as hours (e.g. `25:00`), the message is hard to
understand. This replaces it with a clear, human-readable message.

## Before / After

| Before | After |
| --- | --- |
| before | after |

## Changes

- New `MaxDuration` constraint with the default message
  `A maximum duration of {{ value }} is allowed.`. The validator fills `{{ value }}`
  with the configured maximum, formatted through `App\Utils\Duration`, so a single
  translation covers every maximum duration.
- `WorkingTimeModeDay`: use `new MaxDuration(86400)` instead of `LessThanOrEqual(86400)`.
- `validators.en.xlf`: add the translation unit for the new message.
- `ProfileControllerTest`: add `testContractActionWithTooManyHours` covering the message.

## How it was verified

Reproduced on a running instance (Docker image `kimai/kimai2:apache`, v2.62.0,
with this change applied): entering `25:00` for Monday in Profile → Employment
contract now shows *"A maximum duration of 24:00 is allowed."* instead of the
seconds-based message (see the screenshots above).

## Types of changes

## Checklist
- [ ] I updated the documentation

### DIFF
--- a/src/Validator/Constraints/MaxDuration.php
+++ b/src/Validator/Constraints/MaxDuration.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Kimai time-tracking app.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace App\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+#[\Attribute(\Attribute::TARGET_PROPERTY)]
+final class MaxDuration extends Constraint
+{
+    public const MAX_DURATION_ERROR = '8b2f4a1c-7d3e-4f6b-9a05-1c8e2d7b3f49';
+
+    protected const ERROR_NAMES = [
+        self::MAX_DURATION_ERROR => 'MAX_DURATION_ERROR',
+    ];
+
+    public string $message = 'A maximum duration of {{ value }} is allowed.';
+
+    /**
+     * @param int $value the maximum allowed duration in seconds
+     * @param array<string>|null $groups
+     */
+    public function __construct(
+        public int $value,
+        ?array $groups = null,
+        mixed $payload = null,
+    ) {
+        parent::__construct([], $groups, $payload);
+    }
+}
--- a/src/Validator/Constraints/MaxDuration.php
+++ b/src/Validator/Constraints/MaxDuration.php
@@ -28,9 +28,14 @@ final class MaxDuration extends Constraint
      */
     public function __construct(
         public int $value,
+        ?string $message = null,
         ?array $groups = null,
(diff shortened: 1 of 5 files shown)


kevinpapst (changes requested):
Thanks @JamBalaya56562 👍 

Can you build a new `MaxDuration` constraint, using the default message `A maximum duration of {{ value }} is allowed.`?

The validator then builds the message like this
```
$duration = new \App\Utils\Duration();
$this->context->buildViolation($constraint->message)
                ->setTranslationDomain('validators')
                ->setParameter('{{ value }}', $duration->format($value))
                ->addViolation();
```

This makes it re-usable instead of having multiple translations for different durations.

JamBalaya56562:
Thanks for the guidance @kevinpapst — implemented in ee75535.

- New `MaxDuration` constraint with the default message
  `A maximum duration of {{ value }} is allowed.`
- `MaxDurationValidator` builds the violation as suggested, with the
  `validators` translation domain and the duration formatted through
  `App\Utils\Duration`.
- `WorkingTimeModeDay` now uses `new MaxDuration(86400)`, and the
  24-hour-specific translation is replaced by the generic one.

One detail I would like to confirm: your snippet passes `$value` to
`$duration->format()`. I used the configured maximum (`$constraint->value`)
instead, so the message renders as *"A maximum duration of 24:00 is allowed."* —
passing the submitted value would render *"A maximum duration of 25:00 is
allowed."*, which reads wrong. Happy to change it if you meant something else.

Verified against a running container (the 2.62.0 Apache image with this change
applied): entering `25:00` for Monday under Profile → Employment contract now
shows the new message. The screenshot in the PR description is updated
accordingly.

Pushed as a separate commit so the previously reviewed diff stays untouched.

kevinpapst (dismissed):

kevinpapst (approved):

kevinpapst:
Thanks @JamBalaya56562 👍

PR #6067, opened 2026-07-22
Do not reject bookings that consume no budget

### LINKED ISSUE #6015: "budget used up error" even if the time booking does not need budget
### Describe the issue

when the budget for a project is used up and you do not allow to overbook it in the settings you can't even book times without internal / external prices. However when you increase the budget, let's say 1 cent higher than the actual budget used you can book as many entries as you like as long those entries do not contain prices (price: 0,00). 
Same error in the demo. In our version (2.4.0) you can book it in the "my times" views but booking in the weekly hours lead to "budget used up error"
expected behavior: time tracking of hours should be possible when the hours do not consume budget

### I already tried

### Kimai version

2.4.0

### How do you run Kimai?

Docker

8.5

### Logfile

_No response_

### Screenshots

_No response_

JamBalaya56562:
## Changelog 

Fixed budget check rejecting billable entries that consume no budget (rate 0.00 / zero duration) when a budget was already overbooked ([#6015](https://github.com/kimai/kimai/issues/6015)).

Fixes #6015.

## Description

When a monetary budget is already overbooked (e.g. the budget was lowered after
entries were recorded, with "allow overbooking" disabled), the validator rejects
**every** new billable entry with *"Sorry, the budget is used up."* — even
entries with a rate of 0.00 that do not consume any budget. When the budget is
exactly used up (spent == budget) the same zero-rate entry is accepted, so the
behavior was inconsistent.

Root cause in `TimesheetBudgetUsedValidator::checkBudgets()`:

```php
$fullRate = ($stat->getBudgetSpent() + $rate);
if ($stat->hasBudget() && $fullRate > $stat->getBudget()) { // violation
```

With `spent > budget` this is true even for `$rate = 0`, so the violation is
raised although the new record adds nothing. The time budget check has the same
problem with zero-duration records, and edits that *lower* the rate/duration of
an existing entry (negative delta) were also rejected while overbooked.

## Fix

Only run the budget checks when the validated record actually increases the
consumption (`$rate` / `$duration` hold the delta for updated records):

| Scenario (budget 1000, overbooking disallowed) | Before | After |
| --- | --- | --- |
| spent 1001, new entry rate 0.00 | ❌ rejected | ✅ accepted |
| spent 3601s, time budget 3600s, new entry duration 0 | ❌ rejected | ✅ accepted |
| spent 1002, lower an existing entry from rate 1.00 to 0.00 | ❌ rejected | ✅ accepted |
| spent 900, new entry rate 101.00 | ❌ rejected | ❌ rejected (unchanged) |

Records with a zero or negative delta can never overbook a budget any further.

## Tests

- Added regression cases (`a_b2`, `p_k2`, `c_w2`): zero-rate entry against an
  overbooked budget must not raise a violation — one per activity/project/customer.
- Added `a_a2`: zero-duration entry against an exhausted time budget.
- `a_g0` / `p_g0` / `c_g0` (lowering the rate of an existing entry while
  overbooked) now expect no violation — these cases previously pinned the buggy
  behavior of #6015.
- `a_b` / `p_k` / `c_w` keep covering the "budget exceeded" violation, now with
  an entry that actually adds to the consumption (`spent 900 + rate 101 > 1000`),
  so no violation coverage is lost.

## How it was verified

Ran the updated test class against the released image (`kimai/kimai2:apache`,
PHP 8.3) with `composer install` + PHPUnit:

- **Unpatched validator + new tests: 57 tests, 7 failures** — exactly the new
  regression cases, reproducing #6015.
- **Patched validator: OK (57 tests, 71 assertions).**

## Types of changes

## Checklist
- [x] I updated the documentation

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3128,7 +3128,7 @@ parameters:
             path: src/Validator/Constraints/TimesheetBudgetUsedValidator.php
 
         -
-            message: "#^Parameter \\#4 \\$duration of method App\\\\Validator\\\\Constraints\\\\TimesheetBudgetUsedValidator\\:\\:checkBudgets\\(\\) expects int, float\\|int\\|null given\\.$#"
+            message: "#^Parameter \\#4 \\$durationDelta of method App\\\\Validator\\\\Constraints\\\\TimesheetBudgetUsedValidator\\:\\:checkBudgets\\(\\) expects int, float\\|int\\|null given\\.$#"
             count: 3
             path: src/Validator/Constraints/TimesheetBudgetUsedValidator.php
 
--- a/src/Validator/Constraints/TimesheetBudgetUsedValidator.php
+++ b/src/Validator/Constraints/TimesheetBudgetUsedValidator.php
@@ -155,7 +155,7 @@ final class TimesheetBudgetUsedValidator extends ConstraintValidator
                 $activityDuration = $duration;
             }
             $stat = $this->activityStatisticService->getBudgetStatisticModel($activity, $dateTime);
-            $this->checkBudgets($constraint, $stat, $value, $activityDuration, $activityRate, 'activity');
+            $this->checkBudgets($constraint, $stat, $value, $activityDuration, $activityRate, 'activity', $duration ?? 0, $rate);
         }
 
         if (null !== ($project = $value->getProject())) {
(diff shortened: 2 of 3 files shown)


kevinpapst:
@JamBalaya56562 there is a related existing integration test failing now, please have a look:
```
1) App\Tests\Controller\TimesheetControllerTest::testCreateActionWithOverbookedActivity
Expected 1 validation errors, found 0. Expected: #timesheet_edit_form_activity. Found: .
Failed asserting that 0 matches expected 1.
```

JamBalaya56562:
Thanks for catching that @kevinpapst — fixed in 8940261.

The first commit guarded the checks with the rate/duration **delta**, which wrongly
skipped edits that shrink an entry whose remaining values still overbook the budget —
exactly the scenario of `testCreateActionWithOverbookedActivity`.

The guard now uses the values of the validated record itself: a record with a rate of
0.00 never consumes the money budget and a record without duration never consumes the
time budget (the #6015 case), while every record that does consume budget is still
checked against the resulting total. I also added a unit case (`a_h3`) that pins the
"shrinking is not enough if the remaining total still overbooks" behavior.

Verified locally against the released image (PHP 8.3 + MariaDB):
- `TimesheetBudgetUsedValidatorTest`: OK (58 tests, 73 assertions)
- `TimesheetControllerTest --filter testCreateActionWithOverbookedActivity`: OK (1 test, 4 assertions)

Side note: the "Frontend verification" failure seems unrelated to this PR — `pnpm audit`
reports a high severity advisory for `fast-uri`.

kevinpapst (approved):

kevinpapst:
Thanks @JamBalaya56562 👍

PR #6075, opened 2026-07-26
Preserve multi-select values on reload

### LINKED ISSUE #4689: Multi-filter gets deleted on new selection
### Describe the issue

When selecting multiple customers and projects in the search and then selecting another customer, only the first project is kept in the list of the selected projects, the other ones are removed.

That happens because the javascript that handles API filled selects does not respect `select multiple`.

### I already tried

### Kimai version

2.12.0

### How do you run Kimai?

KIMAI CLOUD

8.1

### Logfile

_No response_

### Screenshots

https://github.com/kimai/kimai/assets/533162/1c12dd75-5b5e-45ea-9583-b82a052c97ff

Marukome0743:
## Description

Fixes #4689.

API-backed multi-select fields now preserve every still-valid selected value when their options are reloaded. Single-select fields keep their existing behavior.

The previous implementation stored only `node.value` before rebuilding the options. For a multiple select, that value represents only the first selected option. The updated implementation stores `selectedOptions` as an array, restores every value that still exists after the reload, and passes the restored array to Tom Select through the `data-reloaded` event.

### User impact

Selecting another customer in a multi-customer and multi-project filter no longer removes all previously selected projects except the first one. Projects that are not present in the reloaded API result remain correctly excluded.

### Visual verification

A browser regression harness reproduced the old behavior with 1 of 3 selections preserved and verified the fix with 3 of 3 selections preserved.

| Before | After |
|:--:|:--:|
| Before: only the first project remains selected | After: all valid projects remain selected |

### Validation

- Focused ESLint for `assets/js/forms/KimaiFormSelect.js`: passed after disabling the unrelated pre-existing `no-useless-assignment` finding at line 323.
- Production frontend build: passed with existing Sass deprecation warnings.
- PHP CS Fixer core check in the official Kimai Docker image: passed.
- Browser regression harness: before 1/3 selections, after 3/3 selections.
- Full `pnpm lint`: still reports 7 pre-existing errors outside this change, including the existing line 323 finding in `KimaiFormSelect.js`.

## Types of changes

## Checklist

### DIFF
--- a/assets/js/forms/KimaiFormSelect.js
+++ b/assets/js/forms/KimaiFormSelect.js
@@ -217,7 +217,7 @@ export default class KimaiFormSelect extends KimaiFormTomselectPlugin {
             console.log('Missing select: ' + selectIdentifier);
             return;
         }
-        const selectedValue = node.value;
+        const selectedValue = node.multiple ? [...node.selectedOptions].map(option => option.value) : node.value;
         // this could mean the field is required
         const allowEmpty = node.tomselect.settings.allowEmptyOption === true;
         // this means the field has no emPty option and should be pre-selected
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):
Thanks @Marukome0743 - looks great 👍

PR #6079, opened 2026-07-26
Rename customer "Contact" field to "Contact person"

### LINKED ISSUE #5770: Improve customer "Contact" wording
### Describe the issue

Instead of using "contact" I would use "contact person" instead to be more precise what this field means as contact could also imply a way how to contact them like email or whatever but her we are talking about an actual person. In German it could be translated as Ansprechpartner instead of Kontakt.

### I already tried

### Kimai version

2

### How do you run Kimai?

Virtual Server or alike

8.5

### Logfile

_No response_

### Screenshots

_No response_

JoeJoeflyn:
Fixes #5770.

## Description

The customer `contact` field stores a person name, but the label "Contact" was ambiguous and could be read as a contact method (email/phone). The same translation key was shared with the InvoiceTemplate `contact` field, which is a free-text textarea with different semantics.

Adds a dedicated `contact_person` translation key for the customer field (form label, detail view, list column header, export column header) while keeping the original `contact` key for InvoiceTemplate, per the maintainer's plan in #5770.

## Types of changes

## Checklist
screenshot-2026-07-26_21-50-08
screenshot-2026-07-26_21-50-22
screenshot-2026-07-26_21-50-33
screenshot-2026-07-26_21-50-51

### DIFF
--- a/src/Controller/CustomerController.php
+++ b/src/Controller/CustomerController.php
@@ -85,7 +85,7 @@ final class CustomerController extends AbstractController
         $table->addColumn('number', ['class' => 'd-none w-min']);
         $table->addColumn('company', ['class' => 'd-none']);
         $table->addColumn('vat_id', ['class' => 'd-none w-min']);
-        $table->addColumn('contact', ['class' => 'd-none']);
+        $table->addColumn('contact', ['class' => 'd-none', 'title' => 'contact_person']);
         $table->addColumn('city', ['class' => 'd-none']);
         $table->addColumn('country', ['class' => 'd-none w-min']);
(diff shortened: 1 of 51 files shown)


kevinpapst (changes requested):
Thank you 👍 

I would like to see another approach:
use the `bin/console kimai:translations --copy-resname` command to copy the `contact` key to a new `contact_info` key.

Change the english `contact` key to `Contact person` and the de one to `Kontaktperson`. 

Then use the new `contact_info` key for the InvoiceTemplate label.

That prevents us from having missing translations in both screens and we do not need to remap the Customer key to a different label.

JoeJoeflyn:
Thanks, fixed.

PR #6097, opened 2026-08-02
show weekly total in calendar week view

### LINKED ISSUE #5841: Show weekly total in Calendar's week view
### Describe the problem

I cannot easily see the duration of the whole week in week view.

Just like the very useful total duration below each date of every day of the week, show the duration of the whole week below the week number in the first column.

_No response_

### Screenshots

_No response_

JoeJoeflyn:
Closes #5841

Shows the total duration for the whole week below the week number in the first column of the calendar week view, matching the style of the existing daily totals in the column headers.

screenshot-2026-08-03_04-52-22

### DIFF
--- a/assets/js/widgets/KimaiCalendar.js
+++ b/assets/js/widgets/KimaiCalendar.js
@@ -766,11 +766,16 @@ export default class KimaiCalendar {
 
         const dailyTotals = document.querySelectorAll('.fc-dailytotal');
         dailyTotals.forEach(element => element.remove());
+        let weekTotalSeconds = 0;
         for (const dateValue in durations) {
             const durationValue = durations[dateValue];
 
             if (view === 'timeGridWeek') { // this is the week view
                 const headerCells = document.querySelectorAll(`th.fc-col-header-cell[data-date="${dateValue}"]`);
+
+                if (headerCells.length > 0) {
+                    weekTotalSeconds += durationValue;
+                }
 
                 headerCells.forEach(cell => {
                     const newElement = document.createElement('div');
@@ -778,6 +783,17 @@ export default class KimaiCalendar {
                     newElement.textContent = DATES.formatSeconds(durationValue);
                     cell.appendChild(newElement);
                 });
+            }
+        }
+
+        // show the weekly total next to the week number in the week view
+        if (view === 'timeGridWeek') {
+            const weekCushion = document.querySelector('.fc-timegrid-axis-cushion');
+            if (weekCushion !== null) {
+                const weekTotalElement = document.createElement('div');
+                weekTotalElement.classList.add('fc-dailytotal', 'fc-weektotal');
+                weekTotalElement.textContent = DATES.formatSeconds(weekTotalSeconds);
+                weekCushion.appendChild(weekTotalElement);
             }
         }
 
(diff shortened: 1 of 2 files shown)


kevinpapst (changes requested):
The overall approach is sound and fits the existing code: it reuses the already-computed `durations` map instead of re-iterating events, and it tags the new node with `fc-dailytotal` so the existing cleanup at `KimaiCalendar.js:767` removes it on every re-render — no duplicates when `eventsSet` fires repeatedly per event source. Good detail.

### Issues

**1. Total lands in the wrong cell when week numbers are disabled — blocking**

`document.querySelector('.fc-timegrid-axis-cushion')` is not scoped, and FullCalendar renders that class in two places:

- the **header** axis cell — only when `options.weekNumbers` is true (`@fullcalendar/timegrid/main.js:120-127`); otherwise the header `<th>` is just an empty `.fc-timegrid-axis-frame` with no cushion.
- the **all-day row** axis cell — rendered unconditionally as a `<span>` holding the "all-day" label (`@fullcalendar/timegrid/main.js:148`). `allDaySlot` is left at its default (`true`) in Kimai.

Kimai passes `weekNumbers: this.options['showWeekNumbers']` (`KimaiCalendar.js:128`), backed by the `calendar.week_numbers` setting. With that set to `false`, the first match in document order becomes the all-day cushion, and the weekly total renders next to the "all-day" label instead. (It also puts a `<div>` inside a `<span>` — invalid nesting.)

Scope the lookup to the header table, which uniquely contains the head axis:

    const weekCushion = document.querySelector('.fc-col-header .fc-timegrid-axis-cushion');

Worth considering whether the total should still render when week numbers are off — falling back to `.fc-col-header .fc-timegrid-axis-frame` would cover that, since the frame is always present.

**2. Possible clipping in the axis cell**

`.fc-timegrid-axis-cushion` is capped at `max-width: 60px` and its parent `.fc-timegrid-axis-frame` is `overflow: hidden` (timegrid `main.css`). Worth verifying with a three-digit hour total (>100h across a week is plausible for an admin viewing a team calendar) that nothing gets cut off.

**3. `text-align: center` is a no-op**

The new element is `display: inline-block`, so it shrinks to its text width — there's nothing to center. The parent frame is `justify-content: flex-end`, so the total right-aligns regardless. Either drop the rule or make the element `display: block` so it actually aligns under the week number.

**4. Styling diverges from `.fc-dailytotal`**

`.fc-dailytotal` has no CSS in the project at all — the daily totals are unstyled divs. Adding `padding: 2px 4px` only to the week total makes it sit slightly differently from the six daily totals on the same header row. Either drop the padding or apply it to both for a consistent header row.

### Smaller notes

- **The `headerCells.length > 0` guard deserves a comment.** Its effect is to exclude event dates that have no visible column — hidden weekends (`weekends: false`) and dates shifted out of range by the `.toUTC()` bucketing above. That's the right call (the weekly total then matches the sum of what's actually displayed), but the intent isn't obvious from the code. One sentence would help.
- **Accessibility:** a bare `37:30` under `W32` has no label. Same gap as the daily totals, but under a week number it reads as more ambiguous. A `title` attribute would be cheap.

### Tests & verification

Manual matrix worth walking before merge:

- `calendar.week_numbers` **on and off** (issue 1)
- weekends hidden (`calendar.weekends: false`)
- a week containing an all-day absence or public holiday
- RTL locale (the calendar honours a `direction` option)
- switching month → week → day → week, to confirm no stale or duplicated totals

JoeJoeflyn:
Thanks, addressed

PR #6111, opened 2026-08-07
Add customer filter to the user list reports

a1rwulf:
## Motivation
I want to see all hours booked by each person for all projects across a customer.
The monthly total just shows total hours per customer, with no ability to filter for persons.

## Description
The weekly, monthly and yearly "users" reports could only be narrowed down by project, so there was no way to see how much time each person booked across all projects of one customer.

Adds a customer select next to the existing team and project filters. The filter is applied in TimesheetStatisticService by joining the project and matching its customer, mirroring the existing project filter. Both filters combine.

Also drops the duplicate customer property from CustomerMonthlyProjects, which shadowed the one now provided by AbstractUserList. Its public API is unchanged.

## Types of changes

## Checklist

### DIFF
--- a/src/Reporting/MonthlyUserList/MonthlyUserListForm.php
+++ b/src/Reporting/MonthlyUserList/MonthlyUserListForm.php
@@ -9,6 +9,7 @@
 
 namespace App\Reporting\MonthlyUserList;
 
+use App\Form\Type\CustomerType;
 use App\Form\Type\MonthPickerType;
 use App\Form\Type\ProjectType;
 use App\Form\Type\ReportSumType;
@@ -31,6 +32,10 @@ final class MonthlyUserListForm extends AbstractType
         ]);
         $builder->add('team', TeamType::class, [
             'multiple' => false,
+            'required' => false,
+            'width' => false,
+        ]);
+        $builder->add('customer', CustomerType::class, [
             'required' => false,
             'width' => false,
         ]);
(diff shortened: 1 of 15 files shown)


a1rwulf:
@kevinpapst would you be interested in this contribution?

a1rwulf:
ping

kevinpapst:
Please provide some screenshots

a1rwulf:
> Please provide some screenshots

Screenshot added to the PR description.
Basically besides projects it allows to filter for customer in the weekly, monthly and yearly reports for all users.

src/Reporting/MonthlyUserList/MonthlyUserListForm.php
```diff
@@ -34,6 +35,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'required' => false,
             'width' => false,
         ]);
+        $builder->add('customer', CustomerType::class, [
```
kevinpapst: Choosing a customer doesn't narrow the project dropdown. The weekly and yearly forms have the same issue. A user can pick Customer A and a project that belongs to Customer B, and the report then shows 0 hours for everyone with no hint why. 
Check other Kimai toolbars. I guess the setting 'project_enabled' => true might be enough, but you need to try that.

a1rwulf: Good catch, thanks! Fixed in bf6bbc51:

- The customer select now uses `project_enabled`, so changing the customer reloads the project dropdown with that customer's projects (same as in the other toolbars).
- It also sets `ignore_date`, so projects that have already ended still show up. Otherwise you couldn't pick them for past weeks/months/years, and the reloaded list would differ from the one on page load.
- When the report is opened with a customer already selected (e.g. after submitting or via URL), the project dropdown is filtered on the server side too.

Same change for the weekly, monthly and yearly reports, with controller tests for both the reload wiring and the server-side filtering.

PR #6113, opened 2026-08-08
feat: show locale formatting examples in user preferences (#4429)

### LINKED ISSUE #4429: (Nice to have) Demo language string formatting in the preferences page
### Describe the problem

Knowing the formatting that a given languages will use is a bit cumbersome to have to look out in another page.

Under the `Language` selector, show demo values of the fields `DATE`, `TIME`, `DURATION`, `DECIMAL`, `PRICE` and a toggle for `RTL` as shown in the `/help/locales` page but only for the selected language.

#### Extra Nice to Have
As an extra on the demo values, I would suggest using a fixed set of values that are hard to misinterpret and not current server (or client) values.

 - The `TIME` field should use afternoon value so it is clear that `15:34` or `03:34 PM` is either 24-hours or 12-hours notation without having to look for the `24-HOUR` boolean column.
 - The `DATE` field should have a date that year, month and day cannot be misinterpreted.
     - I may suggest `2035-11-25`. `35` cannot be anything else than the year as day can be at most `31`, same thing for month and day, `25` cannot be a month so it is a day, leaving `11` to be a month for sure.
 - The `DURATION` field should show a duration that cannot be misinterpreted for a time, so I would use more than 24 hours duration (like `25:48`).
 - The `PRICE` field, I don't know to which extend this is feasible, but since the currency is set at the Kimai instance level (at least, I can't a user changing it for himself), I would use the instance currency to demo the values.

Do nothing and let the lazy user click on the link 😅

### Screenshots

_No response_

alex-frolov:
## Description

Show demo values for the formatting of the currently active locale below the language selector on the user preferences page, as requested in #4429.

The values are calculated server-side for the user's current locale and are fixed demo values that cannot be misinterpreted:
- `DATE`: 2035-11-25 (year/month/day cannot be confused)
- `TIME`: 15:34 (afternoon, makes the 24h/12h notation obvious)
- `DURATION`: 25:48 / 25.80 (hours + decimal representation of the duration)
- `RATE`: 2794.83 (formatted with the instance default currency)

The examples update live when the language changes, without a page reload, so unsaved settings are not lost. The values for date/time are formatted with the existing Kimai frontend helpers, and labels stay server-side translated.

## Types of changes

## Checklist
- [x] I updated the documentation (not needed — UI feature)

Fixes #4429


kevinpapst:
Hi Alex, can you share a screenshot?

alex-frolov:
Screenshot-2026-08-08-11 14 05

kevinpapst:
You added it to the field that is responsible for translating the UI, not the field that is responsible for formatting the data. That is the one below and it has a link to a separate page where all the formats are shown.

alex-frolov:
Thanks for the feedback, you are right!

I moved the "Formatting examples" block below the **Locale** field ("Time, date and currency format") — that is the one that actually controls the data formatting. The demo values are calculated from the user's current locale, so the block now sits right next to the field it describes.

alex-frolov:
Screenshot-2026-08-08-13 06 14

alex-frolov:
Hi Kevin,

Just a gentle follow-up on my last change (Aug 8) — I moved the "Formatting examples" block below the **Locale** field as you suggested and shared a screenshot. Did you have a chance to take a look?

Happy to adjust anything else if needed. Thanks!

kevinpapst:
There is the help page, which shows all available formats.
Your additions have a small UX boost, but what really would help is what you mentioned as not included: `no live update on language switch`.
There are some helpers existing that convert between the backend and frontend formats, so they can be used in Javascript. 
Also I would I make it one row and moving the word "examples" in front of the actual examples

alex-frolov:
Hi @kevinpapst, thanks for the feedback!

I implemented the live update: the formatting examples below the locale field now update immediately when the locale changes — no page reload, so unsaved settings are not lost.

How it works:
- The server passes date/time formats and the RTL flag for all known locales, plus raw demo values (ISO date without timezone offset, duration, amount, instance default currency) to the template.
- The new KimaiLocaleDemo plugin (registered in KimaiLoader) listens to the locale select and recalculates date/time via KimaiDateUtils (Luxon DATE.format), duration via formatSeconds, decimal/price via Intl.NumberFormat — using the existing JS helpers you mentioned.
- Labels stay server-side translated; only the values update, and the demo block stays a single line with the label in front.

Screenshots with three different locales selected (nl_NL, pl_PL, zh_CN) show the live update in action.

Could you confirm this matches what you had in mind? Happy to adjust anything.
kimai-screen-locale-1-2026-08-15
kimai-screen-locale-2-2026-08-15
kimai-screen-locale-3-2026-08-15

kevinpapst (changes requested):
Lets simplify it further:
- Remove the string `Formatting examples`
- Remove the `RTL` - the three locales that use it, know that anyway 
- Merge "Duration an Decimal duration" into one: `Duration: 25:48 / 25.80`
- Remove the string `Decimal`
- Replace `Price` with the existing translation for `rate` (not ideal, but better than new translations)

This allows to remove all new translations.

And the last point: make it inline javascript for the one page, otherwise it loads on every page without being used.

alex-frolov:
Thanks for the review! All points are implemented:

- Removed the "Formatting examples" label
- Removed the RTL flag
- Merged duration and decimal duration into a single value: Duration: 25:48 / 25.80
- Removed the "Decimal" label
- The money demo reuses the existing rate translation, no new translations are needed
- The JavaScript is now inline on the preferences page only, so nothing extra is loaded on other pages

All checks pass: phpstan, php-cs-fixer, ProfileControllerTest (37 tests).
kimai-screen-locale-2026-08-17-1
kimai-screen-locale-2026-08-17-2
kimai-screen-locale-2026-08-17-3

alex-frolov:
Hi @kevinpapst,

Thanks again for the review — the simplification is all in: removed the label/RTL/decimal duplicates, reused the existing `rate` translation, and the JavaScript is inline on the preferences page only. I also just fixed a small bug in that inline JS: for `zh_Hant_TW`-style locales the tag now converts **all** separators (`replace(/_/g, '-')`), not just the first one.

One question about the live update: the date/time examples are rendered with the existing `KimaiDateUtils.format()` helper, which intentionally hardcodes `en-us` (so `ko` shows `AM/PM` rather than 오후/오전 — matching what Kimai itself displays). That is consistent with the rest of the UI, so I have left it as is.

Copilot suggested switching to **server-generated examples per locale** for full localization, which would also make the live update match the server-rendered value exactly. Worth doing here — or is staying aligned with `KimaiDateUtils` (en-us) the right call? Happy to adjust either way.

Thanks!

alex-frolov:
Hi @kevinpapst,

Going with the current approach: the date/time examples stay aligned with `KimaiDateUtils` (which formats with `en-us`), so the demo matches what Kimai actually displays elsewhere. Server-rendered examples per locale would give full localization, but that's a bigger change — better as a follow-up if you want it.

From my side the PR is complete, every point from your review is in:

- removed the `Formatting examples` label
- removed `RTL`
- merged duration and decimal into one value (`Duration: 25:48 / 25.80`)
- removed `Decimal`
- `Price` reuses the existing `rate` translation

No new translation keys, and the JavaScript is inline on the preferences page only.

One thing is blocking a clean signal though: the workflow runs on this PR have never actually executed. Lint PHP, Tests, Frontend and Actions Security Analysis are all still in `action_required` (awaiting maintainer approval) on every commit so far. Could you approve them so CI can run? Happy to fix whatever it turns up.

Thanks!

kevinpapst:
Thanks for the update. 

There is large backlog of PRs and other things waiting, so I cannot tell you when I get to this here, just that I will at some point.

alex-frolov:
Hi @kevinpapst,

Thanks for approving the workflows on Sep 7. A quick summary of what they showed on 596235c:

- Tests ✅
- Lint PHP ✅
- Actions Security Analysis ✅
- Frontend ❌, only at the `pnpm audit` step: 4 high advisories in transitive dependencies (`fast-uri` via webpack-encore, GHSA-jqff-g426-hqxp). This is unrelated to the PR.

Unfortunately I force-pushed right after the run, which reset the approval. Sorry about that. I've now rebased onto the current `main`, and `pnpm audit` passes locally ("No known vulnerabilities found"). Whenever you get to it, the workflows need one more approval. I won't push anything until then, so the approval sticks.

No rush, I know the backlog is long. Thanks!

kevinpapst:
I know you invested quite a lot of time to find the best solution. Which was very valueable, as it showed that the approach has some problems, which are hard to solve. So I "stepped back" and thought about it again to find the simplest possible solution.

This is my approach: https://github.com/kimai/kimai/pull/6186

Most of the changes are "noise", creating the default setting for the users locale.
The most important change is the one in `src/Form/Type/UserLocaleType.php`.

Thank you very much for your participation. I know it can be frustrating if a PR does not get merged, but being able to show the locale in all places where the dropdown is actually used, seemed more important to me. Your work was important research and I really appreciate that you tried several approaches!

PR #6134, opened 2026-08-18
Fix weekly invoice grouping merging the same week from different years

TowyTowy:
## Problem

`WeeklyInvoiceCalculator::getIdentifiers()` groups invoice items by `format('W')`, which is the ISO week number on its own with no year:

```php
return [
    $invoiceItem->getBegin()->format('W')
];
```

So an invoice covering more than one year sums unrelated weeks into a single line. Timesheets from 2025-07-21 and 2026-07-20 are both week `30` and land in the same entry, with their durations and rates added together.

Every other calculator keys on a fully qualified value (`DateInvoiceCalculator` uses `Y-m-d`), so the bare week number looks like an oversight rather than a deliberate choice.

## Fix

Group on `format('o-W')`.

`o` is the ISO week-numbering year, which is the correct partner for `W` and not interchangeable with `Y`: 2024-12-30 belongs to ISO week 1 of 2025, so `Y-W` would key it as `2024-01` and split one ISO week across two invoice entries. `o-W` keys it as `2025-01`, together with 2025-01-02.

| date | `W` | `o-W` | `Y-W` |
|---|---|---|---|
| 2025-07-21 | 30 | 2025-30 | 2025-30 |
| 2026-07-20 | 30 | 2026-30 | 2026-30 |
| 2024-12-30 | 01 | 2025-01 | 2024-01 |

## Tests

Added `testSameWeekNumberInDifferentYearsIsNotMerged`, which feeds the calculator two timesheets one year apart in the same ISO week and asserts two entries come back. It fails before the change (actual size 1) and passes after.

Full `tests/Invoice/` suite is green: 207 tests, 1466 assertions.

## AI disclosure

Written with Claude (Claude Code). I verified the failing-before and passing-after states locally and checked the `o` versus `Y` distinction against real dates.

### DIFF
--- a/src/Invoice/Calculator/WeeklyInvoiceCalculator.php
+++ b/src/Invoice/Calculator/WeeklyInvoiceCalculator.php
@@ -24,7 +24,10 @@ final class WeeklyInvoiceCalculator extends AbstractSumInvoiceCalculator impleme
         }
 
         return [
-            $invoiceItem->getBegin()->format('W')
+            // The ISO week-numbering year ("o"), not the calendar year, because an
+            // ISO week can span two calendar years. Without a year the same week
+            // number from different years would be summed into one invoice entry.
+            $invoiceItem->getBegin()->format('o-W')
         ];
     }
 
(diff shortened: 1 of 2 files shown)


kevinpapst:
@TowyTowy The new test passes equally with `Y-W`, so it doesn't lock in the `o` semantics the code comment argues for. A later "simplification" to `Y-W` would keep the suite green while splitting one ISO week across a year boundary into two invoice lines. Want to add that one test to make it perfect?

kevinpapst (commented):
What n interesting edge case. 
I hope nobody really writes invoices spanning multiple years 🙈 

Thanks for the fix 👍

kevinpapst (approved):

PR #6146, opened 2026-08-21
Shorter duration entry: Interpret ints < 10 as hours and >= 10 as minutes

hacklschorsch:
## Description

When we see a plain integer duration value greater than 9, assume the user meant minutes (instead of hours).

Rationale:  Minimally faster time entry / less hovering the keyboard.  I assume most people don't track time shorter than 10 minutes.  Should one want to, they still can, using tried and tested syntax like `0:10`. 

A time tracking software I like has a "heuristic" like this.
It's a tiny feature, but one I grew to really like.
My new employer uses Kimai for time tracking, and I miss this - it's just a very little papercut/quality of life issue, but I run into it rather frequently, having entered a duration into Kimai about 700 times over the last fours months.
 
## Types of changes

## Checklist
- [x] I updated the documentation (see [here](https://github.com/kimai/www.kimai.org/pull/761))

### DIFF
--- a/assets/js/plugins/KimaiDateUtils.js
+++ b/assets/js/plugins/KimaiDateUtils.js
@@ -289,7 +289,11 @@ export default class KimaiDateUtils extends KimaiPlugin {
             let c = parseInt(duration);
             const d = parseInt(duration).toFixed();
             if (!isNaN(c) && duration === d) {
-                duration = (c * 3600).toString();
+                if (c < 10) {
+                    duration = (c * 3600).toString();
+                } else {
+                    duration = (c * 60).toString();
+                }
                 luxonDuration = Duration.fromISO('PT' + duration + 'S');
             }
         }


kevinpapst:
Thanks for sharing the idea!

What about typing `11m` ? That already works and there is no need to define arbitraries boundaries that no-one undertstands without consulting the docs.

hacklschorsch:
> 11m

IMHO that's slightly better than having to type a colon (which with a DE keyboard layout means I have to have both hands on the keyboard), but still much worse than being able to do data entry with one hand staying on the numpad and the other staying on the mouse (and not having to constantly switch my hands around).

Software devs regularly overestimate how GUIs needs to be "logical" or "consistent".
Humans, not machines, use GUIs.  GUIs need to be friendly so they spark joy, learnable so one doesn't have to consult a manual, and they have to be efficient so working with them (especially when it's chores) doesn't drag the user down.

I would be surprised if you would not be convinced by the proposed behavior if you tried entering a week of time tracking data at once, like I do every week.  That usually costs me 15m to 0.5, and is one of my most dreaded chores.

hacklschorsch:
IIANM, the 'duration' and 'start/end time' fields already behave slightly differently:

The ["Create & edit timesheets" video](https://www.youtube.com/watch?v=qVexU6KyWUE&t=104s) shows that entering `1345` in the end time field is interpreted as `13:45`.
Entering `1345` in the duration field, however, currently is interpreted as `1345:00`.

PR #6148, opened 2026-08-21
Use only translated locales for routes and cache warming

### LINKED ISSUE #6064: Secured_area session lifetime seams to be ignored
### Describe the issue

I've set in `config/packages/local.yaml`:

```yaml
framework:
    session:
        gc_maxlifetime: 31556952
        cookie_lifetime: 31556952

security:
    firewalls:
        secured_area:
            remember_me:
                lifetime: 31556952
```

which worked in an older Kimai release (don't know which was the last version it worked) but now it doesn't work anymore. I constantly need to login again when I do some actions in the invoice menu even though I shouldn't login for at least one year. Did something change in that area that I missed?

### I already tried

### Kimai version

2.61.0

### How do you run Kimai?

Virtual Server or alike

8.4

### Logfile

_No response_

### Screenshots

_No response_

Roshan931:
## Root cause

framework.enabled_locales contains ~300 locales, exploding translation cache warmup memory

## Changes

Changes cache warmup to only use translated locales and not all locales with subregions (which are used for data formatting.

Changes route requirements to only accept translated locales, as the ones with subregions would created cache entries during runtime.

## Details

Limit framework.enabled_locales to the 42 locales that actually ship translation files (new kimai_translated_locales parameter) instead of all ~300 known locales, and restore the explicit _locale route requirements based on %app_locales% so URL/locale handling stays exactly as before. Symfony's Translator::warmUp() keeps every warmed catalogue (including a full copy of each fallback catalogue) in memory, so 300 locales x ~1500 messages exceeded the 128 MB limit.


kevinpapst:
Thanks for the idea. Some profiling infos would be good, before and after.

Roshan931:
I profiled Symfony Translator 6.4.42 against all 616 Kimai XLIFF resources in isolated PHP 8.5.9 processes: warming 303 locales peaked at 58 MiB and failed under a 48 MiB limit, while warming the 42 translated locales peaked at 42 MiB and completed under the same limit.

kevinpapst:
I took your PR and adapted it slightly to fix a bug that was apparently existing for a longer time: you could use untranslated locales in the url, so `/en_CY/` was possible without having translations.
This could be a BC for bookmarks with such invalid locales, but its unlikely, as you could not geenrate themn with Kimai directly. 

The PR speeds up cache warmup and reduces memory usage, awesome 🎉 

And it reduces the file cache size in var/cache/ from ~53 MB to 20 MB 😮

kevinpapst:
Test failures are unrelated and fixed in #6172

kevinpapst (approved):

kevinpapst:
Thanks @Roshan931 for the suggestion 👍

PR #6183, opened 2026-09-13
CI: Speedup PHPStan analysis

staabm:
## Description
PHPStan maintainer here

setup result cache persistence in CI per documentation in https://staabm.github.io/2023/10/21/phpstan-result-cache-gotchas.html#github-actions 

## Types of changes

## Checklist


kevinpapst:
Thanks @staabm 👍 

- Coverage is a good catch, a leftover from the time before I split this action from phpunit.
- I didn't know about the result cache, thanks for sharing!

But why the `--error-format=checkstyle | cs2pr` vs `vv` change? You are obviously the best person I can find to answer that question 😄

.github/workflows/linting.yaml
```diff
@@ -48,6 +48,14 @@ jobs:
                     path: "${{ env.composer_cache_directory }}"
                     key: ${{ runner.os }}-${{ matrix.php }}-${{ hashFiles('**/composer.lock') }}
 
+            - name: "Restore PHPStan result cache"
+              uses: actions/cache/restore@v3
+              with:
+                  path: ./tmp
+                  key: "result-cache-v1-${{ matrix.php-version }}-${{ github.run_id }}"
+                  restore-keys: |
+                      result-cache-v1-${{ matrix.php-version }}-
```
kevinpapst, line 57: is the minus at the end of `result-cache-v1-${{ matrix.php-version }}-` correct?

.github/workflows/linting.yaml
```diff
@@ -64,10 +72,10 @@ jobs:
                 run: vendor/bin/php-cs-fixer fix --dry-run --verbose --config=.php-cs-fixer.dist.php --using-cache=no --show-progress=none --format=checkstyle | cs2pr
 
             -   name: Run PHPStan for application
-                run: vendor/bin/phpstan analyse -c phpstan.neon --no-progress --error-format=checkstyle | cs2pr
```
staabm, line 67: `--error-format=checkstyle | cs2pr` is not required for phpstan, as when no `--error-format` given it will auto-detect that its running inside github-actions and produce github-action optimized annotations natively.

`cs2pr` is useful for tools which are not able to render github-action annotations on their own, but are able to produce checkstyle-format

.github/workflows/linting.yaml
kevinpapst: Sorry, did not read the linked page, will do now 😁

.github/workflows/linting.yaml
kevinpapst: Thanks for the explanation!

.github/workflows/linting.yaml
staabm: its correct, as it needs to match `key: "result-cache-v1-${{ matrix.php-version }}-${{ github.run_id }}"` without the last variable `${{ github.run_id }}`

.github/workflows/linting.yaml
```diff
@@ -48,6 +48,14 @@
                     path: "${{ env.composer_cache_directory }}"
                     key: ${{ runner.os }}-${{ matrix.php }}-${{ hashFiles('**/composer.lock') }}
 
+            - name: "Restore PHPStan result cache"
+              uses: actions/cache/restore@v3
```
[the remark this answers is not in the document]
kevinpapst: @staabm there is actions/cache@v6 already: https://github.com/actions/cache/releases

any specific reason for v3 or just copy & paste from the linked page?

.github/workflows/linting.yaml
```diff
@@ -48,6 +48,14 @@ jobs:
...
+              with:
+                  path: ./tmp
```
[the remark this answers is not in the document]
staabm: right - I see you are using `tmpDir` option in `phpstan.neon`. we need to match this path here.

.github/workflows/linting.yaml
```diff
@@ -77,3 +85,10 @@
 
             -   name: Check for security issues in packages
                 run: symfony security:check
+
+            - name: "Save result cache"
+              uses: actions/cache/save@v3
```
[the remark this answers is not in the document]
kevinpapst: That should be the correct sha
```suggestion
              uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```

staabm:
in the CI output we see

```
vendor/bin/phpstan analyse -c phpstan.neon --no-progress -vv
Result cache not used because the cache file does not exist.
Spawned worker 1/4 checked in: turbo off, OPcache on, trusted types off
Spawned worker 2/4 checked in: turbo off, OPcache on, trusted types off
Spawned worker 3/4 checked in: turbo off, OPcache on, trusted types off
Spawned worker 4/4 checked in: turbo off, OPcache on, trusted types off
Result cache is saved.

 [OK] No errors                                                                 
Elapsed time: 26.74 seconds
Peak memory: 60 MB (main process), 290 MB (largest of 4 spawned workers)
```

this means PHPStan is not able to use PHPStan turbo - which requires PHP 8.3+.
in case you want faster PHPStan in CI we could use PHP 8.3 (or even 8.5) and define a `phpVersion` in `phpstan.neon`.
that way you can unlock PHPStan turbo, while keeping PHP version analysis for PHP 8.2+

kevinpapst:
> in case you want faster PHPStan in CI we could use PHP 8.3 (or even 8.5) and define a phpVersion in phpstan.neon.

Yes, lets do that 👍 in [3 months](https://www.php.net/supported-versions.php) 8.2 support will be removed anyway

kevinpapst:
Down from [34 + 32 = 66](https://github.com/kimai/kimai/actions/runs/34606176226/job/103285021736) to [27 + 26 = 53](https://github.com/kimai/kimai/actions/runs/34765727251/job/103746514502?pr=6183) seconds on 8.2 without "Turbo" and without using the cache ?!?

staabm:
> Down from [34 + 32 = 66](https://github.com/kimai/kimai/actions/runs/34606176226/job/103285021736) to [27 + 26 = 53](https://github.com/kimai/kimai/actions/runs/34765727251/job/103746514502?pr=6183) seconds on 8.2 without "Turbo" and without using the cache ?!?

thats by disabling of `pcov` coverage driver - for which overhead can be up to 40% - see https://github.com/phpstan/phpstan-src/pull/6253

.github/workflows/linting.yaml
```diff
@@ -77,3 +79,69 @@ jobs:
...
                 run: symfony security:check
+
+            - name: "Save result cache"
+              uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+              if: always()
+              with:
+                  path: ./var/cache/phpstan
+                  key: "result-cache-v1-${{ matrix.php-version }}-${{ github.run_id }}"
+
+    static-analysis:
```
staabm: added a separate PHP 8.3 job, so we can leave lint/phpunit on PHP8.2 runtime

kevinpapst: Setting up two jobs, two clones, two composer installs ... consumes much more resources. Lets bump the complete job to 8.3. That's okay for me, considering that 8.2 is EOL shortly.

.github/workflows/linting.yaml
```diff
@@ -77,3 +71,65 @@ jobs:
...
+
+            -   name: Determine composer cache directory
+                id: composer-cache
+                run: echo "composer_cache_directory=$(composer config cache-dir)" >> $GITHUB_ENV
+
+            -   name: Cache Composer dependencies
+                uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+                with:
+                    path: "${{ env.composer_cache_directory }}"
+                    key: ${{ runner.os }}-${{ matrix.php }}-${{ hashFiles('**/composer.lock') }}
```
staabm: IMO when you are using `shivammathur/setup-php` dependencies are already cached and this step is not necessary

phpstan.neon
```diff
@@ -31,6 +32,7 @@ parameters:
     paths:
         - src
     tmpDir: %rootDir%/../../../var/cache/phpstan
+    resultCachePath: %tmpDir%/phpstan.php
```
staabm, line 35: use a separate result cache file between `phpstan.neon` and `tests/phpstan.neon`, so we will have 2 caches.

before adding `resultCachePath` the same file was overridden and could not be re-used

.github/workflows/linting.yaml
staabm: just to make you aware:

```
vendor/bin/php-cs-fixer fix --dry-run --verbose --config=.php-cs-fixer.dist.php --using-cache=no --show-progress=none --format=checkstyle | cs2pr
You are running PHP CS Fixer on PHP 8.3.33, but the minimum PHP version supported by your project in composer.json is PHP 8.2. Executing PHP CS Fixer on newer PHP versions may introduce syntax or features not yet available in PHP 8.2, which could cause issues under that version. It is recommended to run PHP CS Fixer on PHP 8.2, to fit your project specifics.
If you need help while solving warnings, ask at https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/discussions/, we will help you!
PHP CS Fixer 3.95.25 Adalbertus by Fabien Potencier, Dariusz Ruminski and contributors.
PHP runtime: 8.3.33
Loaded config default from ".php-cs-fixer.dist.php".
Running analysis on 3 cores with 10 files per process.
```

in the github action logs

phpstan.neon
kevinpapst: what is `%tmpDir%` ? Here in Github it doesn't matter. But I don't want it locally to point to /tmp/.

staabm: `%tmpDir%` will resolve to `%rootDir%/../../../var/cache/phpstan` which is defined 1 line above. so the same dir it was using before

.github/workflows/linting.yaml
kevinpapst: The good old warning ... can be ignored.

staabm:
PHPStan works now as expected:

grafik

kevinpapst:
Holy crap 😮 that's less than 4 seconds compared to 66 before. 

For the phpstan-src discussion the relevant improvement was disabled PCOV with 66 vs 53 seconds - so about 20%, if I am not mistaken.

Thanks for this "drive by" improvement @staabm 🎉  
Very much appreciated 👍

.github/workflows/linting.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
 
         strategy:
             matrix:
-                php: ['8.2'] # use lowest supported PHP version to prevent introducing syntax from newer PHP versions
+                php: ['8.3'] # use lowest supported PHP version to prevent introducing syntax from newer PHP versions
```
kevinpapst:
```suggestion
                php: ['8.3'] # usually lowest supported PHP version, but 8.2 is EOL in Dec. 2025 and PHPStan Turbo needs 8.3
```

kevinpapst (approved):

PR #6189, opened 2026-09-15
fix(api): preload team members users and preferences to avoid N+1

### LINKED ISSUE #6174: GET /api/teams/{id} runs two extra queries per team member
### Describe the issue

`GET /api/teams/{id}` costs **two extra queries per team member**, so the response time of
the endpoint grows linearly with team size.

`Team::$members` is a lazy `OneToMany` and `TeamMember::$user` a lazy `ManyToOne`, and the
`Team_Entity` serialization group exposes both:

```php
// src/Entity/Team.php:53-59
#[ORM\OneToMany(mappedBy: 'team', targetEntity: TeamMember::class, cascade: ['persist', 'remove'], fetch: 'LAZY', orphanRemoval: true)]
#[Serializer\Groups(['Team_Entity'])]
private Collection $members;

// src/Entity/TeamMember.php:29-35
#[ORM\ManyToOne(targetEntity: User::class, inversedBy: 'memberships')]
#[Serializer\Groups(['Default', 'Entity', 'Team_Entity'])]
private ?User $user = null;
```

So for each `TeamMember` the JMS serializer:

1. initialises the `User` proxy — `DoctrineProxySubscriber::onPreSerialize()` → `__load()`
   → **1 `SELECT` on `kimai2_users`**;
2. serializes the `locale` / `language` / `timezone` virtual properties, which go through
   `User::getPreferenceValue()` → `User::getPreference()` and initialise the lazy
   `User::$preferences` collection → **1 `SELECT` on `kimai2_user_preferences`**.

```php
// src/Entity/User.php:438-445
public function getPreference(string $name): ?UserPreference
{
    if ($this->preferences === null) {
        return null;
    }

    foreach ($this->preferences as $preference) {   // <- initialises the lazy collection
```

**Measured cost.** A team with N members, each holding one user-preference row:

| members | SQL queries for `GET /api/teams/{id}` |
|---|---|
| 1 | 10 |
| 2 | 12 |
| 4 | 16 |
| 8 | 24 |
| 16 | 40 |

Exactly `8 + 2×N`.

**Affected endpoints.** `GET /api/teams/{id}`, plus everything that returns the team
entity afterwards: `POST /api/teams`, `PATCH /api/teams/{id}`, and the
member / customer / project / activity sub-resource endpoints — they all serialize with
`GROUPS_ENTITY`. `GET /api/teams` (the collection) is **not** affected: its `Collection`
group does not expose `members`.

**Why the existing loaders don't catch it.** Kimai already solves this class of problem
with `src/Repository/Loader/*Loader.php`, and `TeamLoader` does batch-load
`members` + `members.user`. Two gaps:

* it never runs on the single-entity path — `TeamController::getAction(Team $team)`
  resolves the team through `MapEntity`, bypassing the repository method that applies the
  loader;
* even where it does run, it does not preload `user.preferences`, unlike `UserLoader`,
  which fetches them explicitly:

  ```php
  // src/Repository/Loader/UserLoader.php
  $qb->select('PARTIAL user.{id}', 'preferences')
      ->from(User::class, 'user')
      ->leftJoin('user.preferences', 'preferences')
  ```

### How to reproduce

1. In the `dev` environment, call `GET /api/teams/{id}` for a team with several members.
2. Open that request in `/_profiler` and look at its Doctrine panel.
3. Observe one `SELECT ... FROM kimai2_users WHERE id = ?` and one
   `SELECT ... FROM kimai2_user_preferences WHERE user_id = ?` per member, on top of the
   team and members queries. Adding members to the team increases the count by two each.

`GET /api/teams` stays flat for comparison.

### Possible solution

Either run `TeamLoader` on the single-entity path as well, or fetch-join the members'
users when the `Team_Entity` group is in play. Either way `TeamLoader` also needs the
users' `preferences`, the way `UserLoader` already does it — otherwise the second query
per member remains.

Happy to open a PR if you have a preference between the two approaches.

### How this was found

By running [query-guard](https://github.com/alex-frolov/query-guard) — a PHPUnit extension
that traces Doctrine queries during an ordinary test run — over kimai's own existing test
suite on `main`. No new tests, no synthetic data, no modified scenarios: just the SQL the
suite already produces, with all 4343 tests green.

Both halves of this one are **confirmed lazy loads rather than heuristic guesses**: the
tool inspects the call stack at query time, and found the uninitialized
`User::$preferences` `PersistentCollection` by name in one case and the uninitialized
`User` entity proxy in the other. That is why the report can name the exact association.

The scaling table above is a separate measurement against the endpoint with the member
count varied, not a suite counter.

### I already tried

### Kimai version

2.66.0 (`main` @ `5b0f4bb8`)

### How do you run Kimai?

Docker

8.3

### Logfile

Not applicable — this is a query-count issue, not an error. Doctrine ORM 2.20.13 / DBAL
3.10.6, MySQL 8.4. Happy to share the full query trace if useful.

### Screenshots

_No response_

Arslan-TR:
## What

`GET /api/teams/{id}` (and every other Team API action that returns the `Team_Entity` group: `PATCH`, and the member/customer/project/activity add+remove endpoints) resolves the team via Symfony's `#[MapEntity]`, which bypasses `TeamRepository`'s query methods where `TeamLoader` normally batch-preloads associations.

The JMS serializer then lazily initializes, per team member: the member's `User` proxy, and that user's `preferences` collection (needed for the `locale`/`language`/`timezone` virtual properties) — 2 extra queries per member, so response time scales linearly with team size.

## Fix

- `TeamRepository::loadTeam()`: runs `TeamLoader` on a single already-resolved `Team`, mirroring the existing `UserRepository::loadUserByIdentifier()` + `UserLoader` pattern used for the same problem on the User side.
- Call it from every `TeamController` action that serializes with `GROUPS_ENTITY`.
- `TeamLoader`: batch-preload the members' users' `preferences`, the same way `UserLoader` already does for its own loads.

## Testing

`php -l` on all 3 changed files. Traced the query plan by hand against `TeamLoader`/`UserLoader`'s existing (tested) queries — no local DB available in this environment to run the full suite/phpstan, CI will cover that.

Fixes #6174


kevinpapst (dismissed):

kevinpapst:
Thank you very much for your help @Arslan-TR 👍

kevinpapst (approved):

